PR #15, opened 2015-07-02
tests will break when pytest upgrades implicitly on travis, let's pin the version for now.

glyph:
This is a temporary fix for #14 .


hynek:
Thanks but it was easier to fix the right way. :)

glyph:
Works for me :).

PR #17, opened 2015-07-07
Support optional values through a new validator.

sam-falvo:
Unfortunately, I'm unable to run tests locally b/c of severe dependency
hell.  I am unable to get tox, detox, python setup.py test, and other
means of running tests to work at all.  So, since I'm blocked on a
project by the lack of optional validator support, I am going to close
my eyes, look away, and pull the trigger, and hope I hit the target.

If someone who has a working Python configuration can please be kind
enough to let me know if my changes fail any tests, I'd be very
appreciative.

Even better, I'd love to know why a stock Python distribution with
dependencies installed is incapable of running the tests for attrs.
However, this isn't the right forum to answer that question.

(This PR has the 17+ debugging commits cleaned up.)

### DIFF
--- a/attr/validators.py
+++ b/attr/validators.py
@@ -89,10 +89,6 @@ def provides(interface):
 
 @attributes(repr=False)
 class _OptionalValidator(object):
-    """
-    A validator that makes an attribute optional.  An optional field is one
-    which can be set to None in addition to its intended data type.
-    """
     validator = attr()
 
     def __call__(self, inst, attr, value):
@@ -109,8 +105,9 @@ class _OptionalValidator(object):
 
 def optional(validator):
     """
-    A validator that makes an attribute optional.  An optional field is one
-    which can be set to None in addition to its intended data type.
+    A validator that makes an attribute optional.  An optional attribute is one
+    which can be set to None in addition to satisfying the requirements of the
+    sub-validator.
 
     :param validator: Any other validator you wish to make optional.
     """
--- a/attr/validators.py
+++ b/attr/validators.py
@@ -85,3 +85,33 @@ def provides(interface):
     the value it got.
     """
     return _ProvidesValidator(interface)
+
+
+@attributes(repr=False)
+class _OptionalValidator(object):
+    """
+    A validator that makes an attribute optional.  An optional field is one
+    which can be set to None in addition to its intended data type.
+    """
+    v = attr()
+
+    def __call__(self, inst, attr, value):
+        if value is None:
+            return
+        return self.v(inst, attr, value)
+
+    def __repr__(self):
+        return (
+            "<optional validator for {type} or None>"
+            .format(type=repr(self.v))
+        )
+
+
+def optional(x):
+    """
+    A validator that makes an attribute optional.  An optional field is one
+    which can be set to None in addition to its intended data type.
+
+    :param x: Any other validator you wish to make optional.
+    """
+    return _OptionalValidator(x)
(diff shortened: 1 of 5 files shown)


hynek:
As for your dependency hell, the answer is to use virtualenvs for everything:  https://hynek.me/articles/virtualenv-lives/  For CLI tools I like to use pipsi that will install them into separate virtualenvs (that’s where I have tox in).  To run the tests you only should require tox and run it.

But as you see, your changes passed in Travis so there’s that. :)

***

The idea makes sense to me, a few comments:

- You need to add the function to https://raw.githubusercontent.com/hynek/attrs/master/docs/api.rst so it’s included in the documentation.
- The changelog entry should be a fully qualified name that links to the docs. (i.e. Add \`attr.validators.optional <https://attrs...>\`_.)
- Since `optional` is agnostic of which validator it is used with, you can remove the interface-related tests.

The rest will follow inline, thank you for your contribution!

attr/validators.py
```diff
@@ -85,3 +85,33 @@ def provides(interface):
     the value it got.
     """
     return _ProvidesValidator(interface)
+
+
+@attributes(repr=False)
+class _OptionalValidator(object):
+    """
+    A validator that makes an attribute optional.  An optional field is one
+    which can be set to None in addition to its intended data type.
```
hynek: This is wrong, `optional` works with *any* validator.

attr/validators.py
```diff
@@ -85,3 +85,33 @@ def provides(interface):
...
+        return (
+            "<optional validator for {type} or None>"
+            .format(type=repr(self.v))
+        )
+
+
+def optional(x):
+    """
+    A validator that makes an attribute optional.  An optional field is one
+    which can be set to None in addition to its intended data type.
```
hynek: same

attr/validators.py
```diff
@@ -85,3 +85,33 @@ def provides(interface):
...
+        return self.v(inst, attr, value)
+
+    def __repr__(self):
+        return (
+            "<optional validator for {type} or None>"
+            .format(type=repr(self.v))
+        )
+
+
+def optional(x):
```
hynek: I’m not enthusiastic about `x` as the parameter name.  Let's call it `validator` which it is. :)

attr/validators.py
sam-falvo: I don't understand this feedback.  Can you clarify what is misleading?

hynek: I’m saying you’re selling your validator short. :)

`to None in addition to its intended data type.` is too limiting.  `attrs` may ship only with datatype related validators, but as ppl have plenty of custom ones.  The current doc string makes it sound like optional will not work with them.

docs/changelog.rst
```diff
@@ -16,7 +16,9 @@ The third digit is only for regressions.
 Changes:
 ^^^^^^^^
 
-- none yet
+- Add optional_ validator.
```
hynek: I’m sorry I have to pick on this again but the changeling is where people learn about new things in a version so it should be more inviting.

Suggested alternative:

```
Add :func:`attr.validators.optional` that wraps other validators allowing attributes to be ``None``.
```

sam-falvo:
I was wondering if you'd had a chance to re-review this patch?  This is a project blocker for me.  Thanks!

hynek:
I'm sorry I haven't got a notification about your changes so please ping a PR if you address a review.

Right now at a conference but I'll look at it when I'm back.

hynek:
Merged in b54bbaaa53096b2920e7637f55b0521ce92204a2 – thanks!

PR #20, opened 2015-07-21
Fix the repr to work for non-redecorated subclass.

cyli:
Currently, if you subclass an `@attr.s`-decorated class without re-decorating with another `@attr.s`, the `repr` displays the name of the superclass, not the subclass.


hynek:
Thanks!

Issue #26, opened 2015-09-10
Allow conversion of parameters


radix:
I'd personally be able to get rid of a lot of boilerplate (either at call sites or as `__init__` methods) from my classes if attrs allowed me to supply a conversion callable with my `attr.ib`s. 

e.g.

```python
from pyrsistent import pmap

@attr.s
class Foo(object):
    mapping = attr.ib(coerce=pmap)
```

The `coerce` argument (naming to be bikeshedded) would be called with the value passed in after validation, and the return value would be used as the actual attribute value.

Would such a feature be accepted as a patch?

radix:
or... maybe coerce should happen before validation. maybe that is one reason why this hasn't been done yet :)

hynek:
Maybe a more generic name like "hook" would be more apt?

econchick:
Omg I need this feature. NEED.

PR #27, opened 2015-09-16
Allow auto-converting of attribute values with a `convert` arg to attr.ib()

radix:
Fixes: #26 

Pretty straightforward change. I decided to go with running convert post-validator. Bikeshedding welcome.


hynek:
On an Italian beach so only briefly:

- change log please
- two spaces after periods
- an example pls

Thanks! :)

radix:
I think I may have convinced myself that conversion should go before validation, not after.

For example, say I have a validator that wants to make sure that it receives a `PMap` with `"name"` and `"id"` keys. I would want it to happen after conversion, so that my plain dict could be converted to a `PMap` before validation.

However, this would mean that converters may end up raising exceptions for bad inputs instead of validators. For example, if you pass an integer instead of a mapping, the converter will just blow up when trying to convert it to a pmap, instead of the validator (perhaps) raising a nicer exception. I think this is probably acceptable, because I think it's more important to be able to get data into the right shape before validation happens.

I'm gonna change that in my branch.

docs/changelog.rst
```diff
@@ -16,7 +16,7 @@ The third digit is only for regressions.
 Changes:
 ^^^^^^^^
 
-None
+- Add a `convert` argument to :func:`attr.ib`, which allows specifying a function to run on arguments.  This allows for simple type conversions, e.g. with ``attr.ib(convert=int)``. `[26] <https://github.com/hynek/attrs/issues/26>`
```
hynek: semantic new lines pls

docs/examples.rst
```diff
@@ -329,6 +329,43 @@ You can also disable them globally:
    TypeError: ("'x' must provide <InterfaceClass __builtin__.IFoo> which 42 doesn't.", Attribute(name='x', default=NOTHING, validator=<provides validator for interface <InterfaceClass __builtin__.IFoo>>, repr=True, cmp=True, hash=True, init=True), <InterfaceClass __builtin__.IFoo>, 42)
 
 
+Conversion
+----------
+
+Attributes can have a ``convert`` function specified, which will be called with
```
hynek: semantic new lines pls

docs/examples.rst
```diff
@@ -329,6 +329,43 @@ You can also disable them globally:
...
+.. doctest::
+
+    >>> @attr.s
+    ... class C(object):
+    ...     x = attr.ib(convert=int)
+    >>> o = C("1")
+    >>> o.x
+    1
+
+Converters are run _before_ validators, so you can use validators to check the
```
hynek: semantic new lines pls

docs/examples.rst
```diff
@@ -329,6 +329,43 @@ You can also disable them globally:
...
+final form of the value.
```
hynek: semantic new lines pls

attr/_make.py
```diff
@@ -81,6 +82,12 @@ def attr(default=NOTHING, validator=None,
 
     :param init: Include this attribute in the generated ``__init__`` method.
     :type init: bool
+
+    :param callable convert: :func:`callable` that is called by
```
hynek: aw I forgot about this syntax! could you convert the remaining arguments to it too for consistency please?

hynek:
I wonder, wouldn’t it be better to make the signature of converts the same as for validators? (i.e. passing the attribute and the instance along with the value)  It gives more flexibility and it’s nice to have some symmetry.

hynek:
*bump*

I would like to do a release with #29, it would be nice if we could get this in too?

radix:
@hynek I think the signature should be just the value to maximize the ability to use existing functions of the form `f(a) -> b` in the Python ecosystem (e.g. built-in functions like `int`, `frozenset`, and things like `pyrsistent.freeze`). I could imagine perhaps adding some introspection code to allow three-argument functions but I don't really see a use case for them now.

Sorry for dropping this for so long... I'll update the branch

Issue #28, opened 2015-10-05
Cost of validator is prohibitive


cournape:
I noticed that creating instances of attr objects was expensive:

```
# python >= 3.3, rough timing
import time

from attr import attributes, attr
from attr.validators import instance_of

@attributes
class Arch(object):
    name = attr("name")

@attributes
class ValidatedArch(object):
    name = attr("name", validator=instance_of(str))

def build_platform():
    return Arch("x86")

def timeit(func):
    times = []
    for i in range(10000):
        start = time.perf_counter()
        func()
        times.append(time.perf_counter() - start)
    return "Min cost across 10000 iterations: {:.2f} us".format(min(times) * 1e6)

print(timeit(lambda: Arch("x86")))
print(timeit(lambda: ValidatedArch("x86")))
```

I see a factor of ~ 50x (0.8us vs 34 us) on my machine. A quick profiling shows that almost all the difference comes from the deepcopy in the `attr._make.fields` function.

For non trivial classes with a few attributes, the construction cost quickly reaches the hundreds of us.

hynek:
Hi, thanks for the detailed report!

I indeed haven’t cared much about creation performance until now and it seems like I was a bit too careful by using a public API that uses deep copy there.

Just iterating over `__attrs_attrs_` does this:

```
$ python t.py # before
Min cost across 10000 iterations: 0.55 us
Min cost across 10000 iterations: 31.03 us
$ python t.py # after
Min cost across 10000 iterations: 0.56 us
Min cost across 10000 iterations: 1.31 us
```

Since the `__attrs_attrs__` object is read-only after creation, it should be safe to not lock anything.

Opinions?

cournape:
That certainly seems to fix the most glaring performance issues I noticed with attrs. If `__attrs_attrs__` is indeed read-only, that seems like a no-brainer ?

cournape:
This issue is a bit critical for us, is there something more I can do to make this fix appear in a next release ?

cournape:
(did not want to sound too pushy, if it is difficult to make it happen soon, we can fairly easily backport this in a bundled attr until a new official release)

hynek:
If you’d have pinged the PR yesterday (GH doesn’t notify me of new pushes…), I might have made a quick release today but now I’m packing up for PyCon JP and won’t be back until next Wednesday.  I will try to find a minute to push it out tho.

cournape:
[...]

PR #29, opened 2015-10-06
ENH: query attributes through private API to speed up validate.

cournape:
Proposal fix for #28

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -10,6 +10,7 @@ It’s the spiritual successor of `characteristic <https://characteristic.readth
 
 The following folks helped forming ``attrs`` into what it is now:
 
+- `David Cournapeau <https://github.com/cournape/>`_
 - `Glyph <https://github.com/glyph/>`_
 - `HawkOwl <https://github.com/hawkowl>`_
 - `Lynn Root <https://github.com/econchick>`_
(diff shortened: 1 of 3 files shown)


hynek:
Thanks!

So since the tests didn’t use to cover the error before (it came thru `fields`), the coverage sank.

But since the ValueError wasn’t documented before, but OTOH the docs explicitly ask for a decorated instance, I’m tending to not handle  that special case at all.

hynek:
you may want to use the delay I explained in the other ticket to add yourself to AUTHORS.rst and add the change to the changelog.  Thanks!

cournape:
I did, though it seems a bit of a steal given the trivial contribution :)

hynek:
I have a simple “every change is scared” policy. :)  Also not the code alone counts but also the way there.  A good bug report is +Inf times better than a half-assed patch.

That said, you forgot the changelog entry. :D  (`docs/changelog.rst`)

cournape:
here we go. Hope you had a good time in Tokyo !

hynek:
It was…intense. :)

Thanks!

cournape:
@hynek is there something I can help w/ to make `15.2.0` happen ?

hynek:
I’m doing some repo janitoring right now and will release afterwards.

PR #35, opened 2016-03-17
Feature/slots

Tinche:


tests/test_dark_magic.py
```diff
@@ -14,15 +14,26 @@ class C1(object):
...
 
+@attr.s(slots=True)
+class C1Slots(object):
+    x = attr.ib(validator=attr.validators.instance_of(int))
+    y = attr.ib()
+
 foo = None
 
 
-@attr.s()
```
hynek: Those parens are on purpose to test a different way trough the decorator.

hynek:
On a first scroll a few nits to keep you busy until I have actual time for a review :):

Please:
1. replace `class_` through `cls`.
2. use my docstring style consistently (mostly test_slots.py)

hynek:
OK one more fluff wish:  could you please make all asserts to be `assert expected == rv`?  I don’t Yoda in `if`s either but I like to have the expected values lined up. :)

hynek:
Just so we don’t forget:

- [ ] document that `slots=True` replaces the class

src/attr/_make.py
```diff
@@ -194,6 +203,21 @@ def wrap(cl):
             cl = _add_hash(cl)
         if init is True:
             cl = _add_init(cl)
+        if slots:
+            cl_dict = dict(cl.__dict__)
+            cl_dict['__slots__'] = tuple(ca_list)
+            for ca_name in ca_list:
+                # It might not actually be in there, f.e. if using 'these'.
```
hynek: f.e. → e.g.

src/attr/_make.py
```diff
@@ -194,6 +203,21 @@ def wrap(cl):
             cl = _add_hash(cl)
         if init is True:
             cl = _add_init(cl)
+        if slots:
+            cl_dict = dict(cl.__dict__)
+            cl_dict['__slots__'] = tuple(ca_list)
```
hynek: *cough* ' → "

(yes I *do* feel like a dick pointing this out, thank you for asking ;))

src/attr/_make.py
```diff
@@ -194,6 +203,21 @@ def wrap(cl):
...
+            for ca_name in ca_list:
+                # It might not actually be in there, f.e. if using 'these'.
+                cl_dict.pop(ca_name, None)
+            cl_dict.pop('__dict__', None)
+
+            if repr_ns is None:
+                class_name = getattr(cl, "__qualname__", cl.__name__)
```
hynek: let’s call it cl_name if everything around is cl_something

src/attr/_make.py
```diff
@@ -194,6 +203,21 @@ def wrap(cl):
             cl = _add_hash(cl)
         if init is True:
             cl = _add_init(cl)
+        if slots:
+            cl_dict = dict(cl.__dict__)
+            cl_dict['__slots__'] = tuple(ca_list)
+            for ca_name in ca_list:
+                # It might not actually be in there, f.e. if using 'these'.
+                cl_dict.pop(ca_name, None)
```
hynek: If I understand it correctly, then using slots means that *only* attr.s attributes are preserved, right?

tests/test_make.py
```diff
@@ -246,26 +246,30 @@ class C(object):
         assert sentinel == getattr(C, method_name)
 
     @pytest.mark.skipif(not PY3, reason="__qualname__ is PY3-only.")
-    def test_repr_qualname(self):
+    @pytest.mark.parametrize('slots_outer', [True, False])
```
hynek: you can combine this to

```python
@pytest.mark.parametrize("slots_outer,slots_inner", [
    (True, False),
    (False, True)
])
```

tests/test_make.py
```diff
@@ -246,26 +246,30 @@ class C(object):
...
-            @attributes
+            @attributes(slots=slots_inner)
             class D(object):
                 pass
 
         assert "C.D()" == repr(C.D())
         assert "GC.D()" == repr(GC.D())
 
-    def test_repr_fake_qualname(self):
+    @pytest.mark.parametrize('slots_outer', [True, False])
```
hynek: same

hynek:
all in all this is exciting :)

src/attr/_make.py
Tinche: Really? I feel the single quotes produce less visual noise.

tests/test_make.py
Tinche: It's not the same though, doing it this way will produce a cartesian product (i.e. test all combinations). Want me to put in all four?

src/attr/_make.py
Tinche: __slots__ will contain only 'a'. Assuming 
```
c = C()
C.b == 42
c.b == 42
c.b = 41  # is an error.
C.b = 41  # isn't, and will work.
```

Tinche:
Pasting the answer link to a review question here so it doesn't get hidden by GitHub:

https://github.com/hynek/attrs/pull/35#discussion_r56902838

tests/test_make.py
hynek: oh poor man’s hypothesis, I see. :)   hm I think I prefer a more explicit version though b/c I’ll quickly forget what this means.  But you can totally precompute the cartesian product somewhere.

src/attr/_make.py
hynek: Yeah that’s interesting.   But yeah, we can’t just add everything to slots as we want so it’s okay.

hynek:
A question?  Wrong link?  Seemed more like a statement of facts. :)

Tinche:
Ok, a clarification then. :)

hynek:
Are you waiting for anything from me?

Tinche:
Er no, I'm at work currently. Will fix the parametrization decorators tonight. That's all for now right?

hynek:
That and docs. :)

FWIW, you can also use [hypothesis](https://hypothesis.readthedocs.org) for the decorators thing.  I tend to introduce it into most of my current projects.

Tinche:
Ok, about the docs. Obviously we need to document the slots argument to attr.s() and make_class, but we should probably have a small section somewhere on the benefits and drawbacks of slot classes. Any preference where?

hynek:
I’d say examples.  Bring an example and explain a bit.  But you don’t have to explain slots completely.  Only what’s special about *our* slots.  Feel free to link a general description.

Tinche:
Ok, I've added hypothesis in. The tests for py27 and pypy are actually failing for me currently, since hypothesis conditionally depends on enum34, and tox creates virtualenvs with ancient pip versions (1.5.6) which don't know to actually install enum34 for hypothesis. I will admit to not knowing how to fix this, and getting a little pissed off for tox for not having a mechanism to deal with this (or at least a discoverable mechanism).

I'll ping here when there's something to show for the docs, might be a few days.

hynek:
You mean local?  Because Travis passes just fine?  Try `tox -r` :)

Tinche:
I did try recreating, no dice. I guess my virtualenv package is just obsolete. Still frustrating to not be able to precisely tailor the test environment without hacks :/

hynek:
I use pipsi for tox and keep my virtualenv/pip up2date, yeah. :)

Tinche:
Just pushed some docs, for your inspection. Out of curiosity, do you get automatic notifications when I push new stuff?

hynek:
Nope, always ping tickets when you do something.   Getting a notification for each push would result in a lot of mails I guess but it would be nice if there’d be a button “notify maintainers” or something. :-/

hynek:
Please  do [semantic new lines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/). :)

Tinche:
Alright, am I doing it right? :)

hynek:
I'll give more detailed feedback tomorrow but you slightly over-did it. :)  No line feeds after commas please. :)

Tinche:
https://raw.githubusercontent.com/brandon-rhodes/blog/master/texts/brandon/2012/one-sentence-per-line.rst seems to use line feeds after commas liberally. ¯\_(ツ)_/¯

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
     ValueError: x must be be at least 0.
 
 
+Slots
+-----
+
+By default, instances of classes have a dictionary for attribute storage.
+This wastes space for objects having very few instance variables.
+The space consumption can become acute
```
hynek: not even a comma! :)

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+when creating large numbers of instances.
+
+Normal Python classes can avoid using a separate dictionary
```
hynek: neither here

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+   ...     y = attr.ib()
+
+
+.. note::
+
+    ``attrs`` slot classes can inherit from other classes
+    just like non-slot classes,
+    but some of the benefits of slot classes are lost if you do that.
+    If you must inherit from other classes,
+    try to inherit only from other slot classes.
```
hynek: make this just two lines please

I’ll really need to work out something to not send people to misleading documents. 😞

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+
+
+Slot classes are a little different than ordinary, dictionary-backed classes:
+
+* assigning to a non-existent attribute of an instance
```
hynek: please start itemizations that are whole sentences with a capital letter.

I also prefer `-` over `*`

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+  will result in an ``AttributeError`` being raised.
+  Depending on your needs, this might actually be a good thing
```
hynek: scratch the “actually” please

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+   ...     x = attr.ib()
+   ...     y = attr.ib()
+   ...
+   >>> c = Coordinates(x=1, y=2)
+   >>> c.z = 3
+   Traceback (most recent call last):
+       ...
+   AttributeError: 'Coordinates' object has no attribute 'z'
+
+* slot classes cannot share attribute names with their instances,
```
hynek: no LF

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+  while non-slot classes can.
+  The following behaves differently if slot classes are used:
+
+.. doctest::
```
hynek: I think this has to be indented to be part of the item?

docs/examples.rst
```diff
@@ -377,6 +377,82 @@ Converters are run *before* validators, so you can use validators to check the f
...
+  ...     x = attr.ib()
+  >>> C.x
+  Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None)
+  >>> @attr.s(slots=True)
+  ... class C(object):
+  ...     x = attr.ib()
+  >>> C.x
+  <member 'x' of 'C' objects>
+
+* since non-slot classes cannot be turned into slot classes
```
hynek: this is just two lines too :)

hynek:
Getting there! 🎉

Tinche:
Alright, I've incorporated your feedback to the best of my abilities :)

hynek:
Change log entry and I think we can merge. :)

hynek:
Nevermind, I’ll do it myself.  Thanks for all the work!

Tinche:
My pleasure, thanks for the reviews.

glyph:
I just saw this land in a release. Once upon a time, @fijal told me that `__slots__` were actually *bad* for PyPy.  Summoning him to this ticket, I hope: did I even understand that correctly at the time? If so, is it still true?  I want to use this because I love the added "no extra attributes" strictness, but I don't want to hurt performance there.

PR #37, opened 2016-04-01
ENH: bundle inspect.isclass to avoid paying the cost of inspect import.

cournape:
Importing `attrs` on my fairly decent machine takes around 20-30 ms. Since I am using `attrs` in CLI tools, every small bit counts, and I'd like to make `import attrs` faster.

`inspect` is known as an expensive module to import, and since we use it only for one trivial oneline method, would it be acceptable to remove `inspect` import ? That makes importing `attrs` more or less free on my machine.

### DIFF
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -10,6 +10,8 @@ PY2 = sys.version_info[0] == 2
 if PY2:
     import types
 
+    # We 'bundle' isclass instead of using inspect as importing inspect is
+    # fairly expensive (order of 10-15 ms for a modern machine in 2016)
     def isclass(klass):
         return isinstance(klass, (type, types.ClassType))
 
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -8,6 +8,11 @@ PY2 = sys.version_info[0] == 2
 
 
 if PY2:
+    import types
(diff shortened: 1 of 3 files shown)


hynek:
I’ve played with it with ipython %time and got consistetly ~15% speed boost which seems worth it.  So TIL although you made it sound more profound. :)

Can you add a comment to _compat that explains this?  I don’t want this reasoning to get lost over the years.

cournape:
So just to give a more precise number for posterity. On my i7 desktop machine:

``` bash
# each time is the best of a 5 iterations loop
time python -c ""
real	0m0.014s
user	0m0.012s
sys	0m0.004s
# master
time python -c "import attr"
real	0m0.028s
user	0m0.024s
sys	0m0.000s
# with this change
time python -c "import attr"
real	0m0.020s
user	0m0.020s
sys	0m0.000s
```

The difference is less striking on my macbook.

hynek:
Thanks!  I hope you’ve noticed the slots changes?  Seems like something you may care about.

cournape:
I haven't ! In my applications, speed is generally more important, but it is good to know we have the option.

hynek:
Leaner software is usually also faster software. :)

Issue #43, opened 2016-05-18
Do we want to guarantee deepcopied attributes when converting ?


cournape:
A reminiscence of #28:

```python
# python >= 3.3, rough timing
import time

from attr import attributes, attr
from attr.validators import instance_of

def decode_if_needed(v):
    if isinstance(v, bytes):
        return v.decode()

@attributes
class Arch(object):
    name = attr("name")

@attributes
class ConvertedArch(object):
    name = attr("name", validator=instance_of(str), convert=decode_if_needed)

def timeit(func):
    times = []
    for i in range(10000):
        start = time.perf_counter()
        func()
        times.append(time.perf_counter() - start)
    return "Min cost across 10000 iterations: {:.2f} us".format(min(times) * 1e6)

print(timeit(lambda: Arch("x86")))
print(timeit(lambda: ConvertedArch(b"x86")))
```

``` bash
Min cost across 10000 iterations: 1.40 us
Min cost across 10000 iterations: 43.72 us
```

Now, the fix may not be as trivial as #28, because I can see how it may be part of the API that the original attribute is not touched when using `convert`, and not using `fields` to iterate in https://github.com/hynek/attrs/blob/master/src/attr/_make.py#L447 would break that promise. As I have not seen any test related to this, maybe that promise could be relaxed ?

hynek:
Eh could you rephrase + elaborate please?  I have a nasty head cold and can’t entirely follow. :)

cournape:
Let me try again :)

As in #28, the speed difference is because we iterate over the attr descriptors using [`fields`](https://github.com/hynek/attrs/blob/master/src/attr/_make.py#L447), which makes a `deepcopy` for every field.

In #28, we've fixed it by iterating using [`inst.__class__.__attrs_attrs__`](https://github.com/hynek/attrs/blob/master/src/attr/_make.py#L434). I was about to prepare a PR to do the same here, but then realized that the deepcopy may have bene intentional in this case. If not, I will be happy to prepare a PR to fix this.

hynek:
Um this would only matter, if someone would modify the attribute list in their container, right? (which they never should)

cournape:
Ok, let me prepare a PR then.

What is the rationale for using `deepcopy` in `fields` then ?

hynek:
I find there’s a difference between a public API like fields and misbehaving in an converter.  But maybe I’m wrong. :)

hynek:
[...]

PR #44, opened 2016-05-21
ENH: speed up convert.

cournape:
See #43

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -420,6 +420,14 @@ def fields(cl):
     return copy.deepcopy(attrs)
 
 
+def _fast_attrs_iterate(inst):
+    """ Fast internal iteration over the attr descriptors.
+
+    Using fields to iterate is slow because it involves deepcopy.
+    """
+    return inst.__class__.__attrs_attrs__
+
+
 def validate(inst):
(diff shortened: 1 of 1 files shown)


hynek:
and it actually helps? :)

cournape:
Yes. On master:

```
# python >= 3.3, rough timing
import time

def decode_if_needed(v):
    if isinstance(v, bytes):
        return v.decode()

@attributes
class Arch(object):
    name = attr("name")

@attributes
class ConvertedArch(object):
    name = attr("name", validator=instance_of(str), convert=decode_if_needed)

def timeit(func):
    times = []
    for i in range(10000):
        start = time.perf_counter()
        func()
        times.append(time.perf_counter() - start)
    return "Min cost across 10000 iterations: {:.2f} us".format(min(times) * 1e6)

print(timeit(lambda: Arch("x86")))
print(timeit(lambda: ConvertedArch(b"x86")))
```

```
Min cost across 10000 iterations: 0.64 us
Min cost across 10000 iterations: 43.74 us
```

With this PR:

```
Min cost across 10000 iterations: 0.61 us
Min cost across 10000 iterations: 2.99 us
```

PR #45, opened 2016-05-26
asdict - propagate dict_factory properly

Tinche:
So yeah, I made a dumb mistake when adding support for `dict_factory` - I forgot to ensure the factory propagates further than the initial class - to attributes containing dicts, lists and other instances of `attrs` classes. It came up instantly at work when we upgraded to actually use the feature :) My bad.

I'm pretty sure the `dict_factory` argument should propagate - one of the use cases is to use an OrderedDict to preserve order, and that's just not going to happen if ordinary dicts are used anywhere in the conversion.

The fix is simple, just propagate the argument in `asdict`.

Since the original tests didn't catch this, I've worked on expanding our Hypothesis strategies. (More challenging than just the fix :) In addition to the `simple_classes` strategy, there is also now a `nested_classes` strategy. `nested_classes` will generate `attrs` classes just like `simple_classes`, but attributes may also contain the following default values:
* instances of other `nested_classes` `attrs` classes
* lists, containing a single element pulled from `nested_classes`
* dicts, containing a key "cls" pointing to a class pulled from `nested_classes`

These cases should cover all cases in `asdict`. There's a new test, `test_recurse_property`, which asserts that for any `nested_classes` class, all dicts produced by `asdict` are of the appropriate class.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
     - python: "3.5"
       env: TOXENV=py35
     - python: "pypy"
-      env: TOXENV=pypy
+      env: TOXENV=pypy HYPOTHESIS_PROFILE=travis_pypy
 
     # Meta
     - python: "3.5"
--- a/conftest.py
+++ b/conftest.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import, division, print_function
 
+import os
 
 import pytest
+from hypothesis import settings
 
 
 @pytest.fixture(scope="session")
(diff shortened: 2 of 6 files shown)


hynek:
I guess we need to remove hypothesis health checks for PyPy?

Tinche:
Huh, weird. Let me install PyPy and see what's going on when I get a free moment.

hynek:
It’s just too slow on Travis.  Just disable the health check on PyPy.

Tinche:
There we go. Thought I stumbled upon a Hypothesis bug for a while, but it was just Tox not propagating environment vars. :)

hynek:
thanks!

Tinche:
Yay \o/

PR #49, opened 2016-08-10
Make copyright reproducible.

mithrandi:
This patch allows for [reproducible builds of attrs](https://bugs.debian.org/833886).

An alternative to all this skullduggery is to make the copyright year a static "2015". According to my understanding of the Berne Convention et al, the "first year of publication" is what you should list, so I think that would be fine?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,6 @@
 # serve to show the default.
 
 import codecs
-import datetime
 import os
 import re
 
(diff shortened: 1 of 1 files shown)


hynek:
Yes I've started to make it static in my projects after reading the same thing you do. :). Although I'm not entirely sure whether each version doesn't constitute as a new publication.   But I also don't care since I don't have the means to stop some anyone in court anyway.  I'm not gonna merge this but if you change it to 2015 only I will.

mithrandi:
Cool, PR updated.

hynek:
Lol merging PRs on phones.  So again: thanks!

PR #51, opened 2016-08-12
Make Attribute instances immutable and slotted.

Tinche:
If this looks good to you, I'll add something to the changelog. Since the docs already define Attributes are read-only, this should be backwards compatible.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,8 +25,9 @@ Changes:
 
 - ``attr.asdict``\ 's ``dict_factory`` arguments is now propagated on recursion.
   `#45 <https://github.com/hynek/attrs/issues/45>`_
-- ``attr.asdict`` and ``attr.has`` are significantly faster.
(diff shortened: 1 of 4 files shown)


hynek:
Yeah, I was always very explicit about this stuff being read-only so I’m not concerned.  As for changelog, maybe just add the function names and the PR# to the existing “faster” entry?

Tinche:
Ping.

hynek:
I’m conflicted about `_fast_attrs_iterate`.  Maybe kill it off but don’t use fields in internal functions and just grab `__attrs__attrs__` ?  Because the new version *is* a slowdown due to the safety checks which we don’t need internally…

By which I mean: `asdict` → use fields.  `_convert` → `__attrs_attrs__`. Does that make sense?

Tinche:
Agreed. (Pushed the change.)

fields() will give a proper error message if junk is passed in, but with a tiny slowdown. _convert can avoid the check. Other internal functions were already using the field directly.

hynek:
Thanks!

Tinche:
❇️❤️

PR #55, opened 2016-08-14
Example implementation of dict-like behavior

kissgyorgy:
Implement `MutableMapping` methods. The `__iter__` method makes `attr.asdict(c)` unnecessary, because `dict(c)` would be possible.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -142,7 +142,7 @@ def _transform_attrs(cl, these):
             had_default = True
 
 
-def attributes(maybe_cl=None, these=None, repr_ns=None,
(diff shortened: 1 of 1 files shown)


glyph:
I would very much not like to see this land.  Making objects iterable exposes a public interface which I don't want on most objects, locks in attribute ordering and even the count of attributes as public.  For example, users might start doing `attr1, attr2, attr3 = some_object` which I _very_ much want to fail.

hynek:
Yes I’m really sorry, I agree with glyph here.  Attrs tries to have a very narrow API surface for very [good reasons](https://www.youtube.com/watch?v=mTj297sGzxw).

And I don’t think it’s too much to ask to type two extra characters. :)

PR #56, opened 2016-08-14
Doc typo fixes

JesseWeinstein:
Thanks for making this library, and for explicitly requesting typo-fixing PRs.

I heard about it from https://lwn.net/Articles/697146

If you'd prefer, I can squash these commits into one; the only reason they are separate is due to GitHub's in-browser editor.


hynek:
Wow, thank you so much!

Squashing isn’t really required by contributors nowadays since GitHub supports “Squash & Merge”. :)

PR #57, opened 2016-08-14
Update why.rst

Rosslaew:
Fix of obvious typo.


hynek:
Thank you!  Turns out, any code that is not automatically tested, contains bugs. ;)

Issue #50, opened 2016-08-11
Thoughts on immutability


Tinche:
Hi,

working on making `Attribute`s immutable, and also thinking about immutability in Python in general, to evaluate how hard would it be to actually allow attrs to create immutable classes.

(Side note: adding `__slots__` to the `Attribute` class is probably a significant win for code bases with a lot of attrs classes, and can be trivially done by renaming the existing field `_attributes` to `__slots__`. This will be in the pull request too.)

Googling and playing around: the only way of making truly immutable classes in Python, I think, is by subclassing namedtuple. I'm not a huge fan of this approach though. My biggest philosophical objection is the fact that if this is done, `isinstance(obj, tuple)` and `isinstance(obj, typing.Sequence)` will be true. Instances of namedtuple subclasses are basically sequences/tuples. To me this is iffy.

Another approach is by disabling `__setattr__`, preferably combined with `__slots__` so the `__dict__` can't be modified directly. This isn't bulletproof, and `__init__` basically depends on it not being bulletproof to set the initial values. The instances can still be changed by doing `object.__setattr__(obj, 'name', val)`. If this is good enough, this approach is relatively straightforward.

What are your thoughts on this subject?

Here's my playground: https://github.com/Tinche/attrs/tree/feature/immutable-attributes

hynek:
Another approach might be using read-only properties?

hynek:
(but the setattr hack is good enough.  for sake of completion, in characteristic I did this this way: https://github.com/hynek/characteristic/blob/master/characteristic.py#L442-L482 💃

hynek:
🎉

Tinche:
Ok, now that the `Attribute` class has been optimized, what do you think about supporting immutability for user attr classes themselves? I think we have three logical courses of action.

* Don't add anything to attrs, let users deal with immutability themselves somehow. Maybe add a recipe or two to the docs?

* Support immutability on a per-attribute basis. I guess you'd be able to declare certain attributes as immutable, then we'd handle it somehow (i.e. generate properties without setters, but technical stuff not important right now)

```
@attr.s
class MyCoolClass(object):
    x = attr.ib()
    y = attr.ib(immutable=True)
```

* Support immutability on a class level. Allow users to declare an entire class as immutable and that's that. Technically, we'd probably use the __setattr__ switcharoo thing like for `Attribute`s.

```
@attr.s(immutable=True)
class MyCoolClass(object):
    x = attr.ib()
    y = attr.ib()
```

[...]

PR #60, opened 2016-08-16
[RFC] Implement immutability

### LINKED ISSUE #3: clean division between "value" and "object"
Objects, as in object-oriented programming, are side-effecty, mutable state, whose methods ought to represent I/O.

Values, as in functional programming, are immutable data, whose methods ought to represent computation.

Python does not have as good a division between these very different sorts of beasts as it should, but it does have one critical distinction: the `__hash__` method.

`characteristic` has this issue which crops up occasionally where you end up with objects that you can't put into a dictionary as a key, because one of its attributes is a dictionary.  Sometimes people expect to be able to do this because `characteristic` makes so many other things easy, and just expect dictionaries to suddenly be immutable; sometimes people expect hash-by-identity.

I propose that attr provide a way to specifically create two types of objects with a distinct interface; one that creates a "value" and one that creates an "object", so that users can see issues around mutability far earlier in the process.

So, for example, the "object" type would:

  - not provide `__hash__` by default, provide an `__eq__` that does structural equality, and `__gt__`/`__lt__` that just raise exceptions
  - if asked, provide an identity-based `__hash__`, but then also switch to an identity-based `__eq__`

and the 'value" type would

  - call `hash()` on all of its arguments at construction time so it would fail immediately if it contained a mutable type
  - fail immediately at class-definition time if any validator is mutable
  - provide immutable descriptors for all its attributes

hynek:
First shot at immutability to fix #3 and #50.

I would very much like some input from y’all that asked for it:  @glyph @radix @Tinche 

A couple of notes:

- I didn’t go for a separate decorator because that would be confusing and harder to implement.
- I called it `frozen` because most Pythonista have heard of `frozenset`.  `immutable` is a bit long, `value` seems confusingly generic and only understandable to people with FP/theory background.
- Originally I wanted to just set `__setattr__` at the end of `__init__` but that doesn’t seem to work unfortunately.  So `object.setattr` it is.

***

I’m not set on any of those decisions and will happily be converted to any other way.  Just please keep in mind, that the feature has to be discoverable and understandable to people who just started programming.

Please paint this bikeshed.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -466,6 +466,21 @@ I guess that means Clojure can be shut down now, sorry Rich!
    >>> i1 == i2
    False
 
+If you're still not convinced, that Python + ``attrs`` is the better Clojure, maybe immutable classes can change your mind:
+
+.. doctest::
+
+   >>> @attr.s(frozen=True)
+   ... class C(object):
+   ...     x = attr.ib()
+   >>> i = C(1)
+   >>> i.x = 2
+   Traceback (most recent call last):
+      ...
+   attr.exceptions.FrozenInstanceError: can't set attribute
+   >>> i.x
+   1
+
 Sometimes you may want to create a class programmatically.
 ``attrs`` won't let you down and gives you :func:`attr.make_class` :
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -466,7 +466,7 @@ I guess that means Clojure can be shut down now, sorry Rich!
    >>> i1 == i2
    False
 
-If you're still not convinced, that Python + ``attrs`` is the better Clojure, maybe immutable classes can change your mind:
+If you're still not convinced that Python + ``attrs`` is the better Clojure, maybe immutable-ish classes can change your mind:
 
 .. doctest::
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -397,10 +397,17 @@ def _add_init(cls, frozen):
     locs = {}
     bytecode = compile(script, unique_filename, "exec")
     attr_dict = dict((a.name, a) for a in attrs)
-    exec_(bytecode, {"NOTHING": NOTHING,
-                     "attr_dict": attr_dict,
-                     "validate": validate,
-                     "_convert": _convert}, locs)
+    globs = {
+        "NOTHING": NOTHING,
+        "attr_dict": attr_dict,
+        "validate": validate,
+        "_convert": _convert
+    }
+    if frozen is True:
+        # Save the lookup overhead in __init__ if we need to circumvent
+        # immutability.
+        globs["_setattr"] = object.__setattr__
+    exec_(bytecode, globs, locs)
     init = locs["__init__"]
 
     # In order of debuggers like PDB being able to step through the code,
@@ -470,11 +477,11 @@ def _attrs_to_script(attrs, frozen):
     Return a valid Python script of an initializer for *attrs*.
 
     If *frozen* is True, we cannot set the attributes directly so we use
-    ``object.__setattr__``.
+    a cached ``object.__setattr__``.
     """
     if frozen is True:
         def fmt_setter(attr_name, value):
-            return "object.__setattr__(self, '%(attr_name)s', %(value)s)" % {
+            return "_setattr(self, '%(attr_name)s', %(value)s)" % {
                 "attr_name": attr_name,
                 "value": value,
             }
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -150,7 +150,7 @@ def _frozen_setattrs(self, name, value):
     """
     Attached to frozen classes as __setattr__.
     """
-    raise FrozenInstanceError("can't set attribute")
+    raise FrozenInstanceError()
 
 
 def attributes(maybe_cls=None, these=None, repr_ns=None,
@@ -572,7 +572,7 @@ class Attribute(object):
                     raise TypeError("Missing argument '{arg}'.".format(arg=a))
 
     def __setattr__(self, name, value):
-        raise AttributeError("can't set attribute")  # To mirror namedtuple.
+        raise FrozenInstanceError()
 
     @classmethod
     def from_counting_attr(cls, name, ca):
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -414,7 +414,7 @@ def _add_init(cls, frozen):
     if frozen is True:
         # Save the lookup overhead in __init__ if we need to circumvent
         # immutability.
-        globs["_setattr"] = object.__setattr__
+        globs["_cached_setattr"] = object.__setattr__
     exec_(bytecode, globs, locs)
     init = locs["__init__"]
 
@@ -487,9 +487,14 @@ def _attrs_to_script(attrs, frozen):
     If *frozen* is True, we cannot set the attributes directly so we use
     a cached ``object.__setattr__``.
     """
+    lines = []
     if frozen is True:
+        lines.append(
+            "_setattr = _cached_setattr.__get__(self, self.__class__)"
+        )
+
         def fmt_setter(attr_name, value):
-            return "_setattr(self, '%(attr_name)s', %(value)s)" % {
+            return "_setattr('%(attr_name)s', %(value)s)" % {
                 "attr_name": attr_name,
                 "value": value,
             }
@@ -500,7 +505,6 @@ def _attrs_to_script(attrs, frozen):
                 "value": value,
             }
 
-    lines = []
     args = []
     has_validator = False
     has_convert = False
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -188,10 +188,18 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
         memory-efficient.  See :ref:`slots` for further ramifications.
     :param bool frozen: Make instances immutable after initialization.  If
         someone attempts to modify a frozen instance,
-        :exc:`attr.exceptions.FrozenInstanceError` is raised.  Please note that
-        this is achieved by installing a custom ``__setattr__`` method on your
-        class so you can't implement an own one.  Please note also that true
-        immutability is impossible in Python.
+        :exc:`attr.exceptions.FrozenInstanceError` is raised.
+
+        Please note:
+
+            1. This is achieved by installing a custom ``__setattr__`` method
+               on your class so you can't implement an own one.
+
+            2. True immutability is impossible in Python.
+
+            3. This *does* have a minor a runtime performance impact when
+               initializing new instances.  In other words: ``__init__`` is
+               slightly slower with ``frozen=True``.
 
         ..  _slots: https://docs.python.org/3.5/reference/datamodel.html#slots
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -3,8 +3,9 @@ from __future__ import absolute_import, division, print_function
 import hashlib
 import linecache
 
-from ._compat import exec_, iteritems, isclass, iterkeys
 from . import _config
+from ._compat import exec_, iteritems, isclass, iterkeys
+from .exceptions import FrozenInstanceError
 
 
 class _Nothing(object):
@@ -145,8 +146,16 @@ def _transform_attrs(cls, these):
             had_default = True
 
 
+def _frozen_setattrs(self, name, value):
+    """
+    Attached to frozen classes as __setattr__.
+    """
+    raise FrozenInstanceError("can't set attribute")
+
+
 def attributes(maybe_cls=None, these=None, repr_ns=None,
-               repr=True, cmp=True, hash=True, init=True, slots=False):
+               repr=True, cmp=True, hash=True, init=True,
+               slots=False, frozen=False):
     """
     A class decorator that adds `dunder
     <https://wiki.python.org/moin/DunderAlias>`_\ -methods according to the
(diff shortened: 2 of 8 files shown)


Tinche:
Cool :)

Some quick opinions:
* `slots` should default to `True` for frozen classes. The presence of a `__dict__` will be useless in normal use anyway. If someone absolutely, positively must change a field (maybe for testing), there's the `object.__setattr__` trick.
* I agree with @glyph in #3 that hashes should be somehow significant. The presence of a hash function is the closest thing Python has to an immutability marker, right? If a class is frozen, I'd consider asserting the presence of `__hash__()` on all provided arguments in `__init__` (maybe just assert it's present and not call it directly to save some time).

You might not like this, but I'd also consider defaulting `hash` to `False` for non-frozen classes. Looking at the code base I'm working on at work, I have 44 attrs classes defined. All of them define `__hash__()` (since I went with the default `hash` setting), but virtually none of them are actually safe to hash since they either contain dicts or contain classes that contain dicts. And this fact reveals itself at hashing time, not creation time. :) *grumbles about the absence of frozendict*

I think it'd be valuable for `frozen` to be more opinionated and guarantee instances can be hashed?

hynek:
1. we cannot change the default of an argument like `hash`, that would break too much stuff :|  I usually just exclude mutable attributes from hash *cough*.
2. due to the lack of a frozendict, ISTM that enforcing the hashability of all attributes renders this feature more or less worthless?
3. And finally a busy @alex wanted to be reminded in 11h. ;)

docs/examples.rst
```diff
@@ -466,6 +466,21 @@ I guess that means Clojure can be shut down now, sorry Rich!
    >>> i1 == i2
    False
 
+If you're still not convinced, that Python + ``attrs`` is the better Clojure, maybe immutable classes can change your mind:
+
+.. doctest::
+
+   >>> @attr.s(frozen=True)
```
sigmavirus24: Since things are not quite immutable in Python, have you considered `frozen='ish'`. It's only an extra character ;)

Tinche: Also remove the comma after "convinced"?

Tinche:
Maybe we can have our cake and eat it too? Leave `attr.s()` as currently implemented and think of it as a low-level, pick-your-exact-settings building block, and introduce @glyph's `attr.object()` and `attr.value()` as opinionated, higher level decorators?

`attr.value()` could just force `slots` and `frozen` to True, default `hash` to True, and insert a check before `__init__` to assert hashes on all arguments. Basically we have the building blocks now to do the first post from #3.

glyph:
@Tinche I think if we're going to do that, we don't need to discuss it here - this PR stands alone, and would be necessary for that naming strategy anyway :)

src/attr/exceptions.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import absolute_import, division, print_function
+
+
+class FrozenInstanceError(AttributeError):
```
Tinche: Raise this at https://github.com/hynek/attrs/pull/60/files#diff-21756223332b04173cec8910fda6655bR576 instead of AttributeError?

Tinche:
@glyph If we're going to do that, I'm :+1: on this.

Tinche:
I'm assuming it'd be a compatibility break, and thus a no-no, to make Factory and the validators frozen and slotted?

hynek:
Factory was always an opaque object so slotted shouldn’t be a problem.  But frozen probably has a runtime impact so unless you can disprove that using benchmarks, I’m not making attrs slower for rather esoteric reasons.

Tinche:
Hm, testing on my work Mac, CPython 3.5:

```
$ pip install -U git+https://github.com/haypo/perf.git#egg=perf
$ # Non-slotted, non-frozen
$ pyperf timeit -g -s "from attr import Factory; l = lambda: 1" "Factory(l)"
....................
402 ns: 10 ############################################
405 ns: 18 ###############################################################################
408 ns:  8 ###################################
410 ns:  9 ########################################
413 ns:  3 #############
416 ns:  4 ##################
418 ns:  2 #########
421 ns:  0 |
424 ns:  1 ####
427 ns:  1 ####
429 ns:  1 ####
432 ns:  0 |
435 ns:  0 |
437 ns:  0 |
440 ns:  0 |
443 ns:  1 ####
445 ns:  0 |
448 ns:  1 ####
451 ns:  0 |
454 ns:  0 |
456 ns:  1 ####

Median +- std dev: 409 ns +- 11 ns
$ Slotted, frozen
$ pyperf timeit -g -s "from attr import Factory; l = lambda: 1" "Factory(l)"
....................
476 ns: 1 ##########
478 ns: 2 ####################
479 ns: 8 ###############################################################################
481 ns: 2 ####################
483 ns: 0 |
484 ns: 8 ###############################################################################
486 ns: 7 #####################################################################
488 ns: 2 ####################
490 ns: 5 #################################################
491 ns: 5 #################################################
493 ns: 5 #################################################
495 ns: 5 #################################################
496 ns: 4 ########################################
498 ns: 1 ##########
500 ns: 2 ####################
501 ns: 0 |
503 ns: 0 |
505 ns: 0 |
507 ns: 1 ##########
508 ns: 1 ##########
510 ns: 1 ##########

Median +- std dev: 489 ns +- 8 ns
$ # Slotted, non-frozen
$ pyperf timeit -g -s "from attr import Factory; l = lambda: 1" "Factory(l)"
....................
294 ns: 5 #################################################
295 ns: 3 ##############################
296 ns: 8 ###############################################################################
297 ns: 7 #####################################################################
298 ns: 5 #################################################
299 ns: 6 ###########################################################
300 ns: 7 #####################################################################
301 ns: 4 ########################################
302 ns: 7 #####################################################################
303 ns: 0 |
304 ns: 1 ##########
305 ns: 2 ####################
306 ns: 0 |
307 ns: 0 |
308 ns: 1 ##########
309 ns: 1 ##########
310 ns: 0 |
311 ns: 1 ##########
312 ns: 1 ##########
313 ns: 0 |
314 ns: 1 ##########

Median +- std dev: 300 ns +- 4 ns
```

So it would appear there is a slight performance penalty to frozen, but a performance benefit to slots, at least on CPython. We're talking hundreds of nanoseconds, however. Maybe we could optimize the `object.__setattr__` part somehow? We're doing an unnecessary lookup (`object.__setattr__) we could be doing during class init, and inject it into globals. With this change:

```
# Slotted, frozen with the optimization
$ pyperf timeit -g -s "from attr import Factory; l = lambda: 1" "Factory(l)"
....................
446 ns:  3 ###############
447 ns:  3 ###############
449 ns:  6 ##############################
450 ns: 16 ###############################################################################
451 ns:  3 ###############
452 ns:  3 ###############
454 ns:  6 ##############################
455 ns:  5 #########################
456 ns:  2 ##########
457 ns:  3 ###############
459 ns:  2 ##########
460 ns:  1 #####
461 ns:  2 ##########
462 ns:  0 |
464 ns:  1 #####
465 ns:  1 #####
466 ns:  0 |
467 ns:  0 |
469 ns:  1 #####
470 ns:  1 #####
471 ns:  1 #####

Median +- std dev: 452 ns +- 6 ns
```

Unfortunately, the perf tool doesn't work on PyPy yet (https://github.com/haypo/perf/issues/12) so this is all the data I have right now.

hynek:
I’ve tried caching too like this:

```diff
diff --git a/src/attr/_make.py b/src/attr/_make.py
index b381b09..9a37b10 100644
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -398,10 +398,17 @@ def _add_init(cls, frozen):
     locs = {}
     bytecode = compile(script, unique_filename, "exec")
     attr_dict = dict((a.name, a) for a in attrs)
-    exec_(bytecode, {"NOTHING": NOTHING,
-                     "attr_dict": attr_dict,
-                     "validate": validate,
-                     "_convert": _convert}, locs)
+    globs = {
+        "NOTHING": NOTHING,
+        "attr_dict": attr_dict,
+        "validate": validate,
+        "_convert": _convert
+    }
+    if frozen is True:
+        # Save the lookup overhead in __init__ if we need to circumvent
+        # immutability.
+        globs["_setattr"] = object.__setattr__
+    exec_(bytecode, globs, locs)
     init = locs["__init__"]

     # In order of debuggers like PDB being able to step through the code,
@@ -471,11 +478,11 @@ def _attrs_to_script(attrs, frozen):
     Return a valid Python script of an initializer for *attrs*.

     If *frozen* is True, we cannot set the attributes directly so we use
-    ``object.__setattr__``.
+    a cached ``object.__setattr__``.
     """
     if frozen is True:
         def fmt_setter(attr_name, value):
-            return "object.__setattr__(self, '%(attr_name)s', %(value)s)" % {
+            return "_setattr(self, '%(attr_name)s', %(value)s)" % {
                 "attr_name": attr_name,
                 "value": value,
             }
```

and I couldn’t find any significant or even consistent performance advantage (MBP late 2013) even with 6 arguments:

```
➤ pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', ['x', 'y', 'z', 'a', 'b', 'c'])" "C(1,2,3,4,5,6)" # optimized
........................................
716 ns:  1 #####
722 ns:  2 ##########
729 ns:  5 ########################
736 ns: 14 ####################################################################
742 ns: 10 #################################################
749 ns: 11 #####################################################
756 ns:  9 ############################################
763 ns: 13 ###############################################################
769 ns: 11 #####################################################
776 ns: 12 ##########################################################
783 ns:  6 #############################
790 ns:  8 #######################################
796 ns:  6 #############################
803 ns:  4 ###################
810 ns:  1 #####
817 ns:  2 ##########
823 ns:  2 ##########
830 ns:  0 |
837 ns:  1 #####
844 ns:  1 #####
850 ns:  1 #####

Median +- std dev: 766 ns +- 27 ns

➤ pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', ['x', 'y', 'z', 'a', 'b', 'c'])" "C(1,2,3,4,5,6)" # old
........................................
712 ns:  2 #########
720 ns:  1 ####
728 ns: 10 ###########################################
736 ns:  8 ##################################
744 ns: 15 ################################################################
752 ns: 16 ####################################################################
760 ns: 15 ################################################################
768 ns: 13 #######################################################
776 ns: 11 ###############################################
784 ns: 11 ###############################################
792 ns:  9 ######################################
800 ns:  1 ####
808 ns:  0 |
816 ns:  1 ####
824 ns:  3 #############
832 ns:  1 ####
840 ns:  1 ####
848 ns:  0 |
856 ns:  1 ####
864 ns:  0 |
872 ns:  1 ####

Median +- std dev: 764 ns +- 28 ns
```

Am I doing something wrong?

ISTM that we’re trying to be too clever here and Python does quite a bit of optimizations that we may even be sabotaging.

Tinche:
The classes you're testing aren't frozen?

hynek:
lol dumbass

OK second try:

```
➤ pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', ['x', 'y', 'z', 'a', 'b', 'c'], frozen=True)" "C(1,2,3,4,5,6)" # old
........................................
1.61 us:  1 #####
1.63 us:  3 ###############
1.65 us:  7 ####################################
1.67 us:  7 ####################################
1.69 us: 13 ###################################################################
1.71 us:  9 ##############################################
1.72 us: 13 ###################################################################
1.74 us: 10 ####################################################
1.76 us: 11 #########################################################
1.78 us:  6 ###############################
1.80 us:  8 #########################################
1.82 us:  9 ##############################################
1.84 us:  7 ####################################
1.85 us:  6 ###############################
1.87 us:  1 #####
1.89 us:  1 #####
1.91 us:  3 ###############
1.93 us:  2 ##########
1.95 us:  1 #####
1.96 us:  0 |
1.98 us:  2 ##########

Median +- std dev: 1.76 us +- 0.08 us

➤ pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', ['x', 'y', 'z', 'a', 'b', 'c'], frozen=True)" "C(1,2,3,4,5,6)" # new
........................................
1.46 us:  1 ####
1.47 us:  3 ###########
1.48 us:  0 |
1.50 us:  5 ##################
1.51 us:  9 ################################
1.52 us: 10 ###################################
1.53 us: 19 ###################################################################
1.55 us: 10 ###################################
1.56 us: 12 ##########################################
1.57 us:  6 #####################
1.59 us:  7 #########################
1.60 us:  7 #########################
1.61 us: 10 ###################################
1.63 us:  4 ##############
1.64 us:  4 ##############
1.65 us:  5 ##################
1.66 us:  4 ##############
1.68 us:  2 #######
1.69 us:  0 |
1.70 us:  1 ####
1.72 us:  1 ####

Median +- std dev: 1.57 us +- 0.05 us
```

Holy shit that’s slow in both cases. 😱  More than 2x slower…I know why I’m not gonna use it. :)

***

As for factories:  making them slotted is great and fine with me but frozen?  Nah, not worth it.

hynek:
I’m kind of still waiting for general approval/feedback btw.

As I’ve made clear, this feature is *not* for me.  I'm not gonna make my code artificially slower just to get sort-of-immutability.  I just don’t mutate. ;)

***

What I’m saying is, that *someone* has to be happy, so I can’t merge this until things are clear to me, that y’all have what y’all want.

Insoleet:
I wonder why immutability is necessary applied to all attributes of a class ? Couldn't immutability only be applied to desired attributes ?

Tinche:
@Insoleet it's just a design choice we've agreed on.

A question for someone more knowledgeable than I (i.e. other people in the conversation): is there a dirtier, faster hack we could be using instead?

hynek:
Circumventing the descriptor gave us some more vroom vroom:

```
➤ pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], frozen=True)" "C(1,2,3)" # old
........................................
763 ns: 28 ####################################################################
767 ns: 16 #######################################
772 ns:  5 ############
777 ns:  5 ############
781 ns:  6 ###############
786 ns:  5 ############
791 ns:  6 ###############
795 ns: 11 ###########################
800 ns: 11 ###########################
804 ns:  3 #######
809 ns:  4 ##########
814 ns:  2 #####
818 ns:  4 ##########
823 ns:  4 ##########
828 ns:  4 ##########
832 ns:  1 ##
837 ns:  2 #####
842 ns:  0 |
846 ns:  0 |
851 ns:  1 ##
856 ns:  2 #####

Median +- std dev: 785 ns +- 24 ns
```

vs.

```
➤ pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], frozen=True)" "C(1,2,3)" # more new
........................................
684 ns:  1 #####
688 ns:  7 ################################
693 ns:  5 #######################
697 ns:  1 #####
701 ns:  7 ################################
705 ns: 13 ###########################################################
709 ns: 15 ####################################################################
713 ns:  9 #########################################
717 ns: 11 ##################################################
721 ns: 11 ##################################################
725 ns: 10 #############################################
729 ns:  7 ################################
733 ns: 10 #############################################
737 ns:  5 #######################
741 ns:  3 ##############
745 ns:  2 #########
749 ns:  0 |
753 ns:  1 #####
757 ns:  1 #####
761 ns:  0 |
765 ns:  1 #####

Median +- std dev: 718 ns +- 16 ns
```

But I’m pretty sure we’re arriving at a point of diminishing returns.

hynek:
OK everyone, last call for conceptual feedback/vetoes?

src/attr/_make.py
```diff
@@ -487,9 +487,14 @@ def _attrs_to_script(attrs, frozen):
     If *frozen* is True, we cannot set the attributes directly so we use
     a cached ``object.__setattr__``.
     """
+    lines = []
     if frozen is True:
+        lines.append(
+            "_setattr = _cached_setattr.__get__(self, self.__class__)"
```
sigmavirus24: Would it be useful to add a comment explaining what exactly this does?

hynek: Is 

```python
            # Circumvent the __setattr__ descriptor to save one lookup per
            # assignment.
```
accurate @Tinche?

Tinche: :+1:

Tinche:
Personally I think this is useful and should go in. Maybe I won't use it for classes that get instantiated a lot in inner loops, but for long-lived classes where you'd like to avoid deepcopying frozen is surely the way to go. (Think of `attr` attributes, where the deepcopy totally dominates the performance of `fields()`.)

Maybe add a note to the docs that a frozen `__init__` will be slightly slower than a non-frozen one? Also, a slotted `__init__` turns out to be faster than non-slotted. On CPython 3.5.

hynek:
You mean like https://github.com/hynek/attrs/pull/60/files#diff-21756223332b04173cec8910fda6655bR200 ? :)

PR #61, opened 2016-08-17
Optimize init for frozen classes.

Tinche:
Should make frozen __init__ a little faster.


hynek:
Uh sorry this PR must have come in when I was busy implementing the same thing, sorry. %)

PR #66, opened 2016-08-19
Added names function which returns attr names for an attrs class.

kissgyorgy:


hynek:
What would be the concrete use case of this function?

Tinche:
Appreciate the thought and effort, but it's kind of a one-liner.

hynek:
I have to agree with Tinche; thank you but sorry I’m trying to keep the API as lean as humanly possible.

kissgyorgy:
> What would be the concrete use case of this function?

I needed the attribute names of the class. I had two choices:
- use `C.__attrs_attrs__` which is ugly and private
- use `[a.name for a in attr.fields(C)]` which is too verbose IMO, and I need to know about `Attribute` objects and what attributes they have

hynek:
I’d still like to know for what that’s useful?  I’m asking because if it’s something super common, I could  live with it.  If it’s rather obscure, you’ll have to live an own function. :)

PR #71, opened 2016-08-22
asdict(): Optionally retain collection types

### LINKED ISSUE #69: asdict(): Optionally retain collection types
When calling `attr.asdict()`, it converts attributes of type `tuple`, `set` to type `list` and subclasses of `dict` to `dict`. This is being done for JSON compatibility, however, there are use cases where it's required to preserve the original type.

Therefore I request an optional argument called `retain_collection_types`, which is set to `False` by default (to not break the existing API). If enabled, the function should not convert any collection type.

vortec:
Fixes #69

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -6,7 +6,8 @@ from ._compat import iteritems
 from ._make import Attribute, NOTHING, fields
 
 
-def asdict(inst, recurse=True, filter=None, dict_factory=dict):
+def asdict(inst, recurse=True, filter=None, dict_factory=dict,
+           retain_collection_types=False):
     """
     Return the ``attrs`` attribute values of *inst* as a dict. Optionally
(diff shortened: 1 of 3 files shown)


src/attr/_funcs.py
```diff
@@ -25,6 +26,11 @@ def asdict(inst, recurse=True, filter=None, dict_factory=dict):
         example, to produce ordered dictionaries instead of normal Python
         dictionaries, pass in ``collections.OrderedDict``.
 
+    :param bool retain_collection_types: Do not convert to ``list`` when
+        encountering an attribute which is type ``tuple`` or ``set``.
+        ``recurse`` must also be ``True``.
```
hynek: I think this is misleading.   More like "Only meaningful if ``recurse`` is ``True``."?

vortec:
Updated.

hynek:
Merged in  c6fbec9 – thanks!

PR #73, opened 2016-08-28
Add a philosophy section

hynek:
Also expand the example class a little bit to stress that we're not just
about data containers.

This is kind of a response to questions and concerns that come up in issues – and places that shall not be named –  frequently.

I’d some love feedback from people who *understand* `attrs` and what it stands for.   Ideally also from people who have better words than I.

I really just want to link there and drop the mike in the future.

cc  @hawkowl @glyph @cyli @econchick @Tinche @cournape @mithrandi

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,9 +33,14 @@ For that, it gives you a class decorator and a way to declaratively define the a
    ... class C(object):
    ...     x = attr.ib(default=42)
    ...     y = attr.ib(default=attr.Factory(list))
+   ...
+   ...     def hard_math(self, z):
+   ...         return self.x * self.y * z
    >>> i = C(x=1, y=2)
    >>> i
    C(x=1, y=2)
+   >>> i.hard_math(3)
+   6
    >>> i == C(1, 2)
    True
    >>> i != C(2, 1)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -105,6 +105,8 @@ Core
 .. autoexception:: attr.exceptions.FrozenInstanceError
 
 
+.. _helpers:
+
 Helpers
 -------
 
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -9,6 +9,39 @@ In order to fullfil its ambitious goal of bringing back the joy to writing class
    :end-before: -testimonials-
 
 
+.. _philosophy:
+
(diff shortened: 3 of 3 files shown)


hawkowl:
👍

Tinche:
If by 'places that shall not be named' you mean Hacker News like was mentioned on Twitter, I don't go there so I don't know exactly what their complaints are.

I use attrs a lot for what you call 'data-only' containers (Java folks would call these DTOs I guess, or data transfer objects), maybe with some properties or helper methods mixed in. You make a point of attrs not being only for this kinds of classes, but I find 1) attrs works best for these kinds of classes and 2) it's actually in your best interest to have many of these kinds of classes. So I like the fact attrs is opinionated (as much as it is) and I like its opinions :) I'm not sure how to turn this paragraph into an actionable item for the philosophy section :)

Another thing I love about attrs is that it's declarative. It's a damn shame we've had to wait this long to get proper declarative classes into Python (with an API I actually want and like to use). An anecdote: I work in mobile games. We use Python on the backend and, nowadays, C# for the games themselves. There is data that's shared between the client and the server, things like request and response formats, the player save game, and game data itself (every game has a database of domain-specific data, for example a cooking game might have a list of food, ingredients, appliances, restaurant furniture etc). Recently I've started using attrs for some of these classes on the backend, where before they used to be pure dictionaries. Now that we have this data in a declarative format in code, I'm seriously considering generating C# classes from them, so we'd have to define our data only once and save a bunch of manual labor. attrs makes this possible and easy.

Maybe another thing we could touch on is structured data vs unstructured data, and how attrs classes are superior to dictionaries. For example, when you turn your unstructured data into classes, there's validation at the point of conversion instead of point of use, which is exactly what you want. Type checking becomes possible (I don't know how well mypy handles attrs, but PyCharm handles them just fine with extra type hints, and I rely on this daily, for correctness and autocomplete).

I'm not sure this is material for the philosophy section, more for the "why" section. My philosophy would be 'use attrs for what it's great for, and reap the massive benefits it brings, even if you have to change your thinking a little'.

hynek:
I’d wish to expand the “why not dicts” section but I’m tired and out of words.  if someone could write something coherent, that would be really greeeeat. ;)

PR #74, opened 2016-08-31
add attr docstring

rdpate:
The generated docstring merely points to the docs elsewhere.  I started with "{} <{}>".format, but I'm new to attrs and wasn't sure if that meets the requirement for all supported Python versions.

### DIFF
--- a/src/attr/__init__.py
+++ b/src/attr/__init__.py
@@ -29,6 +29,7 @@ __version__ = "16.2.0.dev0"
 __title__ = "attrs"
 __description__ = "Attributes Without Boilerplate"
 __uri__ = "https://attrs.readthedocs.io/"
+__doc__ = __description__ + " <" + __uri__ + ">"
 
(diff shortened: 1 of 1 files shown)


rdpate:
Are supported versions clearly written anywhere? I see "rigorously tested on Python 2.7, 3.4+, and PyPy" in the readme -- does this imply 2.6/3.3 are not supported, or might work but not tested (with unannounced breakage, ...)?

Tinche:
Correct, 2.6 and 3.3 are not supported: https://github.com/hynek/attrs/blob/master/CHANGELOG.rst#1600-2016-05-23

We don't test on them so I have no idea if they will work or not (3.3 will probably work), and if they don't work we basically don't care.

hynek:
Heh I can stress as much as I want that small contributions are welcome and people will still apologize for them. :)  Thank you!

PR #75, opened 2016-09-02
Add converters helper module

adevore:
This change adds `list_of`, `set_of`, `frozenset_of`, and `from_dict` in a new converters module. I found similar helper functions useful when converting from JSON with nested objects and arrays of objects.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -257,6 +257,78 @@ Validators
       C(x=None)
 
 
+.. _api_converters:
+
+Converters
+----------
+
+.. autofunction:: attr.converters.list_of
+
+   For example:
+
+   .. doctest::
+
+      >>> @attr.s
+      ... class C(object):
+      ...     x = attr.ib()
+      >>> @attr.s
+      ... class D(object):
+      ...     y = attr.ib(convert=attr.converters.list_of(C))
+      >>> data = {'y': [{'x': 3}, {'x': 4}] }
+      >>> D(**data)
+      D(y=[C(x=3), C(x=4)])
+
+.. autofunction:: attr.converters.set_of
+
+   For example:
+
+   .. doctest::
+
+      >>> @attr.s
+      ... class C(object):
+      ...     x = attr.ib()
+      >>> @attr.s
+      ... class D(object):
+      ...     y = attr.ib(convert=attr.converters.set_of(C))
+      >>> data = {'y': [{'x': 3}, {'x': 4}] }
+      >>> D(**data)
+      D(y={C(x=3), C(x=4)})
+
+
+.. autofunction:: attr.converters.frozenset_of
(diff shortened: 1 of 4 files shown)


Tinche:
Having worked (and still working) on a complex code base (~50 attrs classes), I've needed something like this too, and written it. 

My conclusion is converting unstructured data to and from attrs is better served by a separate component/library, rather than baking this into your own classes, and is what I'll do for the next project. It still might be useful for attrs to have a few good converters bundled, so I'm not generally against including this.

Your example:
```
>>> @attr.s
... class C(object):
...     x = attr.ib()
>>> @attr.s
... class D(object):
...     y = attr.ib(convert=attr.converters.from_dict(C))
>>> data = {'y': {'x': 3} }
>>> D(**data)
D(y=C(x=3))
```
generates an init that only works for dictionaries. So, you can't do (if I'm reading this correctly):
```
>>> D(y=C(x=3))  # will raise an exception
```
which is unfortunate.

Our version is:
```
def ensure_cls(cls):
    """If the attribute is an instance of cls, pass, else try constructing."""
    def converter(val):
        if isinstance(val, cls):
            return val
        else:
            return cls(**val)
    return converter

@attr.s
class D(object):
    y = attr.ib(convert=ensure_cls(C))
```

This is more useful, in my experience. The same goes for collections. I've found `ensure_enum` is very useful as well.

adevore:
That makes sense. I'll make the modifications. Also, perhaps tuple_of would be a good idea for classes that will be used in a hashable context?

adevore:
I have made the necessary modifications. As part of the change, I consolidated the container conversion classes into a single generic class that works with any container whose constructor takes an iterable.

Tinche:
Just thinking about this a little more, can we make these composable? And if we can, is it worth it?

Example:
```
class A:
    x = attr.ib(convert=list_of(from_dict(B)))
```

Maybe this would make more sense with different converter names.

adevore:
I was also doing a little more thinking. Maybe it would be better to have a single function that works with any collection that takes an iterable as an argument, like so:

```
class A:
    x = attr.ib(convert=collect(list, from_dict(B)))
```

That provides more flexibility for custom collections and cuts down on API size.

Lucas-C:
What's this PR status ? Is it ol to be merged ?

Lucas-C:
I think there is some history behind this: https://github.com/hynek/attrs/search?q=list_of&type=Issues&utf8=%E2%9C%93
Looks like https://github.com/Tinche/cattrs is the way to go

hynek:
Yeah sadly as of now I haven't found a way to do this that makes me happy.  Maybe something practical will come out of cattrs or maybe they'll stay separated forever.  Hard to say ATM.

PR #78, opened 2016-09-06
Implement astuple method

Insoleet:
Please checkout my proposal https://github.com/hynek/attrs/issues/77. Do not hesitate to comment this code.
The parameter `tuple_factory` is only tested with the `tuple` type. If you think this should be tested with other factories, do not hesitate to tell me.

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -63,7 +63,8 @@ def asdict(inst, recurse=True, filter=None, dict_factory=dict,
     return rv
 
 
-def astuple(inst, recurse=True, filter=None, tuple_factory=tuple):
+def astuple(inst, recurse=True, filter=None, tuple_factory=tuple,
+           retain_collection_types=False):
     """
     Return the ``attrs`` attribute values of *inst* as a tuple.
 
@@ -78,6 +79,9 @@ def astuple(inst, recurse=True, filter=None, tuple_factory=tuple):
         value as the second argument.
     :param callable tuple_factory: A callable to produce tuples from.  For
         example, to produce ordered tuples instead of normal Python tuples.
+    :param bool retain_collection_types: Do not convert to ``list`` when
+        encountering an attribute which is type ``tuple`` or ``set``.  Only
+        meaningful if ``recurse`` is ``True``.
 
     :rtype: return type of *tuple_factory*
     """
@@ -92,7 +96,7 @@ def astuple(inst, recurse=True, filter=None, tuple_factory=tuple):
                 rv.append(astuple(v, recurse=True, filter=filter,
                                     tuple_factory=tuple_factory))
             elif isinstance(v, (tuple, list, set)):
-                cf = list
+                cf = v.__class__ if retain_collection_types is True else list
                 rv.append(cf([
                     astuple(j, recurse=True, filter=filter,
                            tuple_factory=tuple_factory)
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -64,7 +64,7 @@ def asdict(inst, recurse=True, filter=None, dict_factory=dict,
 
 
 def astuple(inst, recurse=True, filter=None, tuple_factory=tuple,
-           retain_collection_types=False):
+            retain_collection_types=False):
     """
     Return the ``attrs`` attribute values of *inst* as a tuple.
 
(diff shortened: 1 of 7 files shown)


tests/test_funcs.py
```diff
@@ -155,6 +157,100 @@ def test_asdict_preserve_order(self, cls):
         assert [a.name for a in fields(cls)] == list(dict_instance.keys())
 
 
+class TestAsTuple(object):
+    """
+    Tests for `asdict`.
+    """
+    @given(st.sampled_from(IMMUTABLE_SEQ_TYPES))
```
Tinche: You can use SEQUENCE_TYPES here, a list is a valid tuple_factory and something that might be commonly used.

src/attr/_funcs.py
```diff
@@ -63,6 +63,67 @@ def asdict(inst, recurse=True, filter=None, dict_factory=dict,
...
+                                vv,
+                                tuple_factory=tuple_factory
+                            ) if has(vv.__class__) else vv
+                        )
+                        for kk, vv in iteritems(v)))
+            else:
+                rv.append(v)
+        else:
+            rv.append(v)
+    return tuple_factory(rv)
```
Tinche: If we think ``list`` might be a common tuple_factory, we can check here and not do extra work, since rv is already a list.

Something like
```
return rv if tuple_factory is list else tuple_factory(rv)
```

Tinche:
@hynek How do we feel about `filter` in astuple? My gut feeling is a little iffy since if you skip an element in the tuple, it fundamentally changes the meaning of the tuple. But maybe that's what you want sometimes.

Tinche:
Also I'm thinking we need either a dict_factory parameter or to have retain_collection_types to apply to dicts too, since right now astuple will map all OrderedDicts to dicts and there should be a way to avoid that.

I think just having ``retain_collection_types`` apply in the ``elif isinstance(v, dict):`` branch  would be enough. (Use v.__class__ as the dict factory.)

hynek:
I very much think that `filter` makes sense.  Especially in the use case with DB you may want to keep stuff out.  Or drop a password field or something.

Tinche:
Yeah, let's leave it in. I was actually thinking about passwords in the first place. But the exact scenario I was worried about is impossible since no mandatory attributes are allowed after an attribute with a default value, so it's moot anyway.

Insoleet:
I took all your comments into account.

src/attr/_funcs.py
```diff
@@ -63,6 +63,67 @@ def asdict(inst, recurse=True, filter=None, dict_factory=dict,
...
+
+    :param inst: Instance of an ``attrs``-decorated class.
+    :param bool recurse: Recurse into classes that are also
+        ``attrs``-decorated.
+    :param callable filter: A callable whose return code determines whether an
+        attribute or element is included (``True``) or dropped (``False``).  Is
+        called with the :class:`attr.Attribute` as the first argument and the
+        value as the second argument.
+    :param callable tuple_factory: A callable to produce tuples from.  For
+        example, to produce ordered tuples instead of normal Python tuples.
```
Tinche: "to produce ordered tuples instead of Python tuples"? :D

Maybe change it to "to produce lists instead of tuples"

tests/test_funcs.py
```diff
@@ -155,6 +156,115 @@ def test_asdict_preserve_order(self, cls):
         assert [a.name for a in fields(cls)] == list(dict_instance.keys())
 
 
+class TestAsTuple(object):
+    """
+    Tests for `asdict`.
```
Tinche: astuple

Tinche:
Alright, we're getting close for Hynek to take a look. :)
Please add a changelog entry too, you can find examples in CHANGELOG.rst.

src/attr/_funcs.py
Insoleet: Ahah too many copy pastes ;)

Tinche:
I've noticed `retain_collection_type` isn't propagated always in astuple. I've expanded our Hypothesis nested class strategy so tests can catch this, and written a test. I don't really feel like pasting it all in here, so let me issue a pull request onto your repository.

Tinche:
@hynek I think this is in good shape, please take a look when you have a chance.

hynek:
could you please activate the shiny new maintainer [collaboration feature](https://github.com/blog/2247-improving-collaboration-with-forks)? I don’t want to pester you with minutiae.

Insoleet:
Done !

hynek:
I guess this will need some work due to @Tinche’s work on simple_class?

Tinche:
Yeah, this needs a slight rebase. `simple_classes` should now be `simple_classes()`; i.e. the strategy is now parameterized.

Insoleet:
Ok, I fixed it.

hynek:
Thank you!

PR #80, opened 2016-09-07
Generate __init__ with converters inline.

Tinche:
Here's some more horror. :)

The gains are actually more than I expected.

```
> pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', {'x': attr.ib(convert=str), 'y': attr.ib(convert=float), 'z': attr.ib(), 'a': attr.ib()}, slots=True)" "C(1,2,3,4)"
> # Before:
Median +- std dev: 2.41 us +- 0.06 us
> # After:
Median +- std dev: 1.02 us +- 0.03 us
```
Frozen classes:
```
> pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', {'x': attr.ib(convert=str), 'y': attr.ib(convert=float), 'z': attr.ib(), 'a': attr.ib()}, slots=True, frozen=True)" "C(1,2,3,4)"
> # Before: 
Median +- std dev: 2.94 us +- 0.08 us
> # After: 
Median +- std dev: 1.64 us +- 0.04 us
```
I was actually a little weirded out, but the tests pass and playing around in the repl, it seems to work.

I thought the gains would be greatest with about 50% attributes with converters on them, but no, even with all attributes having converters, I see more than double the speed.
```
> pyperf timeit --rigorous  -g -s "import attr; C = attr.make_class('C', {'x': attr.ib(convert=str), 'y': attr.ib(convert=str), 'z': attr.ib(convert=str), 'a': attr.ib(convert=str)}, slots=True)" "C(1,2,3,4)"
> # Before:
Median +- std dev: 3.41 us +- 0.07 us
> #After:
Median +- std dev: 1.37 us +- 0.04 us
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Changes:
 
 - Converts now work with frozen classes.
   `#76 <https://github.com/hynek/attrs/issues/76>`_
+- Instantiation of ``attrs`` classes with converters is now significantly faster.
+  `#80 <https://github.com/hynek/attrs/pull/80>`_
 
 
 ----
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -9,6 +9,7 @@ from .exceptions import FrozenInstanceError
 
 # This is used at least twice, so cache it here.
 _obj_setattr = object.__setattr__
+_init_convert_pat = '__attr_convert_{}'
 
 
 class _Nothing(object):
@@ -411,8 +412,6 @@ def _add_init(cls, frozen):
     globs.update({
         "NOTHING": NOTHING,
         "attr_dict": attr_dict,
-        "validate": validate,
-        "_convert": _convert
     })
     if frozen is True:
         # Save the lookup overhead in __init__ if we need to circumvent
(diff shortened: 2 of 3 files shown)


hynek:
Omg 🙈

hynek:
Before I have closer look at your abomination, please add a changelog entry. :)

Tinche:
Done.

hynek:
At this point I feel like we should write some integration test for `_attrs_to_script` that goes for some typical use cases and just compares strings afterward.  I’m stopping to trust our tests at this point. %)

Tinche:
Not really sure comparing strings of source code will be a better test than testing the actual behavior. I guess we could assert that the generated init is the most optimal it can be, but we'd also have to take into account some internal stuff, like the __attr_validator_x and __attr_convert_x globals. 
🤔

Tinche:
It's kind of a shame there aren't any libraries for generating Python source code. Or at least I can't find any. Libraries that would expose objects and just let you render to source at the end, instead of glueing strings together.

Maybe I'll write something. 😎

hynek:
I’m not really saying we should do it.  I’m just saying that function haunts me at night. ;)

Issue #79, opened 2016-09-07
FrozenInstanceError with pickle in Spark


mathieulongtin:
Using Apache Spark, with attr 16.1.0, I get this error.

```
Caused by: org.apache.spark.api.python.PythonException: Traceback (most recent call last):
  File ".../spark/python/pyspark/worker.py", line 98, in main
    command = pickleSer._read_with_length(infile)
  File ".../spark/python/pyspark/serializers.py", line 164, in _read_with_length
    return self.loads(obj)
  File ".../spark/python/pyspark/serializers.py", line 422, in loads
    return pickle.loads(obj)
  File ".../lib/python2.7/site-packages/attr/_make.py", line 617, in __setattr__
    raise FrozenInstanceError()
FrozenInstanceError
```

Unfortunately, I can't reproduce it with simple code, for some reason. Using IPython, I managed to track it down to rebuilding `Attributes` class:
```
  /usr/lib64/python2.7/pickle.py(1247)load_build()
   1245         if slotstate:
   1246             for k, v in slotstate.items():
-> 1247                 setattr(inst, k, v)
```

Spark uses a custom pickler in order to support local function and classes, but still uses the standard unpickler.

This was not an issue with 16.0.0, so sticking to that for now.

hynek:
Well because 16.0 didn't have frozen classes? :) Pickle seems to use `setattr` therefore you can't use it with frozen classes.

mathieulongtin:
Actually, my code doesn't use frozen. attr.Attribute uses frozen, and that's what breaks. This exact code is working fine with 16.0.0.

hynek:
Oh drat!

mathieulongtin:
So it turns out Spark's pickle does fancy things to pickle things pickle can't pickle. (say that fast).

Basic pickle couldn't serialize an `attr.Attribute` at all and mumbled something about __getstate__. Sure enough, you can fix the whole thing by adding these methods to Attribute:

```
      def __getstate__(self):
          """Play nice with pickle"""
          return tuple(getattr(self, a) for a in self.__slots__)

      def __setstate__(self, state):
          """Play nice with pickle"""
          __bound_setattr = _obj_setattr.__get__(self, Attribute)
          for name, value in zip(self.__slots__, state):
              __bound_setattr(name, value)
```

And it works with Spark.

Basic test:
```
x = attr.Attribute.from_counting_attr('x', attr.ib())
assert x == pickle.loads(pickle.dumps(x))
```

[...]

PR #81, opened 2016-09-08
Issue79 pickle 2

mathieulongtin:
Fix for issue #79. Add `__getstate__` and `__setstate__` to classes with `slots=True`, including `attr.Attribute`. Added test to make sure objects work with pickle.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -441,6 +441,11 @@ Slot classes are a little different than ordinary, dictionary-backed classes:
 - Since non-slot classes cannot be turned into slot classes after they have been created, ``attr.s(.., slots=True)`` will *replace* the class it is applied to with a copy.
   In almost all cases this isn't a problem, but we mention it for the sake of completeness.
 
+- Using pickle with slot classes require pickle protocol 2 or greater. Python 2 uses protocol 0 by default so the
+  protocol would need to be specified. Python 3 uses protocol 3 by default. You can support protocol 0 and 1 by
+  `creating <https://docs.python.org/2/library/pickle.html#object.__getstate__>`_ a ``__getstate__`` and
+  ``__setstate__`` method. Those methods are created for frozen slot classes because they don't pickle otherwise.
+
 All in all, setting ``slots=True`` is usually a very good idea.
 
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -160,13 +160,12 @@ def _slots_getstate__(obj):
     """Play nice with pickle"""
     return tuple(getattr(obj, a) for a in obj.__slots__)
 
+
 def _slots_setstate__(obj, state):
     """Play nice with pickle"""
     __bound_setattr = _obj_setattr.__get__(obj, Attribute)
     for name, value in zip(obj.__slots__, state):
         __bound_setattr(name, value)
-
-
 
 
 def attributes(maybe_cls=None, these=None, repr_ns=None,
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -234,7 +234,7 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
         if frozen is True:
             cls.__setattr__ = _frozen_setattrs
             if slots is True:
-                # slots and frozen require __getstate__ and __setstate__ to work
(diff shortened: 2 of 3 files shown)


Tinche:
This will bring a slight performance penalty to pickling (details at the end). This is not just theoretical for me, we pickle stuff at the end of every request.

Now obviously slower is better than not working, but I would just suggest adding a special case for Python 3: add the special methods iff slots=True and (Python2 or frozen=True). You can look in `attr._compat` for a way of detecting Python 2.

Benchmark example:
1. pip install perf

2. paste into t.py:
```
import attr


@attr.s(slots=True)
class A(object):
    a = attr.ib()
    b = attr.ib()
    c = attr.ib()
    d = attr.ib()
    e = attr.ib()
    f = attr.ib()
```
(pickling doesn't seem to work with just creating the class in the timeit invocation)

3. run `pyperf timeit --rigorous -g -s "import pickle; from t import A; a = A(1, 2, 3, 4, 5, 6)" "pickle.dumps(a)"`
For me, the difference is 3.83 us +- 0.18 (master) to 4.96 us +- 0.13 (yours).

mathieulongtin:
It gets more complicated: pickle protocol 2 and up support slots natively. Python 3.5 uses protocol 3 by defaults. Python 2 only supports protocol 0, 1 and 2 and defaults to 0. 

So do I:

1. set __getstate__/__setstate__ regardless
2. set __getstate__ only when needed by the default protocol (py2 and slots; py3 and slots and frozen); probably slowing down protocol 2 pickling on Python 2.
3. put an option to attrs `pickle_protocol` and set `__getstate__` if needed by the protocol version. It would default to the python version's default protocol.
4. put a module global `attr.pickle_protocol` and use that to decide whether to set __getstate__.

1 causes a performance downgrade with Python 3 and probably with Python 2 using proto2.

2 means Python 3 wouldn't work if pickle is used with protocol 0 or 1. It probably slows down Python 2 if proto2 is used.

3 is more flexible, but could be more problematic, as a library could define classes with attrs that are not compatible with your pickle protocol of choice.

4 is probably the best. It would work like option 2 by default, but it would be possible to say, at the beginning of your program, "I use protocol 2" even though I use Python 2, so don't put __getstate__ that would slow me down. But most of the time, you wouldn't need to think about it.

What do you think?

Tinche:
Before we consider it, is it possible, and if so, how convenient would it be for you and any other PySpark user to use protocol version 2 on Python 2.7?

If it's no problem, we could just document the requirement.

Tinche:
Great piece of research, btw.

mathieulongtin:
PySpark actually uses protocol 2. So it only had an issue because it tried to serialize `attr.Attribute` and other attr internals which are slots and frozen.

@hynek what do you think? I don't want to force a performance penalty on everybody.

A few more options:
1. only use `__getstate__` on frozen classes regardless of Python version, and put a note in the doc that `slots=True` require pickle protocol>=2.
2. don't use frozen on internal attr classes and note that frozen classes don't work well with pickle in the doc.

Tinche:
My personal recommendations (identical to @mathieulongtin except for the third bullet):

* put the methods on `attr.Attribute`, since pickling it won't work without in any case.
* document pickling slot classes may require pickle protocol >= 2.
* put the methods on `attrs` classes that are **both slots and frozen**. (Just frozen works without, so no need for the extra penalty.)

Minimal performance impact (only for classes that are both slots and frozen, and these don't work without it anyway), and I think requiring pickle protocol >= 2 is not a great burden. 

The fact of the matter is, ordinary (i.e. non-attrs) slot classes require a custom `__getstate__`/`__setstate__` anyway for protocol < 2. People can always write their own getstate if they absolutely need protocol < 2 directly in the class. Apparently protocol 2 was introduced in Python 2.3 and is more efficient anyway.

Of course @hynek's is the last word. :)

hynek:
Requiring protocol 2 sounds good to me.

mathieulongtin:
I applied the last suggestion by @Tinche. I consider this complete now. It works fine with PySpark.

hynek:
I’ve made a few minor adjustments and merged it.

Thanks to everyone involved!  I don’t care about pickles myself so it’s great to have domain experts. :)

PR #84, opened 2016-09-09
Improve tests, bandage assoc.

Tinche:
Ok, here's something easy to merge.

I've done some work on the simple_classes strategy. It's somewhat documented, takes some parameters, and implemented in a clearer way. This is in the hopes it will get used more by other contributors in tests.

I've then applied this strategy to assoc tests. The tests picked up failures right away. I've applied a dumb fix to assoc to make slot classes pass.

The tests picked up another failure though: if a class is both slots and frozen, assoc will fail deep in copy.copy. This isn't so easily fixed. 

Plot twist!

Digging around in the source of copy.copy, I see mentions of pickling. Applying these changes on to #81 allows copy.copy() to copy slots+frozen classes.

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 import copy
 
 from ._compat import iteritems
-from ._make import Attribute, NOTHING, fields, _obj_setattr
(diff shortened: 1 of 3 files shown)


hynek:
Merged in 1897dfd.

Issue #82, opened 2016-09-08
attr.filters.include/exclude don't work with attributes on slots classes


rouge8:
`attr.filters.include`/`attr.filters.exclude` don't work with attributes as expected from the docs on classes where `slots=True`.

```python
import attr

@attr.s(slots=True)
class Foo(object):
    bar = attr.ib()
    baz = attr.ib()

f = Foo(bar=1, baz=2)

assert attr.asdict(f, filter=attr.filters.exclude(Foo.bar)) == {'baz': 2}
```

hynek:
OK this is a “fun” one in multiple ways.

- include and exclude take advantage of the fact that classes have the `Attribute`s attached to them which I’ve deprecated in 16.1.  I completely forgot about the filters module TBH.  Meaning if we go through with it, you’d have to go via `Foo.__attrs_attrs__` (or we write some helper function).
- they do an `isinstance(..., Attribute)` to detect it’s an `Attribute` which fails with slots that returns an `<class 'member_descriptor'>`.

Therefore as far as I see it:

1. this feature in this form is literally impossible for ``__slots__`` classes and should be documented as such.
2. we need to deprecate them and replace them with a new API.

I’m happy to be corrected.

I’m sorry. :(

hynek:
Fixed, please use `attr.fields(Foo).bar` once 16.2.0 is out.

rouge8:
Awesome, thanks!

cosmosgenius:
> Fixed, please use `attr.fields(Foo).bar` once 16.2.0 is out.

We should add this as an example for exclude and include...it was confusing from documentation on how to use the filter.

Murtagy:
@hynek I think there is yet no reference for this in the docs.

To prefer using 
```
field = attrs.fields(model_obj.__class__)
fields.field_name
```
over 
```
fields[0] # (which you might generated from list compehension)
# e.g
no_fields = [f for f in fields if f.metadata.get('auto') is True]
attrs.filters.exclude(no_fields)  # does absolutely nothing for slotted class
```

hynek:
@cosmosgenius there is an example in https://www.attrs.org/en/stable/examples.html#converting-to-collections-types

@Murtagy I'm sorry but I'm not sure what exactly you mean, but it doesn't seem to be related to the original problem of accessing Attribute objects directly on classes (which doesn't work for 5 years now).

Murtagy:
Here is full example of the misleading behaviour I faced:
 
 ```python3 
[...]

PR #86, opened 2016-09-10
Implement by_name

hynek:
As a replacement for C.attribute that doesn't work with __slots__ and is
overall bad design.

I’ve also improved the overall exception situation (and its documentation).  Comments anyone?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,8 @@ Changes:
   `#81 <https://github.com/hynek/attrs/issues/81>`_
 - ``attr.assoc`` now works with ``__slots__`` classes.
   `#84 <https://github.com/hynek/attrs/issues/84>`_
+- Add ``attr.by_name`` for accessing ``Attribute`` instances of class objects.
+  ``attr.by_name(C, "x")`` is therefore equivalent to the deprecated ``C.x``.
 
 
 ----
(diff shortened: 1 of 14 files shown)


hynek:
This sort-of fixes #82.

Tinche:
A thought. What about instead of __attrs_attrs__  being a tuple of Attributes, we make it an ordered dictionary of attribute names to attributes?

It's an internal thing, and fields() can be changed to just return list(__attrs_attrs__.values()) or something.

Tinche:
Or, since I'm brainstorming, a namedtuple. That might even give us a prettier API than the dict-like by_name.

Tinche:
If it's a namedtuple, you could do both:

```
for a in fields(C):
    pass
```

and

```
my_attr = fields(C).my_attr
```
(Again, just brainstorming, it's too early for me :p)

hynek:
I kind of forgot, that the whole point of `fields()` is to enable us to do such changes. :D 

The irony of using `namedtuple`s within attrs is left to the reader.

Tinche:
The thought of using an `attrs` slot class here did cross my mind, but I was afraid of attrs-ception.

However, `namedtuples` can't have fields beginning with underscore, and we obviously support attributes with names beginning with underscore. This limitation is probably so you can't clash with a method name, like `_asdict`, `_make` or `_replace`. We are of course better than namedtuple and don't make our users deal with this crap.

I like the API a lot, so let me play around with it, implement it another way.

hynek:
OK the new API is

cr58n9nwiaapzm0

PR #87, opened 2016-09-10
remove unnecessary and ugly `exec_` compat hack

glyph:
The `eval` builtin does exactly this already, with the same signature even.


hynek:
eval

PR #88, opened 2016-09-10
Feature/fields properties

Tinche:
You're gonna loooove the implementation of this.

Well, you might actually love the API.

```
>>> from attr import Factory, fields
>>> fields(Factory)
(Attribute(name='factory', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None),)
>>> fields(Factory).factory
Attribute(name='factory', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None)
>>>
```
The way this is done is, for each attrs class we generate a tiny subclass of tuple. The subclass basically looks like this:
```
class FactoryAttributes(tuple):
    factory = property(itemgetter(0))
```
That's all.

No docs/changelog or anything until you say you like it.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -162,7 +162,7 @@ def _transform_attrs(cls, these):
     ]
     attr_names = [a.name for a in super_cls + non_super_attrs]
 
-    AttrsClass = _make_attr_tuple_class('', attr_names)
+    AttrsClass = _make_attr_tuple_class(cls.__name__, attr_names)
 
     cls.__attrs_attrs__ = AttrsClass(super_cls + [
         Attribute.from_counting_attr(name=attr_name, ca=ca)
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -3,6 +3,8 @@ from __future__ import absolute_import, division, print_function
 import hashlib
 import linecache
 
+from operator import itemgetter
+
 from . import _config
 from ._compat import iteritems, isclass, iterkeys
 from .exceptions import FrozenInstanceError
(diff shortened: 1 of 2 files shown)


hynek:
Jesus where do you get these ideas. 😮

attrs' new tagline: “we subclass tuple so you don‘t have to”

how does not everything break!?

hynek:
On a more serious note: does this handle well if you define two classes with the same name in different namespaces?

Tinche:
Nah I think it's fine.

got some classes lying around from the pickle thing:

```
$ cat t.py
import attr

@attr.s(slots=True, frozen=True)
class A(object):
    y = attr.ib()

    def __getstate__(self):
        return (self.y, )

        self.__init__(*state)

> cat tt.py
import attr


@attr.s(slots=True, frozen=True)
class A(object):
    z = attr.ib()

    def __getstate__(self):
        return (self.z, )

        self.__init__(*state)
> python
Python 3.5.2 (default, Jul  5 2016, 12:43:10) 
[GCC 5.4.0 20160609] on linux
>>> from t import A as TA
>>> from tt import A as TTA
>>> from attr import fields
>>> fields(TA)
(Attribute(name='y', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None),)
>>> fields(TTA)
(Attribute(name='z', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None),)
>>>
```

"We do stupid shit and test it so you don't have to" is actually an OK tagline :)

hynek:
I don't think we need any docs/CL since these ate implementation details.

Tinche:
You asked where I got the idea:  namedtuple! That's basically how it's
implemented. It's in Python, you can read the source easily.

I thought this would be the new name-based API. You know, for
include/exclude? Instead of deprecated class attributes. No?

hynek:
Hmmmm I thought this was more to simplify the planned implementations and pre-computing the attribute map.

But yeah the API is nicer this way.

Issue #12, opened 2015-06-11
asdict() and the way back with “private” attributes


sscherfke:
This is actually not unexpected, but I hoped it would work:

``` python
>>> import attr
>>> @attr.s
... class A:
...     x = attr.ib()
...     _y = attr.ib()
... 
>>> a = A(1, 2)
>>> attr.asdict(a)
{'_y': 2, 'x': 1}
>>> A(**_)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __init__() got an unexpected keyword argument '_y'
```

Maybe a function `attr.fromdict(cls, dct)` would solve this?  It would remove the leading `_` from all dict keys and then create the instance. 

Since you exlicitly talk about [serialization to JSON](https://attrs.readthedocs.org/en/stable/examples.html#converting-to-dictionaries) in the docs, there should IMHO also be an easy way for deserialization.

hynek:
So preamble:  I *do* not think, that attrs should solve every problem ever.  It’s main focus is to reduce common boilerplate.

That said, it seems like a `attr.fromdict(A, a)` may have it’s place for symmetry alone?

sscherfke:
Yes.  An alternative could be to add an additonal kwarg (like `remove_leading__` ;-)) to `asdict()` so that its return value can be passed as kwargs to the class’ `__init__()`.

hynek:
I’m on the fence here a bit.  If you serialize stuff to JSON, you usually would want to exclude private data (which you can with the correct filter; maybe I should add one in fact).  And since the result is just a dict, you can rename the keys if you don't like them.

Another alternative would be to allow filter callables to return the dict name (to make it more general than just stripping _).

sscherfke:
I have some “private“ attributes and manually added properties to make them read-only. That's why they should be included when I convert to a dict.

glyph:
The main thing that *I* want `fromdict` to do is to be able to re-structure-ify semistructured JSON data.  For example, if I have (pardon the formatting, getting the hang of doing stuff like this with ipython):



```python
import attr
from attr.validators import instance_of as an
```


```python
@attr.s()
class B(object):
    z = attr.ib()
@attr.s()
class A(object):
    x = attr.ib(validator=an(B))
    y = attr.ib()
```


```python
b = B("test")
a = A(b, 7)
```


```python
a
```




    A(x=B(z='test'), y=7)




```python
attr.asdict(a)
```




[...]

PR #91, opened 2016-09-11
Add `fromdict` function and `attr.ib(type)` property

ChrisNeveu:
This PR adds a new `type` property to `Attribute`s. I've also included in this PR a motivating example for this addition in the form of a `fromdict` function based on the `type` field. The `type` property could also be useful for possible future integration into mypy. I've never contributed to attrs before so I apologize for the large size of this changeset.

I can split the `type` addition and `fromdict` addition into two PRs if you'd like, I mostly included `fromdict` to show how the addition of the `type` property could be used.

Please let me know if there are any changes you'd like to see to bring this more in line with your overall philosophy for attrs.


src/attr/_make.py
```diff
@@ -248,8 +251,13 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
 
         ..  _slots: https://docs.python.org/3.5/reference/datamodel.html#slots
 
+    :param bool validate_types: Check that each value passed to the constructor
```
ChrisNeveu: This reduces redundancy in having to specify both a type and the `instance_of` validator for that type.

ChrisNeveu:
Apologies for the test failures, there is a slight difference in `typing` between 3.5.2 and 3.5.1 (which I was using locally).

hynek:
Hi Chris thank you for the work you’ve put into this!  Have you seen the rainbow bikeshed on this topic #12?  We haven’t decided on the ideal way to go this way so it’s too early to put time and sweat into implementations. :(

ChrisNeveu:
I've read through that discussion. I think the added `type` field is a great way to resolve some of the concerns in that thread regarding `instance_of`, but even outside of `fromdict` it would be nice to have the field for eventual mypy integration (I'm working on a plugin).

If you're willing to merge just the addition of a `type` field, let me know and I'll remove `fromdict`.

hynek:
I'm honestly sorry for the effort you put into this but for now, http://cattrs.readthedocs.io/ is the playground for this kind of stuff.

PR #96, opened 2016-09-21
Introduce attribute metadata.

Tinche:
Alright, here's a first pass, open for comments. Docs will come after the implementation is settled.

The syntax is:

```
@attr.s
class A:
    a = attr.ib(metadata=<metadata>)
```

where `<metadata>` can be:
- None
- a mapping (technically something that can be passed to the dict constructor)

On Python 2, the metadata (if not None) will be exposed as a plain dict. On Python 3, the original, passed-in mapping will be wrapped in a https://docs.python.org/3.5/library/types.html#types.MappingProxyType.

I'm not entirely happy with None as the default here (it's my pet peeve to have Optional[Collection] instead of just an empty collection) but I feel an empty dict per attribute might be too much memory overhead? MappingProxyType would have solved this elegantly (we would have one empty instance, used everywhere where there was no metadata), but alas.

I've tried marking the metadata field as excluded from hashing but I don't think we care much about that. I put together a property test for this but it kept failing since the `default` field isn't excluded, and the default was getting set to `{}`. Guess we don't care about it too much? Doubt many people will make sets of Attributes.

As a bonus I've made _CountingAttr slotted. There's no reason it shouldn't be.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,10 +11,10 @@ The third digit is only for regressions.
 Changes:
 ^^^^^^^^
 
+- Attributes now can have user-defined metadata which greatly improves ``attrs`` extensibility.
+  `#96 <https://github.com/hynek/attrs/pull/96>`_
 - Don't overwrite ``__name__`` with ``__qualname__`` for ``attr.s(slots=True)`` classes.
   `#99 <https://github.com/hynek/attrs/issues/99>`_
-- Attributes can now have metadata dictionaries.
-  `#96 <https://github.com/hynek/attrs/pull/96>`_
 
 
 ----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ The third digit is only for regressions.
 Changes:
 ^^^^^^^^
 
-- Attributes now can have user-defined metadata which greatly improves ``attrs`` extensibility.
+- Attributes now can have user-defined metadata which greatly improves ``attrs``'s extensibility.
   `#96 <https://github.com/hynek/attrs/pull/96>`_
 - Don't overwrite ``__name__`` with ``__qualname__`` for ``attr.s(slots=True)`` classes.
   `#99 <https://github.com/hynek/attrs/issues/99>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Changes:
 
 - Don't overwrite ``__name__`` with ``__qualname__`` for ``attr.s(slots=True)`` classes.
   `#99 <https://github.com/hynek/attrs/issues/99>`_
+- Attributes can now have metadata dictionaries.
+  `#96 <https://github.com/hynek/attrs/pull/96>`_
 
 
 ----
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -458,7 +458,7 @@ Slot classes are a little different than ordinary, dictionary-backed classes:
     ... class C(object):
     ...     x = attr.ib()
     >>> C.x
-    Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None)
+    Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None, metadata=mappingproxy({}))
     >>> @attr.s(slots=True)
     ... class C(object):
     ...     x = attr.ib()
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -422,7 +422,7 @@ All ``attrs`` attributes may include arbitrary metadata in the form on a read-on
     >>> attr.fields(C).x.metadata['my_metadata']
     1
 
-Metadata is currently not used by ``attrs``, and is generally meant to enable rich functionality in third-party libraries.
+Metadata is not used by ``attrs``, and is meant to enable rich functionality in third-party libraries.
 The metadata dictionary follows the normal dictionary rules: keys need to be hashable, and both keys and values are recommended to be immutable.
 
 If you're the author of a third-party library with ``attrs`` integration, please see :ref:`Extending Metadata <extending_metadata>`.
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -405,6 +405,29 @@ Converters are run *before* validators, so you can use validators to check the f
     ValueError: x must be be at least 0.
 
 
+.. _metadata:
+
+Metadata
+--------
+
+All ``attrs`` attributes may include arbitrary metadata in the form on a read-only dictionary.
+
+.. doctest::
+
+    >>> @attr.s
+    ... class C(object):
+    ...    x = attr.ib(metadata={'my_metadata': 1})
+    >>> attr.fields(C).x.metadata
+    mappingproxy({'my_metadata': 1})
+    >>> attr.fields(C).x.metadata['my_metadata']
+    1
+
+Metadata is currently not used by ``attrs``, and is generally meant to enable rich functionality in third-party libraries.
+The metadata dictionary follows the normal dictionary rules: keys need to be hashable, and both keys and values are recommended to be immutable.
+
+If you're the author of a third-party library with ``attrs`` integration, please see :ref:`Extending Metadata <extending_metadata>`.
+
+
 .. _slots:
 
 Slots
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -17,7 +17,7 @@ So it is fairly simple to build your own decorators on top of ``attrs``:
    ... @attr.s
    ... class C(object):
    ...     a = attr.ib()
-   (Attribute(name='a', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None),)
+   (Attribute(name='a', default=NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None, metadata=mappingproxy({})),)
 
 
 .. warning::
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -51,11 +51,11 @@ Here are some tips for effective use of metadata:
 
 - To avoid metadata key collisions, consider exposing your metadata keys from your modules.::
 
-    from mylib import my_metadata_key
+    from mylib import MY_METADATA_KEY
 
     @attr.s
     class C(object):
-      x = attr.ib(metadata={my_metadata_key: 1})
+      x = attr.ib(metadata={MY_METADATA_KEY: 1})
 
   Metadata should be composable, so consider supporting this approach even if you decide implementing your metadata in one of the following ways.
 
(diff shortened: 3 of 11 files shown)


hynek:
FTR I’m mostly fine with the approach and am waiting for feedback from other people that wanted metadata in #63 most notably @wearpants :)

wearpants:
Looks basically like what I'm after :-)

Though I'd agree with @Tinche's instinct that metadata should always be a mapping even when empty - memory overhead for an extra dict doesn't seem like a big deal to me, it's per-class attr-using class not per instance (and will generally stay out of cache lines if unused, and maybe even get swapped out if you're lucky with page layouts). Could use MappingProxy conditionally if it's available but probably overkill.

hynek:
FWIW we could define a dict-like singleton and be done with it.

wearpants:
Hmm, singleton would mean metadata is read-only after creation - that what we want?

hynek:
I thought that’s the plan but in any case it’s better to start stricter and loosen up if necessary.  Unless someone has some great example why we mutability?

Tinche:
OK, let's agree that an empty dict is a preferable API than an Optional dict. I'll play around and whip something up.

@wearpants Yeah, attributes are immutable and metadata as part of an attribute should be immutable, ideally.

wearpants:
Wfm

Tinche:
Here, metadata is immutable on Python 2 now too (best effort basis, via a simple immutability wrapper in `_compat.py`). We use a singleton for empty metadata now.

When I was testing this, instead of adapting the `simple_classes` Hypothesis strategy we already have to sometimes include metadata, I defined another strategy just for metadata tests. This was a dumb decision! If I hadn't done that, I'd have found out earlier that MappingProxyType on Python 3 breaks pickle. (That's fixed too.)

hynek:
I have so minor stuff but would like to settle on the big picture first.  @glyph you OK with this?

glyph:
@hynek The `mappingproxy` boilerplate makes me a bit sad, but I guess it's a total implementation detail, so that's fine.

Rather than allowing `None`, why not just set the default to `{}` and then do the `mappingproxy` construction in `_CountingAttr` rather than `attr`?

Also, I would very much like to see some information on both setting and getting metadata in the docs, if the feature is going to be added.  Writing the docs before adding a controversial feature can often uncover weird little edge cases.  I don't think that it will do so in this case, since just looking up in a `dict` is so simple - but I always think it's not going to do so until I do it!

hynek:
I guess what glyph is saying: @tinche pls finalize.  :)

Tinche:
I've been a little sick these past few days, hence the slow progress.

While working on the docs, this style of metadata composability came up. What do we think?

```
# Library one.
my_type_metadata = '__my_type_metadata'

def typed(cls, default=attr.NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None, metadata={}):
    metadata = dict() if not metadata else metadata
    metadata[my_type_metadata] = cls
    return attr.ib(default, validator, repr, cmp, hash, init, convert, metadata)

# Library two.
my_json_metadata = '__my_json_metadata'

def json_name(name, default=attr.NOTHING, validator=None, repr=True, cmp=True, hash=True, init=True, convert=None, metadata={}):
        if isinstance(default, attr._CountingAttr):
            # Wrapping a prepared _CountingAttr instance.
            metadata = dict() if not default.metadata else default.metadata
            metadata[my_json_metadata] = name
            return default
        else:
            # Wrapping ``attr.ib()``.
            metadata = dict() if not metadata else metadata
            metadata[my_json_metadata] = cls
            return attr.ib(default, validator, repr, cmp, hash, init, convert, metadata)

# The user code.
@attr.s
class C(object):
    x = json_name('X', typed(int, default=1, init=False))
    y = json_name('Y', default='A')
attr.fields(C).x.metadata[my_type_metadata]
<class 'int'>
attr.fields(C).x.metadata[my_json_metadata]
'X'
```

Hot or not? Top or flop?

glyph:
1. I like the idea.
2. `_CountingAttr` is private, so, that's not great.
3. We can only stack these in one order, `typed`, then, `json_name`, which means that some random library can define "extension" metadata in a way that doesn't compose with another library which also expects to be at the "bottom" of the wrapper-stack.
4. I'd probably go with `*optional_wrapped_attr`, and check its length, rather than multi-purposing `default`

Anyway, it's a promising direction :)

wearpants:
Is this something that attrs should be supporting directly or is this up to
users to implement? I can see a role for providing general helpers for
composition, but not sure attrs should provide implementation for specific
use cases like JSON... Reminds me a bit of the canonical code snippets in
itertools docs.

But yeah this looks promising

>

Tinche:
I was working on the docs and I wanted to put a pattern for defining attribute wrappers under Extending, and this came up. I'll use the second type :)

hynek:
So I’m back from ZA; did I miss something? :)

Tinche:
I've added some docs. I think recommending a composable attr.ib wrapper like I mentioned can wait for another PR since there are additional changes (and discussion) required for it.

hynek:
Before I go into details:

1. semantic new lines
2. double quotes
3. `"""\ndocstring\n"""`

please :)

Tinche:
@hynek I've addressed the docstrings and double quotes. I don't really understand semantic newlines in docs so you might have to swoop in and do some formatting.

hynek:
Semantic newlines just means to press enter after each sentence. ;)  On a first glance it looks fine.

Final RfC for @glyph and @wearpants is hereby open.  I’ll try to implement something I had in mind with it to check its usability.

wearpants:
Will look over in next day or two

wearpants:
Looks good to me :-)

hynek:
First of all: @Tinche I’m super sorry I didn’t get around earlier.

But.

On this rather sad day in history, I’m kind of excited, because this is what I did: https://gist.github.com/hynek/78fef47fea3cbc6b683cd8ab0fb68567

I used the meta data feature to implement a declarative way for having configuration via env variables!

First file is client side usage, middle file is implementation, and final file is program output.

I’d love to hear some thoughts on this!

***

There is one question tho: if we add attribute meta data, wouldn’t it also make sense to add class meta data?  In my example I could put most of the arguments of `environ_to_app_config()` into some class meta data for an even nicer experience.

(Sorry for disgusting legacy Python; I used it within a Twisted app that can’t be ported yet.)

wearpants:
Insufficiently cadfineated to understand the implementation, but THAT'S AWESOME @hynek.

Not sure I follow on class metadata - couldn't regular class vars do that? 

Might be nice to have an `attr.contrib` package for including batteries like this (or perhaps a recipe collection somewhere)

hynek:
I’ve updated it and now I’m reasonably excited as much as I can get excited about software while being depressed about the world.

Just for posterity, *this* is what application code boils down to:

```python
from app_config import (app_config, env_var, env_var_from_vault,
                        env_var_group, environ_to_app_config)


@app_config(prefix="APP", vault_prefix="WHOIS_{env}")
class WhoisConfig:
    @app_config
    class Prometheus:
        address = env_var(default="127.0.0.1")
        port = env_var(default="0")
        consul_token = env_var_from_vault()

    env = env_var()
    prometheus = env_var_group(Prometheus)


if __name__ == '__main__':
    ac = environ_to_app_config(WhoisConfig)

    print(ac)
```

and it gives you this:

```console
$ env APP_ENV=dev APP_PROMETHEUS_PORT=7000 SECRET_WHOIS_DEV_APP_PROMETHEUS_CONSUL_TOKEN=abc python app.py
WhoisConfig(env='dev', prometheus=WhoisConfig.Prometheus(address='127.0.0.1', port='7000', consul_token='abc'))
```

docs/examples.rst
```diff
@@ -405,6 +405,29 @@ Converters are run *before* validators, so you can use validators to check the f
...
+
+    >>> @attr.s
+    ... class C(object):
+    ...    x = attr.ib(metadata={'my_metadata': 1})
+    >>> attr.fields(C).x.metadata
+    mappingproxy({'my_metadata': 1})
+    >>> attr.fields(C).x.metadata['my_metadata']
+    1
+
+Metadata is currently not used by ``attrs``, and is generally meant to enable rich functionality in third-party libraries.
```
hynek: I think we can remove “currently” and “generally”.

attrs will *never* use metadata; we have API freedom. :)

docs/extending.rst
```diff
@@ -36,3 +36,43 @@ So it is fairly simple to build your own decorators on top of ``attrs``:
...
+If you're the author of a third-party library with ``attrs`` integration, you may want to take advantage of attribute metadata.
+
+Here are some tips for effective use of metadata:
+
+- Try making your metadata keys and values immutable.
+  This keeps the entire ``Attribute`` instances immutable too.
+
+- To avoid metadata key collisions, consider exposing your metadata keys from your modules.::
+
+    from mylib import my_metadata_key
```
hynek: Since it’s a constant, we should upper it I think?

docs/extending.rst
```diff
@@ -36,3 +36,43 @@ So it is fairly simple to build your own decorators on top of ``attrs``:
...
+      x = attr.ib(metadata={my_metadata_key: 1})
+
+  Metadata should be composable, so consider supporting this approach even if you decide implementing your metadata in one of the following ways.
+
+- Expose ``attr.ib`` wrappers for your specific metadata.
+  This is a more graceful approach if your users don't require metadata from other libraries.
+
+  .. doctest::
+
+    >>> my_type_metadata = '__my_type_metadata'
```
hynek: In my case with app_config I went for `object()`.  It’s a bit ugly in the repr but it’s more fool proof because the keys just *can’t* clash.  Opinions?

hynek:
Oh an Changelog please. :)

16.3.0 goes out once this is merged.

Tinche:
Will get around to this today/tomorrow :)

hynek:
flake8 is failing

Tinche:
Huh, flake8 put out a new release and now it wants two blank lines *after* class definitions too :)

Anyway, I addressed all your points except what to use as metadata keys. My gut says strings are better since the attribute repr will be nicer and I don't think clashes will be a big deal in practice. Also that one person needed to serialize attributes, don't know how this would play out if using object() instances as keys. I don't feel that strongly about it, so please feel free to change the docs yourself if you want to. :)

CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ Changes:
 
 - Don't overwrite ``__name__`` with ``__qualname__`` for ``attr.s(slots=True)`` classes.
   `#99 <https://github.com/hynek/attrs/issues/99>`_
+- Attributes can now have metadata dictionaries.
```
hynek: I feel like you’re underselling your work here!

How about something like:

> Attributes now can have user-defined attribute metadata which greatly improves `attrs`’s extensibility.

?

Tinche:
Deal. I also moved it up since it's more interesting than a qualname bugfix :)

hynek:
I’m confident like this it's grammatically wrong: either drop `attrs` for good or make it possessive (“`attrs`’s” or “`attrs`’”).

Tinche:
Tweaked it :)

hynek:
Thanks everyone!

PR #98, opened 2016-09-23
Ensure the LICENSE file is included in wheel metadata

pombredanne:
Without this small setup.cfg addition a built wheel does not contain your LICENSE file.
With this, the license file will be added to a built wheel metadata.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,3 +7,7 @@ testpaths = tests
 
 [bdist_wheel]
 universal = 1
+
+[metadata]
+# ensure LICENSE is included in wheel metadata
+license_file = LICENSE


hynek:
Oh!  Do you have a link to relevant documentation/PEP?  I’ll probably have to do that for all my projects.

pombredanne:
@hynek there is no pep and little doc on this, this is just an extra thing for wheels.
This comes from https://bitbucket.org/pypa/wheel/issues/47/include-full-license-text-in-dist-info
This is an area that is not great IMHo but is evolving (Daniel Holth is working on a PEP and I committed to help).
For now, this is the way short of using package data, otherwise your license file would not be included.

See also https://bitbucket.org/pypa/wheel/issues/138/include-more-than-one-license-file-or

hynek:
OK thanks.  This lack of documentation is unfortunate. :(

PR #100, opened 2016-09-25
don't overwrite __name__ with __qualname__

rdpate:
https://github.com/hynek/attrs/issues/99

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -285,12 +285,10 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
                 cls_dict.pop(ca_name, None)
             cls_dict.pop('__dict__', None)
 
-            if repr_ns is None:
-                cls_name = getattr(cls, "__qualname__", cls.__name__)
-            else:
-                cls_name = cls.__name__
-
-            cls = type(cls_name, cls.__bases__, cls_dict)
+            qualname = getattr(cls, "__qualname__", None)
+            cls = type(cls.__name__, cls.__bases__, cls_dict)
+            if qualname is not None:
+                cls.__qualname__ = qualname
 
         return cls
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,8 @@ The third digit is only for regressions.
 Changes:
 ^^^^^^^^
 
-*none*
+- Don't overwrite ``__name__`` with ``__qualname__`` for ``attr.s(slots=True)`` classes.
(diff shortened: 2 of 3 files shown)


src/attr/_make.py
```diff
@@ -285,12 +285,10 @@ def wrap(cls):
...
 
-            if repr_ns is None:
-                cls_name = getattr(cls, "__qualname__", cls.__name__)
-            else:
-                cls_name = cls.__name__
-
-            cls = type(cls_name, cls.__bases__, cls_dict)
+            qualname = getattr(cls, "__qualname__", None)
+            cls = type(cls.__name__, cls.__bases__, cls_dict)
+            if qualname:
```
Tinche: I already know @hynek is going to ask, so: please make this 'if qualname is not None'

hynek:
Wait, why is this backward-incompatible?  Could someone actually sum up the nature of the changes?  I’m in the middle of preparations for a DevOps Sprint + PyCon ZA so I don’t have the FOSS time I’d like to have.

Tinche:
This is the issue:
```
import attr

@attr.s(slots=True)
class A:
    @attr.s(slots=True)
    class B:
        pass

A.B.__name__
'A.B'
```
(It should be just 'B'.)

I don't think it's a backwards-incompatible change, think @rdpate got a little overzealous in the changelog (I didn't even notice it until now). It's just a bug fix imo, and a minor one at that.

rdpate:
I would have not have listed it backwards-incompatible, except I noticed the 16.1.0 change of "cl" to "cls" was also listed backwards-incompatible:

> All instances where function arguments were called cl have been changed to the more Pythonic cls. Since it was always the first argument, it's doubtful anyone ever called those function with in the keyword form. If so, sorry for any breakage but there's no practical deprecation path to solve this ugly wart.

Fixing \_\_name\_\_ seemed at least as big as changing "cl", so I went with that.

rdpate:
> Could someone actually sum up the nature of the changes?

Don't overwrite \_\_name\_\_ with \_\_qualname\_\_.  Preserve \_\_qualname\_\_ instead of discarding it.   It should do nothing else, and only matters when slots=True.

hynek:
The cls thing was possibly breaking if someone always uses kwargs and I fought with myself for a long time whether it's breaking. As Tin mentioned everything is breaking to someone :D but this is a plain bug fix to me.

rdpate:
Sounds good to me. :) I've simplified the changelog description and folded the commits together.

hynek:
Thanks!

PR #111, opened 2016-11-11
Add ability for ``__post_init__`` method to be defined.

tbeadle:
If this method is defined for a class, it will get executed at the end
of ``__init__``.  Previously, there was no way to extend what was
done during ``__init__`` when using ``@attr.s``.


src/attr/_make.py
```diff
@@ -680,6 +680,10 @@ def fmt_setter_with_converter(attr_name, value_var):
                                                     a.name))
         names_for_globals[val_name] = a.validator
         names_for_globals[attr_name] = a
+    lines.extend([
+        "func = getattr(self, '__post_init__', None)",
```
hynek: This is approach means that `attrs` gets universally slower for everyone.  `attrs` is explicitly static in its approach.  We need to check for the existence of the method when assembling the class and call it unconditionally if it’s present.

src/attr/_make.py
```diff
@@ -446,7 +446,11 @@ def _add_init(cls, frozen):
         sha1.hexdigest()
     )
 
-    script, globs = _attrs_to_script(attrs, frozen)
+    script, globs = _attrs_to_script(
+        attrs,
+        frozen,
+        post_init=hasattr(cls, '__post_init__')
```
hynek: [No hasattr ever](https://hynek.me/articles/hasattr/) please.  This lends itself to `getattr(cls, "__post_init__", False)`.

docs/examples.rst
```diff
@@ -551,6 +551,25 @@ You can still have power over the attributes if you pass a dictionary of name: `
    >>> i.y
    []
 
+Sometimes, you want to have your class's ``__init__`` method do more than just
```
hynek: Semantic new lines pls.

CHANGELOG.rst
```diff
@@ -13,6 +13,9 @@ Changes:
 
 - Don't overwrite ``__name__`` with ``__qualname__`` for ``attr.s(slots=True)`` classes.
   `#99 <https://github.com/hynek/attrs/issues/99>`_
+- Allow for a ``__post_init__`` method that, if defined, will get executed at
```
hynek: We use semantic new lines in prose. Simply make each sentence a new line please.

Also please add a link to this PR like in the entry before.

tests/test_make.py
```diff
@@ -302,6 +302,22 @@ class D(object):
...
+        @attributes
+        class C(object):
+            x = attr()
+            y = attr()
+
+            def __post_init__(self2):
+                self2.z = self2.x + self2.y
+
+        c = C(x=10, y=20)
+        assert hasattr(c, 'z')
```
hynek: no hasattr

tests/test_make.py
```diff
@@ -302,6 +302,22 @@ class D(object):
...
+        assert c.z == 30
```
hynek: `expected == actual` please

src/attr/_make.py
```diff
@@ -680,6 +684,8 @@ def fmt_setter_with_converter(attr_name, value_var):
                                                     a.name))
         names_for_globals[val_name] = a.validator
         names_for_globals[attr_name] = a
+    if post_init:
+        lines.append("self.__post_init__()")
```
mathieulongtin: I'd say post validation. I'd expect `__post_init__` to be the very last thing init calls. Beside, we can call `attr.validate`.

Tinche: I'm voting post validation/conversion. The name `__post_init__` itself suggests it's called after the generated `__init__`, so I'd expect fields to be validated, converted and set. Having it called post-conversion lowers the amount of possible states we can expect the object to be in.

hynek: So be it.

We’ll have to add the pass-self-to-factories thing then because otherwise there’s no way to validate dependent attributes.

glyph: The other way to validate dependent attributes would be for `__post_init__` to raise an exception, no?

hynek: Sure but I like the idea of separating the factory process from the validation process.

tests/test_make.py
```diff
@@ -302,6 +302,22 @@ class D(object):
         assert C.D.__name__ == "D"
         assert C.D.__qualname__ == C.__qualname__ + ".D"
 
+    def test_post_init(self):
```
tbeadle: `make_class` already supports it.  A `__post_init__` key can be defined with the dictionary passed in to make_class:

```python
>>> import attr
>>> def foo(self):
...     print('in foo')
...
>>> C1 = attr.make_class('C1', {'x': attr.ib(default=42), '__post_init__': foo})
>>> c = C1()
in foo
```

wearpants:
Late to the party, but yes absolutely post conversion

tests/test_make.py
hynek: Fair enough, please just add it to our testing strategy then.

hynek: oh wait you did never mind :)  I’ll have another deeper look on the weekend

hynek:
I’ve fixed it up so we don’t have ping-pong over minutiae anymore, thank you very much for your contribution!

I’ve decided to rename `__post_init__` to `__attrs_post_init__` for consistency with `__attrs_attrs__` and to be explicit about it’s heritage.

If someone has good arguments against it, feel free to open a new issue before 16.3.0 is out.

PR #118, opened 2016-11-30
ENH: forbid delete attr from frozen class

tacaswell:
Raise FrozenInstanceError when trying to delete an attribute from
a frozen class.

On non-frozen classes `del` results in un-shadowing the `attr.ib` object which is a bit surprising.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,8 +24,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
-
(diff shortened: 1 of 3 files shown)


hynek:
This makes sense.  Would you please add a changelog entry?

tacaswell:
Sorry that took so long.  Went for 'change' not 'API break'.

hynek:
Thanks!  I’m [fixed](https://github.com/hynek/attrs/commit/48943778461d103455f36fca8b8875f4db6be13a) it myself but next time use semantic line feeds please. :)

PR #119, opened 2016-11-30
Feature/better simple classes

Tinche:
This should be safe to merge in any case. Generate better test cases.

The .gitignore addition is for my convenience ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ htmlcov
 dist
 .cache
 .hypothesis
+.venv*
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ htmlcov
 dist
 .cache
 .hypothesis
-.venv*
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -74,6 +74,14 @@ def gen_attr_names():
             yield outer + inner
 
 
+def maybe_underscore_prefix(source):
+    """A generator to sometimes prepend an underscore."""
```
hynek: *cough* `"""` on separate lines *cough*

I should really look into pylint so I don’t get to be the asshole that pesters everyone. :)

Tinche:
There, easily fixed. I really do prefer the one-line version when the comment is short-ish and fits on one line though.

.gitignore
```diff
@@ -6,4 +6,5 @@ docs/_build/
 htmlcov
 dist
 .cache
-.hypothesis
\ No newline at end of file
+.hypothesis
+.venv*
```
hynek: oh and please don't do this I don’t want put stuff there that doesn’t affect everyone.  do you know about `.git/info/exclude`?

Tinche:
Googling around, I added the rule to the global gitignore file (~/.config/git/ignore). Problem fixed forever.

PR #122, opened 2016-12-08
docs: "less [...] calls" -> "fewer [...] calls"

rjw57:
Since function calls are discrete one has "fewer" of them.


hynek:
Thanks!

PR #126, opened 2016-12-15
Add a testtools matcher-based validator.

exarkun:
Fixes https://github.com/hynek/attrs/issues/13

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -275,6 +275,23 @@ Validators
       >>> C(None)
       C(x=None)
 
+.. autofunction:: attr.validators.matches
+
+   For example:
+
+   .. doctest::
+
+      >>> from testtools.matchers import GreaterThan
+      >>> @attr.s
+      ... class C(object):
+      ...     x = attr.ib(validator=attr.validators.matches(GreaterThan(10)))
+      >>> C(42)
+      C(x=42)
+      >>> C(5)
+      Traceback (most recent call last):
+         ...
+      TypeError: ("'x' is invalid: 5 <= 10')
+
 
 Deprecated APIs
 ---------------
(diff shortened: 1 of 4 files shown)


hynek:
Hm it certainly intriguing to pass on this kind functionality to someone who considers on of their core features.

But is this something people actually *do*?  Use testtools in normal code?

exarkun:
What happened to https://github.com/hynek/attrs/issues/13#issuecomment-113104985 which seemed to be explicit approval of this feature?

No, I've never done this before - but I contributed this upstream so I *could* do it with an official attrs API instead of carrying the implementation in my random no-one-will-ever-look-at-it codebase.  I started with a custom, attrs-only implementation of a testtools AllMatches-alike and was quite happy to be able to throw it away and use someone else's code (and at the same time open up many other combining operators).

exarkun:
@hynek ping

hynek:
What happened?  Dementia happened. :D 

So I’m cool with adding a testtools-based matcher if enough people think it’s a good idea and they would use it (@cyli? @glyph? @PsyMar?) but I’m not entirely happy with name that makes it look like it’s self-implemented.  `testtools_matches()` OTOH seems a bit of mouthful?  Opinions?

PsyMar:
Hm, never encountered testtools before (I only started python last year, don't blame me).  It says it derives from unittest, which never looked particularly helpful, but maybe that's just because of the examples I've seen.  I'll look through those docs...

as for a name, I don't mind long names, but I switched to Python from Java (and previously C++ which the boost libraries are AWFUL about long names.)

hynek:
I’m sorry but I’m going to close this now.  Thinking about it, testtools seems like a poor choice to use as the official validation vehicle.

I think I’d rather add something like https://github.com/alecthomas/voluptuous which is more generic and production tested (by myself no less :)).

Sorry for the work jp! :(

exarkun:
Meh :/

Issue #129, opened 2017-01-05
Post init hook is not run if validators are disabled


p-himik:
Here's the `__init__` code generated by attrs:
```python
def __init__(self, name, loc):
    self.name = name
    self.loc = loc
    if _config._run_validators is False:
        return
    __attr_validator_name(self, __attr_name, self.name)
    __attr_validator_loc(self, __attr_loc, self.loc)
    self.__attrs_post_init__()
```

PR #130, opened 2017-01-05
Fix post init hook not being called if validators are disabled (#129)

p-himik:


hynek:
Well that sure is an oops! :)  Could you add tests and a changelog entry please?

src/attr/_make.py
```diff
@@ -700,15 +700,14 @@ def fmt_setter_with_converter(attr_name, value_var):
...
-        lines.append("if _config._run_validators is False:")
-        lines.append("    return")
-    for a in attrs_to_validate:
-        val_name = "__attr_validator_{}".format(a.name)
-        attr_name = "__attr_{}".format(a.name)
-        lines.append("{}(self, {}, self.{})".format(val_name, attr_name,
-                                                    a.name))
-        names_for_globals[val_name] = a.validator
-        names_for_globals[attr_name] = a
+        lines.append("if _config._run_validators:")
```
hynek: Please use the more explicit `if _config._run_validators is True:`.

CHANGELOG.rst
```diff
@@ -26,6 +26,8 @@ Changes:
 
 - Raise ``FrozenInstanceError`` when trying to delete an attribute from a frozen class.
   `#118 <https://github.com/hynek/attrs/pull/118>`_
+- Fix __attrs_post_init__ not being run when the validation is disabled.
```
hynek: Please put `__attrs_post_init__` in double backticks to make it code (like FrozenInstanceError above).

hynek:
Thanks!

PR #149, opened 2017-02-19
Suppress HealthCheck.too_slow for TestAsTuple.test_recurse_property

floppym:
On very slow hardware, this test fails due to a hypothesis health check.

Bug: https://bugs.gentoo.org/show_bug.cgi?id=608570


hynek:
That suppresses the health check only on the one test, right?  Wouldn’t it be better to allow to suppress them completely for such cases?  Summoning @Tinche :)

Tinche:
Sure. @floppym, could you suppress the slow warning on all test cases that make use of nested_classes?

floppym:
@Tinche Maybe I'm missing something, but this is the only test I see that references `nested_classes`.

Tinche:
@floppym TestAsDict.test_recurse_property, TestAsTuple.test_recurse_property, TestAsTuple.test_recurse_retain :)

floppym:
Oh, duh. I force pushed a new changeset.

hynek:
Did the org transition break CodeCov? 😭

hynek:
Thank you for your contribution!

PR #152, opened 2017-02-23
pass bases keyword argument to make_class

brainysmurf:
Currently attr is hard-coded to use `(object,)` as the bases argument in the `type()` call. Adding bases as keyword argument would allow `attr.make_class` to be used in more situations (where you have to use a subclass, and you have to make a class dynamically).

Testing-wise, not sure if another test is needed for this well-understood feature; I ran pytest with this modification and it passed all of them.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,12 +8,6 @@ The third digit is only for regressions.
 17.1.0 (UNRELEASED)
 -------------------
 
-Changes:
-^^^^^^^^
-
-- ``attr.make_class`` accepts new keyword argument ``bases``.
-  Previously, if you were creating a class dynamically the call to ``type(n, b, d)`` was hard-coded to be ``(object,)``. By exposing this keyword it's possible to dynamically create subclassed classes. Use cases for this may be, for example, class decorators.
-
 
 Backward-incompatible changes:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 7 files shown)


hynek:
Wellll, testing-wise you’ve lowered the code coverage. :)

Before you spend more time coding: could you provide us with a practical example of what you want to achieve?  That’s also something that should go into examples.rst if we admit this change.

brainysmurf:
I coded up an example. I'd like to build a model where equality is defined by the attributes and properties being equal (equal key with equal values). This works to the specifications:

```python

import attr

class Base:
    def __eq__(self, other):
        defaults = dir(attr.make_class('None', {}))
        keys = []
        for attr_or_prop in [key for key in dir(self) if key not in defaults]:
            keys.append(attr_or_prop)
        return all([getattr(self, key) == getattr(other, key) for key in keys])

@attr.s(cmp=False)
class ModelOne(Base):
    stuff = attr.ib(default='')

@attr.s(cmp=False)
class ModelSame(Base):
    stuff = attr.ib(default='')

@attr.s(cmp=False)
class ModelDifferent(Base):
    stuff = attr.ib(default='different!')

one = ModelOne()
same = ModelSame()
different = ModelDifferent()

assert one == same
assert one != different
assert same != different
```

Now I'd like to do the same behavior by making the class dynamically with `attr.make_class` (I have good reasons to do this, I'm actually writing a class decorator and building properties on it):

```python
attributes = {
    'stuff': attr.ib('')
}

# cannot use subclass, what do do?
one = attr.make_class('One', attributes)
same = attr.make_class('Same', attributes)

assert one == same  # AssertionError
```

By passing bases I can do this, but is there a cleaner way of doing this?

hynek:
Right.  

So the new functionality has to be tested to make sure it works and doesn’t break in the futures.  Simply write a test that creates a class that uses a custom subclass and you’re golden.   I would assume it also can be simplified by pulling the `(object,)` into the signature as a default argument.

Also you’re gonna want everyone to know about that new feature so you need to

1. add the argument to the docstring
2. add the change to the changelog.

Feel free to check out https://github.com/python-attrs/attrs/blob/master/CONTRIBUTING.rst if you’re unsure how to do that!

Thanks!

brainysmurf:
As far as using `(object,)` for the default argument, I got into the habit of always using `None`, but the point is actually to use immutable objects, which a tuple is.

brainysmurf:
Submitted for your approval :)

hynek:
Thank you for your contribution!  I’ve added some minor fixes I didn’t want to boss you around for. :)

PR #155, opened 2017-02-27
Preserve metaclass when slots=True

### LINKED ISSUE #154: Metaclass lost with slots=True
Hello.

Using a metaclass with slots=True, e.g.
```
@attr.s(slots=True)
class A(metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def f(self):
        pass
```
Will lead to `A` being instantiable and `type(A)` being `<class 'type'>` (unlike the case with slots=False), as for slots to work the new class is generated in [_make.py](https://github.com/python-attrs/attrs/blob/master/src/attr/_make.py#L365) with:

`cls = type(cls.__name__, cls.__bases__, cls_dict)` [1]

This doesn't happen if this line is changed to:

`cls = type(cls)(cls.__name__, cls.__bases__, cls_dict)` [2]

With this change `type(A)` is `<class 'abc.ABCMeta'>` and A is no longer instantiable.
Can the class generation for slots be changed to something like [2] to avoid losing metaclasses or it is intentionally written as it is now and I am missing something?

YevIgn:
Fixes #154

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,3 +3,4 @@ pytest
 zope.interface
 pympler
 hypothesis
+six
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -362,7 +362,7 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
             cls_dict.pop("__dict__", None)
 
             qualname = getattr(cls, "__qualname__", None)
-            cls = type(cls.__name__, cls.__bases__, cls_dict)
+            cls = type(cls)(cls.__name__, cls.__bases__, cls_dict)
             if qualname is not None:
                 cls.__qualname__ = qualname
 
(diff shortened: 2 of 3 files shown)


hynek:
fyi the tests fail on py2/pypy2

YevIgn:
Sorry, doubt I can fix the new test, without bringing analog of `six.add_metaclass` to the code, or using an ugly hack with `eval`.

Tinche:
Hm. Can we bring in six as a test dependency?

hynek:
I don’t mind.  I don't want runtime deps to check for PY3 but reimplementing metaclass magic just for tests seems like a bridge too far.

hynek:
Tin please take this; I have willfully forgotten everything I ever knew about meta classes.

Tinche:
Sure thing. I'm sick atm though so it'll have to wait a little

hynek:
Get well soon!

hynek:
I’ve updated the branch and it passes now.

Thanks for the input and I hope you feel better!

PR #156, opened 2017-02-27
update setup.py to permit 'python setup.py develop' workflow

reece:
See README.rst for explanation of changes.

We could use something like `TEST_REQUIRES = open("dev-requirements.txt").readlines()` to lessen duplication, but that file is not in MANIFEST.in, which means that it would fail when built from a package built by setup (I think).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,11 @@
-.tox
-.coverage*
-*.pyc
 *.egg-info
+*.pyc
+.cache
+.coverage*
+.eggs
+.hypothesis
+.tox
+dist
 docs/_build/
 htmlcov
-dist
-.cache
-.hypothesis
+venv
--- a/README.rst
+++ b/README.rst
@@ -71,6 +71,28 @@ Never again violate the `single responsibility principle <https://en.wikipedia.o
 
 .. -testimonials-
 
+
+Development
+===========
+
+To prepare a development environment, type::
+
+  pyvenv venv
+  pip install -U pip
+  python setup.py develop
+
+(If you don't have pyvenv installed, consider `virtualenv
+<https://virtualenv.pypa.io/>`__ or `virtualenvwrapper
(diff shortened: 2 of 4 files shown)


hynek:
I’m generally more than happy to see contributions that help people contributing but I see a few issues here:

- setup_requires is installed before *each* installation.  That means that attrs would depend on pytest-runner now no?
- tests_require is related to `python setup.py test` which we explicitly do not support (mostly because it doesn’t use pip for downloading/installing dependencies).  Is there any other way to use it?
- contribution information belongs into CONTRIBUTIONS.rst.  I suppose adding a note/link to README.rst would be a good thing though!

Tinche:
I'm kinda meh on things like this. Do we even want to support a setup.py based workflow? AFAIK `python setup.py develop` is deprecated in favor of `pip install -e .`.

When I was in the Java world this was an interesting idea to me because I was used to Maven being the front-end to basically all tasks (compiling, packaging, testing, dependencies). Now after several years working with Python my personal conclusion is that setup.py is a lousy interface for these kinds of operations. In fact I hate fiddling with setup.py, if there was a mature configuration file based alternative I'd switch my projects over.

I guess I'm -0 on this, if other maintainers like it and the work is already done, alright, but if we're just adding a second way of running pytest, what's the point.

hynek:
Yeah as I wrote: `python setup.py test` is a no-go.  The whole packaging community is moving away from it.  I *would* merge a good description of how to set up a local non-tox dev environment + a link from the README.

reece:
As a potential contributor, I think the only thing I care about really is knowing how to get started.  That's not obvious from the current docs (apologies if I missed it). As @hynek wrote, instructions in README.rst on setting up a non-tox dev environment would be very helpful.

Python packaging being what it is, there's unfortunately no truly standard way to admin a package. The closest that I know of is the `python setup.py {develop,install,test,bdist}` world, with all of its warts.

And, thanks for attrs... I'm making lots of use of it!

Issue #151, opened 2017-02-20
Support Python 3.6 attribute annotations.


Tinche:
Python 3.6 allows this:
```
@attr.s
class A:
    a: int = attr.ib()
```

The annotation is available as `A.__annotations__['a']`. So, I suggest that on 3.6, we add an attribute to `attr.Attribute` called `annotation` and expose these values there if they are present.

This might be an easy one for someone looking to get started contributing.

hynek:
Feel free to use the according label generously. :)

DRMacIver:
I'm having a look into this. Some brief thoughts and questions:

* I think it's important to support annotation on < 3.6 if you're going to support it on 3.6. How do you feel about an a = attr.ib(annotation=int) as an alternate API that does the same thing?
* As well as exposing it on the Attribute object this feels like it should really be made available as standard type annotations on various methods. I can think of three places where the information needs to be propagated: init, property setter, and property getter. Can you think of anywhere else?
* Propagating the type to methods is arguably a breaking change because it could cause code that currently type checks to fail type checking. Do you consider it to be one?
* If yes, hiding it behind an off by default propagate\_types flag would be sufficient. This seems like a good thing to have anyway, though my preference would be on by default.
* How do you want to test Python 3.6 specific code? Currently your tests are all in a single undifferentiated directory.

DRMacIver:
Actually, here's a more fundamental problem with this proposal which I think means it can't/shouldn't work as written: It's flat out lying to the type checker.

In the provided example you've declared that A.a should be an int, but it's actually an Attribute object. Thus correct code using it won't type check, and incorrect code using it will.

hynek:
Seems like we need a different approach.

Tinche:
> In the provided example you've declared that A.a should be an int, but it's actually an Attribute object. Thus correct code using it won't type check, and incorrect code using it will.

Actually, I'd have to had used `ClassVar` to have done that. In reality, `A.a` won't be an Attribute object but won't exist if we're dealing with dict classes or be a special descriptor if dealing with slot classes. `A(1).a` will be an int, though, and this is what's claimed by the annotation.

Also note this example from the official Python docs:
```
class Employee(NamedTuple):
    name: str
    id: int
```
Lying to the type checker too?

hynek:
Uh I'm sorry it looked uncontroversial to me.  I'll have a deeper look in the next days.  Please everyone make your points. :)

hynek:
[...]

PR #158, opened 2017-02-27
list of validator

reece:
From the new docs:
"A validator that raises a TypeError if the initializer is called with a non-list type or with a list that contains one or more elements of a wrong type. None values are not permitted. As with instance_of, checks are perfomed using isinstance() therefore it’s also valid to pass a tuple of types."

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,7 +49,10 @@ Changes:
 - Validators can now be defined conveniently inline by using the attribute as a decorator.
   Check out the `examples <https://attrs.readthedocs.io/en/stable/examples.html#validators>`_ to see it in action!
   `#143 <https://github.com/python-attrs/attrs/issues/143>`_
-
+- New ``list_of`` validator ensures that an attribute is a homogeneous list with elements of a specified class or subclass.
+  See `examples <https://attrs.readthedocs.io/en/stable/examples.html#validators>`_.
+  `#157 <https://github.com/python-attrs/attrs/issues/157>`_
+  
 
 ----
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -257,6 +257,36 @@ Validators
       TypeError: ("'x' must be <type 'int'> (got None that is a <type 'NoneType'>).", Attribute(name='x', default=NOTHING, validator=<instance_of validator for type <type 'int'>>, repr=True, cmp=True, hash=None, init=True), <type 'int'>, None)
 
 
+.. autofunction:: attr.validators.list_of
+
+
+   For example:
+
+   .. doctest::
+
+      >>> @attr.s
+      ... class C(object):
+      ...     x = attr.ib(validator=attr.validators.list_of(int))
+
+      >>> C([42])
+      C(x=[42])
+
+      >>> C(42)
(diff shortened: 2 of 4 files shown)


hynek:
OK so let’s do this.

But if we tackle this, I would suggest: `container_of(container_type, element_type, empty_ok=True)` with a bunch of convenience helpers `list_of()`, `set_of()`, `tuple_of()`…

hynek:
> empty_ok=True means that the container can be empty, right?

yep!

> What about also adding none_ok, permitting None values? Or could optional() be used for elements?

I’d say that exactly what optional() is for. :)  So yes, you’ll have to handle that case.

That reminds me: unless someone has good arguments, I don’t think we should do `element_type` but `validator`!  And `validator` needs also to take a list. :)

> What about mappings?

Baby steps. :)

reece:
I see what you did there with the "you'll need to..." !

I appreciate attrs, but I'm not up for implementing this now. You might prefer to close this PR and migrate comments to #39.

hynek:
No problem!  This is kind of the reason, why we still don’t have it: it’s more work than it looks initially if you want to cover all bases. :-/

wsanchez:
This feels like a bit of reinventing the type annotation system in Py3.

If I understand this correctly, the feature is `validator=list_of(int)`.

Why not `validator=instance_of(List[int])`?

That is, I think I'd rather write this code:

```python
from pathlib import Path
from typing import Iterable

from attr import attrib, attrs


@attrs()
class ThingWithPaths(object):
    _paths = attrib(validator=is_type(Iterable[Path]))

    def exists(self):
        for path in self._paths:
            if not path.exists():
                return False
        return True

thing1 = ThingWithPaths([Path("."), Path("foo")])
print("thing1 exists:", thing1.exists())

thing2 = ThingWithPaths([".", "foo"])
print("thing2 exists:", thing2.exists())
```

So what's missing is `is_type`, instead of a series of `foo_of` validators.

Another advantage here is that if we have can find the type annotation (eg #151), then is could even be simpler:  `_paths = attrib(validator=typing)` where the type for the validator to check is obtained from the annotations.

IlyaSkriblovsky:
@wsanchez, did you tried to execute your code? :)
`typing` doesn't work for dynamic checking using `isinstance`:
```python
>>> isinstance(['qwe', 'asd'], List[int])
True
```

wsanchez:
@IlyaSkriblovsky Yeah, I'm figuring that out, which is weak sauce but that's why I changed the code to `is_type` from `instance_of`.  It wasn't meant to be a thing that already works so much as what would be more awesome.  :-)

I'll add here a reference to a ticket to fix that, which you just sent to Glyph: https://github.com/python/typing/pull/377

I guess it's just unfortunate not to be able to leverage typing better.  From your work on cattrs, I got the impression that it was possible, but that was optimistic, I guess.

hynek:
So my understanding is that `is_type` with py3 syntax is not feasible anytime soon?  In any case, as much as I loath Python 2, I’m sure a Python 3-only option would not stop people opening an issue and asking for this every week. :)

So I guess we still need a solution for all the legacy-impaired.

Closing this at the request of the reporter tho.

Issue #97, opened 2016-09-21
Support for annotations (PEP 484 type hints)


carljm:
Hi Hynek!

I'd like the generated `__init__` signature to include type annotations in Python 3.

Somebody mentioned this in passing over in #63, but after reading through that ticket and glancing at the PR, I'm still not sure how that feature would let me actually get type annotations into the generated `__init__` signature. It seems like this requires dedicated support in attrs, probably as a separate kwarg to `attr.ib`.

Any thoughts?

hynek:
My thoughts are that that makes sense to have and shouldn’t be *too* hard to achieve.

The question is how to integrate it.  We could pull the type information from the instance_of validator or bite the bullet and introduce a full-fledged on type field that would also help with the deserialization efforts.

carljm:
I'll have to think a bit about those options. It seems really strange for attr.ib "core" to pull info like that out of a special validator (it's like adding some kind of implicit side-channel to the validator API that only this one special validator gets to use). Also, PEP 484 type hints aren't necessarily expressed the same as runtime types (e.g. for optional types). On the other hand, it's obviously closely related, and would be ugly to have to specify both, especially when they are the same.

The other problem with this whole idea is that PEP 484 hints are mostly for static analysis, and having annotations in import-time generated code isn't going to help those analyzers at all :-) So even with support in attrs to make these annotations available at runtime, making use of them with a type analyzer would require something else (an extension to the static analyzer to understand `attr.ib`? A generator for type stub files for `attr` classes?)

Tinche:
Another thought to consider. In 3.6, this will be possible IIUC:

```
@attr.s
class A:
    a: int = attr.ib()
```

I think it would be a great waste not to expose this type information somehow. Probably make `attr.Attribute` have a `type` property? In that case we might want to allow versions earlier than 3.6 to set this property somehow too, maybe via an attr.ib() argument.

In any case, once we have this info, it's not hard to add the annotations to the generated init if we want.

kmike:
Before Python 3.6 annotating attributes was also possible; it is in PEP 484:

```py
@attr.s
class A:
    a = attr.ib(default=None)  # type: int
```

PyCharm seems to understand such annotations; A().a inferred type is `int` in PyCharm (likely it just thinks this is a regular class attribute).

[...]

PR #159, opened 2017-03-02
Public API to check a declared attribute type and to check if it is mandatory/optional

smarie:
Hello everyone,

First of all thanks a lot for this great library ! I am sad I found it only a couple days ago, otherwise I would probably not have spent time writing [classtools-autocode](https://github.com/smarie/python-classtools-autocode). Indeed its main idea was to remove some of the boilerplate and link with [PyContracts](https://andreacensi.github.io/contracts/) for validation.

Anyway, today I make this pull request because in another project of mine currently in development mode, [parsyfiles](https://github.com/smarie/python-parsyfiles/tree/develop), I perform some type introspection to understand what is required to parse an object. I use PEP484 type hints for that, but I wanted to support attrs too so I implemented something that does the job, and now I am proposing it to you.

The proposal is made of two functions : 'guess_type_from_validators', and 'is_mandatory'. Feel free to change the naming, I am not particularly good at finding great names.

For 'is_mandatory': it is simply checking if the validator is an instance of 'optional'.

For 'guess_type_from_validators': type information is extracted from the 'instance_of' validator. In order to help users rely on this validator instead of doing the same type checking in their custom validators, I introduce the 'chain' validator. Note that this might now be useless, since I saw in the source that you allow lists of validators.
In the future if [issue 97](https://github.com/python-attrs/attrs/issues/97) is resolved, the 'guess_type_from_validators' method should be updated to also use the new python 3.6 type hints as an additional source of type hint.

I added some unit tests for all the code, and updated the doc in api.rst.

Let me know what you think.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -149,6 +149,42 @@ Helpers
       False
 
 
+.. autofunction:: attr.validators.is_mandatory
+
+   For example:
+
+   .. doctest::
+
+      >>> @attr.s
+      ... class C(object):
+      ...     x = attr.ib(validator=attr.validators.instance_of(int))
+      ...     y = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(int)))
+      >>> attr.validators.is_mandatory(attr.fields(C).x)
+      True
+      >>> attr.validators.is_mandatory(attr.fields(C).y)
+      False
+
+    ..  versionadded:: ??? (to complete)
+
+
+.. autofunction:: attr.validators.guess_type_from_validators
+
+   For example:
+
+   .. doctest::
+
+      >>> @attr.s
+      ... class C(object):
+      ...     x = attr.ib(validator=attr.validators.instance_of(int))
+      ...     y = attr.ib()
+      >>> attr.validators.guess_type_from_validators(attr.fields(C).x)
+      int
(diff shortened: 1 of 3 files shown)


hynek:
So the good news is that `chain` is not necessary anymore: as of the (unreleased) 17.1, you can pass a list of validators and it just works.

`guess_type_from_validators` is certainly very useful but I’m afraid a bit too hacky to be part of an official API (I do encourage anyone to release their own validators and other attrs-related tools tho!  I just try to keep attrs slim and opinionated.  I prefer no API in core over an API that contains the word `guess` :)).

`is_mandatory` opens two interesting questions:

1. `optional()` should be able to consume lists like `validator=` does. (filed as #161)
2. where to put it? It’s more introspection than a validator…

smarie:
Thanks for the quick review @hynek, 
Agreed that you can get rid of `chain`.. if `optional()` is able to consume lists.
I will keep my 'hacky guess' method in the parsyfiles project, no problem about it not being part of the public attr API - indeed that was borderline.

Tinche:
Pulling metadata from validators is something I've played around with and ultimately I think is just too hacky. It won't work with other validators, what about using converters instead of validators, etc.

@smarie I would advise using metadata.

hynek:
Yeah so what Tin said.  Thanks for sharing your work but this is nothing for attrs proper.

PR #163, opened 2017-03-06
Document another reason why not namedtuples

qdamian:
Hi,

Just a minor addition to the docs. This was why I opted for attrs vs namedtuples.

Damian


hynek:
Maybe it would be better to stress that attrs leaves the *choice* to the user?  Because you can totally have immutable classes using `frozen=True` but you don’t *have* to.

qdamian:
You're right. I've pushed another patch.

hynek:
thanks!

Issue #160, opened 2017-03-02
Allowing to transform the attributes in properties to implement validation in setters


smarie:
Dear attrs community members

One thing I implemented in [autoclass](https://github.com/smarie/python-autoclass) that I feel is missing in `attrs` is the ability to automatically generate the property setter that includes validation. This might be optional, for example

```python
import attr
from attr.validators import instance_of
@attr.s(validation_setters=True)
class C(object):
     x = attr.ib(validator=instance_of(float))
```

would transform x into a property (+ a private field) behind the scenes, equivalent to

```python
@attr.s(these={'_x': attr.ib(validator=instance_of(float))})
class C(object):

    @property
    def x(self):
        return self._x

    @x.setter
    def set_x(self, val):
        x_attr = attr.fields(C).x
        x_attr.validator(self, x_attr, val)
        self._x = val
```

Note that the above code right now throws an `AttributeError: can't set attribute` so this would need to be fixed first.

Do you think that it would make sense to add this feature ?

Thanks !

hynek:
Well there have been quite a bit of requests that in the end ask for that: frozen single attributes and validation on setting.   We haven’t quite decided which route to take though. :|

Tinche:
I think this is an interesting idea and we should probably support it eventually. Users are constantly surprised there's no validation when setting the attributes.

The main problem is it will make accessing the attribute slower too.

hynek:
At some point we’ll have to write a similar monstrous generator for `__setattr__` as we now have for `__init__`. :(

smarie:
Do you need help (=contribution) for this feature to be added ? I can have a look at what you did to generate `__init__` and replicate for `__setattr__`. Just let me know if that work started or no yet, so that I know if I have to start from an existing branch or not. 
regards
Sylvain

hynek:
I think this is a bit dependent on #172

smarie:
Makes sense if we want to be consistent - thanks Hynek, I'll follow that topic closely then.

jsmolina:
It could be even easier just using setattr behaviour, isn't it?

```
def __setattr__(self, key, value):
        """
        This method will magically add the normal validation to from_dict
        :param key:
        :param value:
        :return:
        """
        x_attr = getattr(attr.fields(self.__class__), key)
        if x_attr.validator:
            x_attr.validator(self, x_attr, value)

        super(MyAttrClassBase, self).__setattr__(key, value)
```

smarie:
[...]

PR #172, opened 2017-03-28
[WIP] Frozen attribute support

Tinche:
WIP. Do not merge yet :)

Got initial frozen attribute support working for dict classes using the magic of ✨multiple inheritance✨.

Now to think of something clever for slot classes.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -351,7 +351,7 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
             cls.__hash__ = None
 
         if init is True:
-            cls = _add_init(cls, effectively_frozen)
+            cls = _add_init(cls, slots, effectively_frozen)
         if effectively_frozen is True:
             cls.__setattr__ = _frozen_setattrs
             cls.__delattr__ = _frozen_delattrs
@@ -370,6 +370,17 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
             cls = type(cls)(cls.__name__, cls.__bases__, cls_dict)
             if qualname is not None:
                 cls.__qualname__ = qualname
+
+            ca_attrs = fields(cls)
+            for ca_attr in ca_attrs:
+                if not ca_attr.frozen:
+                    continue
+                ca_name = ca_attr.name
+                if init and ca_attr.init:
+                    setattr(cls, '__{ca_name}_slot'.format(ca_name=ca_name),
+                            getattr(cls, ca_name))
+                setattr(cls, ca_name,
+                        FrozenSlotDescriptor(getattr(cls, ca_name)))
 
         return cls
 
@@ -536,7 +547,7 @@ def _add_repr(cls, ns=None, attrs=None):
     return cls
 
 
-def _add_init(cls, frozen):
+def _add_init(cls, slots, frozen):
     """
     Add a __init__ method to *cls*.  If *frozen* is True, make it immutable.
     """
@@ -552,6 +563,7 @@ def _add_init(cls, frozen):
 
     script, globs = _attrs_to_script(
         attrs,
+        slots,
         frozen,
         getattr(cls, "__attrs_post_init__", False),
     )
@@ -649,7 +661,7 @@ def validate(inst):
             v(inst, a, getattr(inst, a.name))
 
 
-def _attrs_to_script(attrs, frozen, post_init):
+def _attrs_to_script(attrs, slots, frozen, post_init):
     """
     Return a script of an initializer for *attrs* and a dict of globals.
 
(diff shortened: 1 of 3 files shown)


Tinche:
Alright, we support slot classes too now. the information below is valid for slot classes.

There is zero overhead for `__init__`. 

```
pyperf timeit --rigorous -g --duplicate 10 -s "import attr; B = attr.make_class('B', {'a': attr.ib(frozen=True), 'b': attr.ib(frozen=True)}, slots=True)" 'B(1, 2)'

Before: Mean +- std dev: 410 ns +- 11 ns
After: Mean +- std dev: 409 ns +- 11 ns
```

Unfortunately, attribute access is much slower:
```
pyperf timeit --rigorous -g --duplicate 10 -s "import attr; B = attr.make_class('B', {'a': attr.ib(frozen=True), 'b': attr.ib(frozen=True)}, slots=True); b = B(1, 2)" 'b.b'

Before: Mean +- std dev: 35.8 ns +- 0.8 ns
After: Mean +- std dev: 391 ns +- 17 ns
```

We wrap the original slot descriptor to block mutation, but we pass through access. However, this involves a pure Python method invocation. I don't know how to avoid this :(

hynek:
Nice, factor 10. :|  I don’t think adding RCEs (i.e. write some C) would really help here in any way?  Very generally speaking, I could live with an optional C to speed things up I guess.  I’m just not eager for it.

Tinche:
It would probably help. We would need a customized version of this: https://github.com/python/cpython/blob/fff9a31a91283c39c363af219e595eab7d4da6f7/Objects/descrobject.c#L594

Can Cython help us out here somehow?

hynek:
I'd hope so.  I'm not adding actual C.

Tinche:
Huh, playing around with Cython a little, managed to get the overhead to just ~3x. 25.8 ns to 74.7 ns.

The numbers are a little different since I'm on my work computer atm.

hynek:
Interesting.

Generally I want to give everyone maximum flexibility so the Cython code would go into a separate package (attrs-accel?) and everything has to be implemented also in pure Python.  I bet it’s faster on PyPy anyways. :)

Tinche:
I've left a post on the cython-users mailing list asking for a little advice.

In the mean time, fiddling around, I believe I've managed to get a Cython solution with only 5% overhead (27 ns vs 26 ns).

Tinche:
Here's the link to the post since it finally got approved, just to be thorough. https://groups.google.com/forum/#!topic/cython-users/QaQDNLE-76Y

Tinche:
Ok, confirming that the speed penalty for frozen slot attributes with the Cython accelerator is within 10%.

Non-frozen access: 35.3 ns +- 1.1 ns
Cython frozen access: 38.5 ns +- 1.2 ns
Python frozen access: 368 ns +- 15 ns

I think this makes the idea of having some Cython code in attrs worthwhile, and when we have this infrastructure in place we might want to offer accelerated versions of other components too. I agree all Cython code should be an optimized reimplementation of Python code - native code should be strictly optional.

How do we go about this though? We can have a separate package for Cythonized components, but having a single sdist that cleanly installs even without a C compiler (and just falls back to Python versoins) would be less maintenance - is it possible? We will also need to get Travis to test both versions of the code, and make wheels where we can.

Somewhat unrelated: it occured to me that for frozen dict classes, we can generate a slightly different `__init__`:
```
def __init__(self, a, b):
    __dict__ = self.__dict__
    __dict__['a'] = a
    __dict__['b'] = b
```
This is functionally identical to the way we do it now, with the cached object.__setattr__, right? This approach totally avoids the `__init__` speed penalty for frozen dict classes.

hynek:
This is pretty amazing.  That’s gonna be the flagship feature of attrs 17.2 I reckon.

Maybe we should brainstorm some what makes sense to accelerate etc.

smarie:
@Tinche , just out of curiosity what would be the overhead of transforming all attributes that require either frozen access or have convertors and/or validators, into python `@property` ? It is fairly easy to add a property to a class :

```python
setattr(clazz, attr_name, property(fget=getter_fun, fset=setter_fun))
```

With `getter_fun` being for example a function or lambda with direct access to the field
```python
getter_fun = lambda self: getattr(self, private_attr_name)
```

And `setter_fun` being a function or lambda including frozen access check or conv + validators as usual for non-frozen attrs - the same you have to day in the `__init__` .
Properties seem to now exist for both python 2 and 3.

I don't know how to make that consistent with the base `Attribute` class, maybe `Attribute` could directly inherit from `property`? Or a flavour of attribute (`PropertyAttribute`) could inherit.
Let me know if that is just stupid, I'm relatively new to Python..

(relates to requirement [160](https://github.com/python-attrs/attrs/issues/160))

Tinche:
It's not stupid, the problem is the getter is slow - it involves a pure Python method call.

The current approach is similar to `@property`, but lower level. `@property` under the hood creates a descriptor which invokes the getter and returns the value; we create a descriptor that wraps the existing descriptor on a very low level and delegates getting to it.

smarie:
ok, thanks for the clarification !

Infernion:
Is there any progress on this feature?

kdggavkc:
Bump? Enjoyed reading the history of this request - is this still in the plan for attrs?

hynek:
Closing due to lack of activity (and at this point: unmergeablity – le churn).

Issue #171, opened 2017-03-28
treatment of private attributes is obscure, nearly undocumented, perhaps not ideal in all cases


glyph:
Some folks, presented with the `attr.s` API, make the entirely reasonable inference that this:

```python
@attr.s
class Foo:
    _bar = attr.ib()
```

is a class you should use like this:

```python
Foo(_bar=1)
```

This is further exacerbated by the fact that `Foo`'s `repr` actually shows you _exactly that string_.

The only example in the docs that I can find is [the code example at the bottom of the `attr.s` API doc](http://attrs.readthedocs.io/en/stable/api.html?#attr.s).

This behavior should be thoroughly documented and explained, and, um, maybe tweaked a little bit, since there are some issues with it.  As it stands now, for example, there's no publicly-named attribute for API doc generation tools, or code-completion things, to pick up to present as the constructor argument.

If I could take a stab at the rationale for this behavior, it's something like this:

> In Python, if you have a function with "private" arguments, they all have _positional names_, which makes them effectively public.  Therefore, there is no point in trying to "hide" your constructor arguments behind `_` prefixes, since a user might happily do `Foo(1)` and never realize that they've done something unsupported.  Therefore `attr.s` always makes the keyword-arg names of your constructor arguments "public" by removing their underscores; the attributes themselves are of course private and can still have the `_` prefix when accessed as `self._something`.

Except, in Python 3, _this isn't true_ any more, because the advent of keyword-only arguments means you _can't_ pass the argument without actually having to type in an `_` in a namespace you don't own, which means it's totally valid to have internal "don't pass me this in the constructor" state.  IMHO this is great - it's a way to force public users through a class-method constructor or factory function without hiding the type itself (and all the methods on it) or resorting to any gross shenanigans to make the constructor literally private.  So it turns out I really would like "private attributes" that don't show up in the constructor either!  (And of course, at the metaprogramming level where Attrs itself lives, the kwarg-only behavior can be emulated on python 2.)

hynek:
As a little history lesson, the origin of the behavior is https://github.com/hynek/characteristic/issues/6.

glyph:
To be clear - especially given that history :) - _I_ find this behavior entirely intuitive, and have not previously found it to be problematic.  However, having had it pointed out to me that it's confusing, I found it reasonable to expect some other behavior unless you have a view of private attributes with very specific nuances.

Tinche:
[...]

PR #176, opened 2017-04-03
BUG: fix evolve behaviour when a validator fails.

### LINKED ISSUE #175: Wrong error message when using evolve
I am looking at the upcoming new `evolve` feature, but I found a bug in the error handling of that function when a validator raises an exception:

``` python
from attr import evolve, attributes, attr

@attributes
class Foo(object):
    bar = attr(validator=instance_of(int))

foo1 = Foo(1)
foo2 = evolve(foo1, bar="2")
```

which gives

``` python
  File "attr-bug.py", line 9, in <module>
    foo2 = evolve(foo1, bar="2")
  File "/home/cournape/src/python/attrs/src/attr/_funcs.py", line 221, in evolve
    "{k} is not an attrs attribute on {cl}.".format(k=k, cl=cls))
attr.exceptions.AttrsAttributeNotFoundError: bar is not an attrs attribute on <class '__main__.Foo'>.
```

which is obviously not true. Looking at the error handling in `evolve`, I think it is too dangerous to assume the reason for `TypeError` like it does. There is a lazy solution to this I will explain in a PR, but there may be better options

cournape:
This is a naive solution to #175. TypeError may be coming from many sources, and it is not correct to assume it always comes from a non-existing attr member as the evolve codes now. Instead, we now check specifically for this case when raising a custom error, and just re-raise in other cases.

Fixes #175

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -218,8 +218,8 @@ def evolve(inst, **changes):
     except TypeError as exc:
         attr_dict = {attr.name: attr for attr in fields(cls)}
         for name in changes:
-            if not name in attr_dict:
+            if name not in attr_dict:
                 k = exc.args[0].split("'")[1]
                 raise AttrsAttributeNotFoundError(
-                    "{k} is not an attrs attribute on {cl}.".format(k=k, cl=cls))
+                    "{} is not an attrs attribute on {}.".format(k, cls))
         raise
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -201,7 +201,8 @@ def evolve(inst, **changes):
     ..  versionadded:: 17.1.0
     """
     cls = inst.__class__
-    for a in fields(cls):
+    attrs = fields(cls)
+    for a in attrs:
         attr_name = a.name  # To deal with private attributes.
         if attr_name[0] == "_":
             init_name = attr_name[1:]
@@ -216,9 +217,8 @@ def evolve(inst, **changes):
     try:
         return cls(**changes)
     except TypeError as exc:
-        attr_dict = {attr.name: attr for attr in fields(cls)}
         for name in changes:
-            if name not in attr_dict:
+            if getattr(attrs, name, None) is None:
                 k = exc.args[0].split("'")[1]
                 raise AttrsAttributeNotFoundError(
                     "{} is not an attrs attribute on {}.".format(k, cls))
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -216,6 +216,10 @@ def evolve(inst, **changes):
     try:
         return cls(**changes)
     except TypeError as exc:
-        k = exc.args[0].split("'")[1]
-        raise AttrsAttributeNotFoundError(
-            "{k} is not an attrs attribute on {cl}.".format(k=k, cl=cls))
+        attr_dict = {attr.name: attr for attr in fields(cls)}
+        for name in changes:
+            if not name in attr_dict:
+                k = exc.args[0].split("'")[1]
+                raise AttrsAttributeNotFoundError(
+                    "{k} is not an attrs attribute on {cl}.".format(k=k, cl=cls))
+        raise
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -24,6 +24,9 @@ from attr import (
 )
 
 from attr.exceptions import AttrsAttributeNotFoundError
+from attr.validators import instance_of
+from attr._compat import TYPE
+
 
 MAPPING_TYPES = (dict, OrderedDict)
 SEQUENCE_TYPES = (list, tuple)
@@ -464,3 +467,16 @@ class TestEvolve(object):
         assert (
             "aaaa is not an attrs attribute on {cls!r}.".format(cls=C),
         ) == e.value.args
+
+
+    def test_validator_failure(self):
+        """
+        Make sure we don't swallow TypeError when validation fails within evolve
+        """
+        @attributes
+        class C(object):
+            a = attr(validator=instance_of(int))
+
+        with pytest.raises(TypeError) as e:
+            evolve(C(a=1), a="some string")
+        assert e.value.args[0].startswith("'a' must be <{type} 'int'>".format(type=TYPE))


Tinche:
So this is error handling case so performance doesn't really matter here, but there might be a more efficient way.

Instead of constructing a dictionary, could you try something like (pseudocode):
```
attrs = fields(cls)
for name in changes:
    if getattr(attrs, name, None) is None:
        raise AttrsAttributeNotFoundError
```
Have to run now, will look again later.

src/attr/_funcs.py
```diff
@@ -216,6 +217,9 @@ def evolve(inst, **changes):
     try:
         return cls(**changes)
     except TypeError as exc:
-        k = exc.args[0].split("'")[1]
-        raise AttrsAttributeNotFoundError(
-            "{k} is not an attrs attribute on {cl}.".format(k=k, cl=cls))
+        for name in changes:
+            if getattr(attrs, name, None) is None:
+                k = exc.args[0].split("'")[1]
```
Tinche: I don't think we need this any more. We have the name of the attribute already in `name`.

cournape: indeed, I made the change mechanically without thinking, sorry.

cournape:
So there is actually a problem with this solution, because `changes` is mutated to handle private args. What is the desired behaviour here ? E.g. which ones should work/fail in the following:

``` python
import pytest

from attr import attributes, attr, evolve
from attr.exceptions import AttrsAttributeNotFoundError
from attr._compat import TYPE


@attributes
class C(object):
    a = attr(validator=instance_of(int))
    _b = attr(validator=instance_of(int))


c1 = C(1, 2)
# Should work ?
c2 = evolve(c1, b=2)
# Should work ?
c3 = evolve(c1, _b=2)
# Should that really work ??
c4 = evolve(c1, _b=3, b=4)
```

Tinche:
Huh, good catch. There was a small discussion around whether to support the setter syntax or the init syntax over at #116. We settled on supporting both, since I figured it was easy to do and there's no ambiguity in the attribute names (i.e. you can't have `_b` and `b` attributes). Let's revisit this.

`c2 = evolve(c1, b=3)` not working at all is definitely not good. The fix depends on the following issue.

I don't actually know what the right way is to resolve `evolve(c1, _b=3, b=4)`. The fact this is not obvious to me is kind of an argument in favor of not supporting both syntaxes. Since this is a design issue and not an implementation thing let's ask @hynek for his opinion.

Given the additional discussion of private attributes from #171, I'm in favor of only supporting the setter syntax in evolve. (Summoning @glyph here as well.) So only `evolve(c1, _b=3)` would be valid, all the others would throw.

Tinche:
This code would be a simple fix I think.

```
        if attr_name[0] == "_":
            init_name = attr_name[1:]
            if attr_name in changes:
                # attr_name (not init_name) in changes, translate
                changes[init_name] = changes.pop(attr_name)
            elif init_name not in changes:
                changes[init_name] = getattr(inst, attr_name)
            # else: init_name is already in changes, leave it there
        else:
            if attr_name not in changes:
                changes[attr_name] = getattr(inst, attr_name)
```

But it still doesn't fix `evolve(c1, b=3, _b=4)`.

cournape:
Fixing should be fairly easy once we agree on what the behaviour should be. Detecting `evolve(b=.., _b=...)` should be straightforward (detect dict length change), and then you can do more work to figure out which attributes are given "twice" in the `except` block.

hynek:
Is this blocking on something?

cournape:
Can you comment on https://github.com/python-attrs/attrs/pull/176#issuecomment-291461219 ?

hynek:
Paging @glyph because he expressed opinions on this exact topic recently.

glyph:
I think the question of which syntax to use actually cuts to the heart of a serious design problem with `evolve`, which is: `evolve` really can't reasonably be a public API.

Consider this case:

```python
import attr
@attr.s
class _Tree(object):            # private type
    _parent = attr.ib(default=None)
    _children = attr.ib(default=attr.Factory(list))
    def add(self, child):
        if child._parent is not None:
            child._parent = self
        self._children.append(child)
def tree(children=()):          # public constructor
    new = _Tree()
    for child in children:
        new.add(child)
# ... application ...
leaf = tree()
branch = tree(leaf)
root = tree(branch)
```

If you call `evolve()` on one of these objects, and pass `parent` (or `_parent`) or `child` but not both, you're creating an inconsistent state, and nothing really told you that you were doing that.

The reason I'm raising this here is that the question as to whether to give `evolve` parameters underscores or not is the question as to whether the parameters are considered "same 'public' status as `__init__` arguments" or "same 'public' status as attributes".  Except it turns out that it doesn't matter because `evolve` is apparently public even when the attributes and constructor are *both* private.

If `evolve` itself remains public, I think that dropping the underscores makes sense, as it's effectively a "void the warranty" type API, where you should either call it on `self` or objects where you explicitly know it's safe (immutable value types, for example).

However, if we did something like this:

```python
@attr.s
class Point3D:
    x = attr.ib()
    y = attr.ib()
    z = attr.ib()
    evolve = attr.evolver()

Point3D(1, 2, 3).evolve(y=7)
```

so that classes could explicitly "opt in" to evolvulation then it would make sense to use the private attribute names in that case.

I suspect that this view may be controversial though. Thoughts?

hynek:
My thoughts at this point is that *nobody* building APIs should even expose the fact their classes are attrs-based.  It has always been my focus to make attrs transparent to endusers.

Honestly, I’ve personally even arrived at the point where I treat `__init__`s as private APIs but I think that’s a bridge too far for most. :)

In my mind we have two conflicting arguments:

1. `__init__` strips underscores because it’s potentially a public API – making `evolve()` behave the same way would be nice
1. `attr.evolve()` should never be part of a public API of anyone so it really should *not* strip.  It’s like a setter for FP-connoisseurs.

Did I forget something?  My mind is wandering towards nr. 2 **or**

s4eamxv

glyph:
My inclination is to say that *for now*, we should go ahead and make it consistent with `__init__`, since the semantics are "make a new thing like this, except...", and "make a new thing like this" is `___init__`, and making it behave differently from `__init__` is just confusing.  (Perhaps `__init__` should accept underscored and non-underscored kw variants as well, I don't have as strong an opinion about that; but however it works, it should be the same between them.)

But, the documentation should be adjusted to make it clear that the first argument to `evolve` must be a namespace that the caller controls, and that just like typing `._`, typing `evolve` voids your warranty.

The only real problem with this is that sometimes the constructor has been made private through some other shenanigans, but `evolve` made it public again indirectly through the instance.  Upon further reflection, this problem goes deeper than `evolve` itself: consider `type(x)(args, for, new, x)`.  No underscores typed there either.

Rather than try to hack up `evolve` to make this continue to appear to work, maybe let's discuss first-class support for private constructors on a different issue, and come up with a way that `fields` and `evolve` should interact with that separately.

hynek:
OK so back to my original hunch: `evolve()` shall behave like `__init__` as well as possible.

hynek:
JFTR @cournape , we have an agreement above. :)   This needs to be fixed before 17.1, so I’d be super grateful if you could finish it up!  No hurry though, there’s still some stuff open.  Just wanted to stress that this should be unblocked (and if not, please yell at us).

Tinche:
Thank you for your efforts @cournape, I've added a few commits and made PR #182 to get this done.

PR #182, opened 2017-05-02
Fix evolve (cont)

Tinche:
This is a continuation of https://github.com/python-attrs/attrs/pull/176.

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -204,16 +204,9 @@ def evolve(inst, **changes):
     attrs = fields(cls)
     for a in attrs:
         attr_name = a.name  # To deal with private attributes.
-        if attr_name[0] == "_":
-            init_name = attr_name[1:]
-            if attr_name not in changes:
-                changes[init_name] = getattr(inst, attr_name)
-            else:
-                # attr_name is in changes, it needs to be translated.
-                changes[init_name] = changes.pop(attr_name)
-        else:
-            if attr_name not in changes:
-                changes[attr_name] = getattr(inst, attr_name)
+        init_name = attr_name if attr_name[0] != "_" else attr_name[1:]
+        if init_name not in changes:
+            changes[init_name] = getattr(inst, attr_name)
     try:
         return cls(**changes)
     except TypeError as exc:
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -193,8 +193,8 @@ def evolve(inst, **changes):
 
     :return: A copy of inst with *changes* incorporated.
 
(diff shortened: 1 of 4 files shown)


Tinche:
One minor nitpick. In case of a non-existent attribute evolve will raise `AttrsAttributeNotFoundError`.

I suggest this is unnecessary. `__init__` doesn't raise a special attrs exception, just a TypeError, let's let evolve do the same (since it's kind of a pseudo `__init__` anyway).

The code will be simpler, and we can avoid a confusing error message like this:

```
@attr.s
class A:
    _a = attr.ib()

>>> attr.evolve(A(1), _a=2)
attr.exceptions.AttrsAttributeNotFoundError: _a is not an attrs attribute on <class '__main__.A'>.

File: tests/test_funcs.py [hunk 1]
```diff
@@ -464,3 +469,33 @@ def test_unknown(self, C):
         assert (
             "aaaa is not an attrs attribute on {cls!r}.".format(cls=C),
         ) == e.value.args
+
+    def test_validator_failure(self):
+        """
+        Make sure we don't swallow TypeError when validation fails within
```
hynek: Time to re-read https://jml.io/pages/test-docstrings.html ! :)

tests/test_funcs.py
```diff
@@ -464,3 +469,33 @@ def test_unknown(self, C):
...
+            a = attr(validator=instance_of(int))
+
+        with pytest.raises(TypeError) as e:
+            evolve(C(a=1), a="some string")
+        m = e.value.args[0]
+        assert m.startswith("'a' must be <{type} 'int'>".format(type=TYPE))
+
+    def test_private(self):
+        """
+        evolve() should act as `__init__` with regards to private attributes.
```
hynek: Better, but still not quite https://jml.io/pages/test-docstrings.html :)

src/attr/_funcs.py
```diff
@@ -201,21 +201,18 @@ def evolve(inst, **changes):
...
+            changes[init_name] = getattr(inst, attr_name)
     try:
         return cls(**changes)
     except TypeError as exc:
-        k = exc.args[0].split("'")[1]
-        raise AttrsAttributeNotFoundError(
-            "{k} is not an attrs attribute on {cl}.".format(k=k, cl=cls))
+        for name in changes:
+            if getattr(attrs, name, None) is None:
+                k = exc.args[0].split("'")[1]
```
hynek: This should be explained in a comment.

glyph:
I've only had time to glance at it (I definitely won't have time to do a full review), but I generally agree. This should probably land.

Tinche:
Review fixes done!

Tinche:
🎉

PR #189, opened 2017-05-13
Add takes_self to Factory and @_CountingAttr.default

### LINKED ISSUE #165: attr default based on other attributes
Hi!

I have unsucessfully tried to define a default value by referencing other attributes. I'm sure the code below doesnt' work for some obvious or fundamental reason, but I would be grateful for comments on how to do something like it:

```python
import attr
import datetime

@attr.s
class Something:
    some_date = attr.ib(validator=instance_of(datetime.date))
    some_number = attr.ib(convert=float)
    name = attr.ib(validator=instance_of(str),
                   default="Generic Name {0} - {1}%".format(
                       some_date.strftime("%d-%b-%Y"),
                       some_number * 100)
                   )

s = Something(some_date=datetime.date.today(), some_number=0.375)
```

I included the `.strftime()` conversion to highlight that `name` doesn't see a float and a date, but a `_CountingAttr` object, hence I get an AttributeError (and a TypeError for `some_number * 100`). Since I can't reference self either, what would be the correct way to do this?

hynek:
Now this is a biggie but I believe it’s worth it for 17.1.  There should be no performance penalty but it’s a feature that people have been yearning for since day one. :)

Let me know what you think, contains some boy scouting.  Most notably the completely hosted markup in api.rst around evolve.   Also WTF was with local dunder variables in methods?  I suppose copy pasta?

This enables the following:

```python
@attr.s
class C:
    x = attr.ib(default=1)
    y = attr.ib(default=attr.Factory(lambda self: self.x + 1, takes_self=True))
    z = attr.ib()

    @z.default
    def whatever(self):
        return self.y + 1
```

Now calling `C()` gives you `C(x=1, y=2, z=3)`.

Fixes #165

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -61,6 +61,12 @@ Changes:
 - Validators can now be defined conveniently inline by using the attribute as a decorator.
   Check out the `examples <http://www.attrs.org/en/stable/examples.html#validators>`_ to see it in action!
   `#143 <https://github.com/python-attrs/attrs/issues/143>`_
+- ``attr.Factory()`` now has a ``takes_self`` argument that makes the initializer to pass the partially initialized instance into the factory.
+  In other words you can define attribute defaults based on other attributes.
+  `#165`_
+- Default factories can now also be defined inline using decorators.
+  They are *always* passed the partially initialized instance.
+  `#165`_
 - Conversion can now be made optional using ``attr.converters.optional()``.
   `#105 <https://github.com/python-attrs/attrs/issues/105>`_
   `#173 <https://github.com/python-attrs/attrs/pull/173>`_
@@ -70,6 +76,7 @@ Changes:
   `#155 <https://github.com/python-attrs/attrs/pull/155>`_
 
 .. _`#136`: https://github.com/python-attrs/attrs/issues/136
+.. _`#165`: https://github.com/python-attrs/attrs/issues/165
 
 
 ----
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -100,13 +100,20 @@ Core
       >>> @attr.s
       ... class C(object):
       ...     x = attr.ib(default=attr.Factory(list))
+      ...     y = attr.ib(default=attr.Factory(
+      ...         lambda self: set(self.x),
+      ...         takes_self=True)
+      ...     )
       >>> C()
-      C(x=[])
+      C(x=[], y=set())
+      >>> C([1, 2, 3])
+      C(x=[1, 2, 3], y={1, 2, 3})
 
 
 .. autoexception:: attr.exceptions.FrozenInstanceError
 .. autoexception:: attr.exceptions.AttrsAttributeNotFoundError
 .. autoexception:: attr.exceptions.NotAnAttrsClassError
+.. autoexception:: attr.exceptions.DefaultAlreadySetError
 
 
 .. _helpers:
@@ -203,11 +210,12 @@ See :ref:`asdict` for examples.
       >>> i1 == i2
       False
 
-    ``evolve`` creates a new instance using ``__init__``. This fact has several implications:
+  ``evolve`` creates a new instance using ``__init__``.
+  This fact has several implications:
 
-    * private attributes should be specified without the leading underscore, just like in ``__init__``.
-    * attributes with ``init=False`` can't be set with ``evolve``.
-    * the usual ``__init__`` validators will validate the new values.
+  * private attributes should be specified without the leading underscore, just like in ``__init__``.
+  * attributes with ``init=False`` can't be set with ``evolve``.
+  * the usual ``__init__`` validators will validate the new values.
 
 .. autofunction:: validate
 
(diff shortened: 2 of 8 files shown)


bluetech:
This is somewhat evil, but have you considered counting the number of arguments to the factory function, instead of taking `takes_self` explicitly? Should be possible with something like `takes_self = len(inspect.signature(factory_fn).parameters) == 1`. Note: I do not endorse this :)

hynek:
Not only did I consider it, I've even implemented it. :)

But it seemed too magicy with too many weird edge cases.  It's always possible to add magic but it's impossible to remove it.  And I reckon the decorator approach will be the main one anyways...

Tinche:
OK, I'm starting looking at this. Let's talk API first :)

The `Factory` with `takes_self` is a little verbose IMO. Would you consider having another class which implies `takes_self`?

```
@attr.s
class A:
    a = attr.ib()
    b = attr.ib(default=attr.SelfFactory(lambda self: self.a + 1))
    c = attr.ib(default=attr.Factory(lambda self: self.a + 1, takes_self=True)  # for comparison
```

Another suggestion. I've been mulling over #178 with spare brain cycles. I've come to realize the wordy
```
@attr.s
class Cache(object):
    _stored = attr.ib(default=attr.Factory(list))
    _by_name = attr.ib(default=attr.Factory(dict))
    _by_id = attr.ib(default=attr.Factory(dict))
```
could be rewritten like this:
```
from attr import Factory as call

@attr.s
class Cache(object):
    _stored = attr.ib(default=call(list))
    _by_name = attr.ib(default=call(dict))
    _by_id = attr.ib(default=call(dict))
``` 
To me this reads much better and is possible today. Honestly I think I would prefer this rather than allow both ``new`` and ``default`` in ``attr.ib()``.

What if we actually implemented ``call`` on top of ``attr.Factory``, and put some magic into it? So if you want, just use ``attr.Factory`` directly, but 99% of usages would be ``call``.

```
from attr import call

@attr.s
class A:
    a = attr.ib()
    b = attr.ib(default=attr.SelfFactory(lambda self: self.a + 1))
    c = attr.ib(default=attr.Factory(lambda self: self.a + 1, takes_self=True)
    d = attr.ib(default=call(lambda self: self.a + 1))
```

hynek:
I like the idea of a smart call.  It solves two problems at once and doesn't risk any backward incompatibilities.

But that should be a separate PR, right?   Let's merge this first and implement call on top of it.  I promise there won't be 17.1 w/o call. ;)

hynek:
Hm, maybe it shouldn't be `@x.default` but `@x.factory` Tho?

Tinche:
:+1: to separate PR.

As for `@a.default` vs `@a.factory`, I think I prefer `default` since:
* it's directly analogous to `attr.ib(default=...)`
* you can't have both `attr.ib(default=...)` and `@a.default`
* there is no `attr.ib(factory=...)`. 

You can't really decorate a value so I'm OK with the fact the decorator is applied to a method and the `attr.ib` argument to a value.

I'll take a look at the implementation first thing I have the chance. You said you wanted this before PyCon, that's on Wednesday?

hynek:
Yeah I travel on Wednesday but accordingly my Tuesday is rather dense so sooner would be better. :)

Tinche:
Given the following test class:
```
import attr


@attr.s(slots=True)
class A:
    a = attr.ib(default=0)
    b = attr.ib(default=attr.Factory(list))
    c = attr.ib(default=attr.Factory(lambda self: self.b, takes_self=True))
    d = attr.ib()

    @d.default
    def _d_default(self):
        return self.a + 1
```
this init gets generated:

```
def __init__(self, a=attr_dict['a'].default, b=NOTHING, c=NOTHING, d=NOTHING):
    self.a = a
    if b is not NOTHING:
        self.b = b
    else:
        self.b = attr_dict['b'].default.factory()
    if c is not NOTHING:
        self.c = c
    else:
        self.c = attr_dict['c'].default.factory(self)
    if d is not NOTHING:
        self.d = d
    else:
        self.d = attr_dict['d'].default.factory(self)
```
which is fine, nothing wrong with it. However, we can speed it up, as always, by shifting work. This init does a dict lookup (on attr_dict), then an instance look up on that, then a method call. This is unnecessary. If we just inject the factory into the globals and generate an init like this:
```
def __init__(self, a=attr_dict['a'].default, b=NOTHING, c=NOTHING, d=NOTHING):
    self.a = a
    if b is not NOTHING:
        self.b = b
    else:
        self.b = __attr_factory_b()
    if c is not NOTHING:
        self.c = c
    else:
        self.c = __attr_factory_c(self)
    if d is not NOTHING:
        self.d = d
    else:
        self.d = __attr_factory_d(self)
```
my benchmark goes from 1.69 us +- 0.05 us to 1.38 us +- 0.07 us, which is significant (~22% speedup).
Here's a patch, apply with git apply: [changes.txt](https://github.com/python-attrs/attrs/files/999772/changes.txt)

There, now the ball is in your court :p

hynek:
Aaaand it’s back to you fine sir!

exarkun:
The renaming of `default` to `_default` in `Attribute`'s initializer broke txaws, fwiw.

exarkun:
https://github.com/python-attrs/attrs/issues/194

Issue #192, opened 2017-05-16
17.1.0 apparently breaks asdict


pauloalem:
Hello there! 
I have upgraded to 17.1.0 and aparently it breaks down on asdict convertion.
I've got this utility validator `optional_instance_of` that ends up raising a
```   
return hash(_attrs_to_tuple(self, attrs))
TypeError: unhashable type: '_OptionalValidator'
```
Here's a sample to trigger that behavior: 

```python
import attr


def optional_instance_of(of):
    return attr.validators.optional(attr.validators.instance_of(of))


@attr.s()
class A:
    a = attr.ib(default=None, validator=optional_instance_of(int))

a = A()
print(attr.asdict(a, filter=attr.filters.exclude(type(None))))
```

Is this related to #142  ?

hynek:
should be fixed by #193

PR #193, opened 2017-05-16
Compatibility fixes.

Tinche:
This just restores `__hash__` to our validators. I'd have preferred to make them frozen, but that requires a discussion and is maybe better left until we have zero-overhead frozen classes.

See #192.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- Validators are hashable again.
+  Note that validators may become frozen in the future, pending availability of no-overhead frozen classes.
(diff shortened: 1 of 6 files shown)


Tinche:
I've pushed another commit in. `_CountingAttr` can't have `.default` since it has a method `.default()`, but `Attribute` can. And technically `Attribute` is a public class, right?

exarkun:
With the 17.1.0 release, my test suite encounters several failures.  With this branch, those failures are gone.  So, from a compatibility perspective, this looks good to me.

hynek:
Could you add a regression test that loops over `*Validator` in validators.py and checks they’ve hash=True?

hynek:
Ah look at that, did a test find a missing piece? ;)

PR #201, opened 2017-06-01
Activate warnings

hynek:
pytest 3.1.0 activated all warnings and 3.1.1 deactivated them again.  I’ve found some latent bugs in my own software so I want them to be always on.  I had to add a few context managers to assoc but that’s worth it.


Tinche:
Ok to delete the branch?

hynek:
yeah i did it myself now but feel free to kill merged branches with 🔥

PR #205, opened 2017-06-06
Fix/slot hashes

Tinche:
This is ready for review I think.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -368,6 +368,8 @@ def attributes(maybe_cls=None, these=None, repr_ns=None,
                 cls = _add_pickle(cls)
         if slots is True:
             cls_dict = dict(cls.__dict__)
+            was_unhashable = ('__hash__' in cls_dict and
+                              cls_dict['__hash__'] is None)
             cls_dict["__slots__"] = tuple(ca_list)
             for ca_name in ca_list:
                 # It might not actually be in there, e.g. if using 'these'.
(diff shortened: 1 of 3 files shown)


src/attr/_make.py
```diff
@@ -368,6 +368,8 @@ def wrap(cls):
                 cls = _add_pickle(cls)
         if slots is True:
             cls_dict = dict(cls.__dict__)
+            was_unhashable = ('__hash__' in cls_dict and
```
hynek: Please add a human-readable comment what this means exactly…otherwise we’ll stumble over it every time we try to understand the code. :)

Also please `"` 🐶

src/attr/_make.py
```diff
@@ -376,6 +378,14 @@ def wrap(cls):
 
             qualname = getattr(cls, "__qualname__", None)
             cls = type(cls)(cls.__name__, cls.__bases__, cls_dict)
+
+            if (not was_unhashable and hash is False and
```
hynek: Same here.  Mother of god, this is like a riddle. :D

tests/test_hash.py
```diff
@@ -0,0 +1,105 @@
...
+import pytest
+
+import attr
+
+from hypothesis import given
+from hypothesis.strategies import booleans
+
+
+@given(slots=booleans(), frozen=booleans(), cmp=booleans())
+def test_no_subclass_no_hash(slots, frozen, cmp):
```
hynek: I was disappointed there wasn’t a `test_no_woman_no_cry()` eventually. ;)

hynek:
Oh and add the PR# to the changelog please.

tests/test_hash.py
hynek: (this is not a review request! ;))

Tinche:
I'll get around to this, just haven't had the time.

hynek:
Well, given Nathaniel's comments I don't even know what to do anymore.  :/ seems like we're digging ourselves deeper.

hynek:
So I’m closing this and we’ll go the other way: break it on non-slots too eventually 🎉  so the behavior is consistent with Python’s.

PR #209, opened 2017-06-11
Pass appropriate stacklevel in assoc deprecation warning.

### LINKED ISSUE #208: assoc warning is unhelpful
Currently the warning emitted looks something like this:
```
/home/mithrandi/deployment/virtualenvs/txaws/site-packages/attr/_funcs.py:173: DeprecationWarning: assoc is deprecated and will be removed after 2018/01.
```
I think this should pass an appropriate stacklevel so you can see where the calling code actually is.

mithrandi:
Fixes #208.

I (deliberately) didn't document this change or write a test for it, let me know if that was a mistake.

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -170,7 +170,7 @@ def assoc(inst, **changes):
     """
     import warnings
     warnings.warn("assoc is deprecated and will be removed after 2018/01.",
-                  DeprecationWarning)
+                  DeprecationWarning, stacklevel=2)
     new = copy.copy(inst)
     attrs = fields(inst.__class__)
     for k, v in iteritems(changes):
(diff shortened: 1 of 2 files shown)


hynek:
Could we test somehow that the correct frame is in the warning?

mithrandi:
We could check the source line [reported in the warning](https://docs.pytest.org/en/features/warnings.html#_pytest.recwarn.WarningsRecorder), but that seem somewhat fragile since code moves around all the time. Maybe just check the source file?

hynek:
yeah or method name or something?

hynek:
Does this look good to you?

hynek:
Ugh pytest changed something. Hang on.

PR #225, opened 2017-08-01
Remove review feedback line from CONTRIBUTING

The-Compiler:
GitHub sends out mails on new PR pushes since some time already.

If this is still there intentionally, feel free to reject this though!

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -18,8 +18,7 @@ Workflow
   This is a hard rule; patches with missing tests or documentation can't be accepted.
 - Make sure your changes pass our CI_.
   You won't get any feedback until it's green unless you ask for it.
-- Once you've addressed review feedback, make sure to bump the pull request with a short note.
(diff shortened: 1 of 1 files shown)


hynek:
Yeah I saw it just today and thought it should be fixed.  I think telling people to ping a ticket once they want a new review is a good thing.  It just shouldn’t be justified with a lie. :)

Just delete the second line and add a “so we know you’re done.”. to the first one :)

The-Compiler:
Done (how meta! 😆)

hynek:
Please no new line after a comma tho. :)

The-Compiler:
Makes sense, next try! :wink:

This is the first time I hear about semantic newlines, by the way - I think it makes a lot of sense!

hynek:
Thanks for hanging in there!

Yep, they make a lot of sense to me too.  Text diff tooling might have improved but I still love the visual representation of the text’s structure.

PR #235, opened 2017-08-18
Add more docs about __attrs_post_init__

### LINKED ISSUE #228: __attrs_post_init__ is missing from API docs
`__attrs_post_init__` does not have an entry on attrs.readthedocs.io - there's only a mention in the examples and a few asides in the API section. It would be great to have first-class treatment, including the signature (`self`), when it's called (after the object is fully initialized) and the problem that it's trying to solve.

hynek:
This should fix #228 – does this meet your expectations @akaptur?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -142,7 +142,7 @@ Generally speaking, it's best to keep logic out of your ``__init__``.
 The moment you need a finer control over how your class is instantiated, it's usually best to use a classmethod factory or to apply the `builder pattern <https://en.wikipedia.org/wiki/Builder_pattern>`_.
 
 However, sometimes you need to do that one quick thing after your class is initialized.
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -135,6 +135,55 @@ Core
...
 
 
+Influencing Initialization
+++++++++++++++++++++++++++
+
+Generally speaking, it's best to keep logic out of your ``__init__``.
+The moment you need a finer control over how your class is instantiated, it's usually best to use a classmethod factory or to apply the `builder pattern <https://en.wikipedia.org/wiki/Builder_pattern>`_.
+
+However, sometimes you need to do that one quick thing after your class is initialized.
+And for that ``attrs`` offers the ``__attrs_post_init__`` hook that is automatically detected and ran after ``attrs`` is done initializing your instance:
```
akaptur: nit: s/ran/run/

hynek:
Thanks!

PR #237, opened 2017-08-20
Expand notes on namedtuples

hynek:
I took some time to make our “why not namedtuples?” section more complete and to the point.

I very much welcome feedback and additions.

### DIFF
--- a/docs/why.rst
+++ b/docs/why.rst
@@ -75,9 +75,7 @@ The most obvious difference between ``namedtuple``\ s and ``attrs``-based classe
    >>> i1 == i2
    False
 
-
-…while a namedtuple is *explicitly* intended to behave like a tuple which means the type of a tuple is *ignored*:
-
+…while a ``namedtuple`` is *intentionally* `behaving like a tuple`_ which means the type of a tuple is *ignored*:
 
 .. doctest::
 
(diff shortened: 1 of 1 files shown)


docs/why.rst
```diff
@@ -57,81 +57,71 @@ Adding an attribute to a class concerns only those who actually care about that
 …namedtuples?
 -------------
 
-The difference between :func:`collections.namedtuple`\ s and classes decorated by ``attrs`` is that the latter are type-sensitive and require less typing as compared with regular classes:
+:func:`collections.namedtuple`\ s are tuples with names, not classes. [#history]_
+Since writing classes is tiresome in Python, now and then someone discovers all the typing they can save and get really excited.
```
markrwilliams: > now and then someone discovers all the typing they ~~can~~ **could** save **with `namedtuples`** and ~get~ **gets** really excited

docs/why.rst
```diff
@@ -57,81 +57,71 @@ Adding an attribute to a class concerns only those who actually care about that
...
    True
 
+Other often surprising behaviors include:
 
-This can easily lead to surprising and unintended behaviors.
+- ``namedtuple``\ s are iterable and indexable which is not what you'd expect from a class.
+- Iterability also implies it's easy to accidentally unpack a ``namedtuple`` which leads to hard-to-find bugs. [#iter]_
+- ``namedtuple`` adds its methods to your instances whether you want or not. [#pollution]_
+- ``namedtuple``\ s are *always* immutable.
+  Not only does that mean that you can't decide for yourself whether your instances should be immutable or not, it also means that if you want to influence your class' initialization (validation?  default values?), you have to implement :meth:`object.__new__` which is a particularly hacky and error-prone requirement for a very common problem. [#immutable]_
```
markrwilliams: Another sentence or two that explains why overriding `__new__` is error-prone and hacky would be good here.

docs/why.rst
```diff
@@ -57,81 +57,71 @@ Adding an attribute to a class concerns only those who actually care about that
...
+- To attach methods to a ``namedtuple`` you have to subclass it.
```
markrwilliams: You might want to discuss how this is particularly bad with `namedtuple`.  Consider the example from the [official documentation](https://docs.python.org/3.7/library/collections.html):

```python
class Point(namedtuple('Point', ['x', 'y'])):
    __slots__ = ()
    @property
    def hypot(self):
        return (self.x ** 2 + self.y ** 2) ** 0.5
    def __str__(self):
        return 'Point: x=%6.3f  y=%6.3f  hypot=%6.3f' % (self.x, self.y, self.hypot)
```

`Point`'s MRO now contains two classes with the same name, defined in the same module:

```
[<class 'point.Point'>, <class 'point.Point'>, <type 'tuple'>, <type 'object'>]
```

where the earlier one has been shadowed by the later one.

This looks really weird in documentation that includes class hierarchies, like that generated by [`autodoc` with `show-inheritance`](http://www.sphinx-doc.org/en/stable/ext/autodoc.html).

docs/why.rst
```diff
@@ -57,81 +57,71 @@ Adding an attribute to a class concerns only those who actually care about that
...
+  Again: common problem, hacky solution.
 
-Opinions on object immutability vary.
-With ``attrs``, the choice is yours.
-Immutable classes are created by passing a ``frozen=True`` argument to the :func:`attr.s` decorator.
-By default, however, classes created by ``attrs`` are regular Python classes and therefore mutable:
-
-.. doctest::
+All these things make ``namedtuple``\ s particular poor choice for public APIs because all your objects are irrevocably tainted.
```
markrwilliams: > All these things make `namedtuple`s **a** ~particular~ **particularly** poor choice for public APIs**,** because all your objects are irrevocably tainted.

docs/why.rst
```diff
@@ -57,81 +57,71 @@ Adding an attribute to a class concerns only those who actually care about that
...
-   Traceback (most recent call last):
-     File "<stdin>", line 1, in <module>
-   AttributeError: can't set attribute
+.. _tuple: https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences
 
-Other than that, ``attrs`` also adds nifty features like validators, converters, and default values.
+.. rubric:: Footnotes
 
-.. _tuple: https://docs.python.org/2/tutorial/datastructures.html#tuples-and-sequences
+.. [#history] The word is that ``namedtuple``\ s have been added to the Python standard library as a way to make tuples in return values more readable.
```
markrwilliams: > The word is that ``namedtuple``\ s ~~have been~~ **were** added to the Python standard library as a way to make tuples in return values more readable.

docs/why.rst
hynek: I would agree but it’s tough to express in a concise manner.  Anyone aware of a blog post or something I could link? :|

Issue #215, opened 2017-06-26
[RFC] First class type support


hynek:
This is kind of part of #214 but it stands for itself too.

Something like:

```python
@attr.s(check_types=True)
class C:
    x = attr.ib(type=int)
```

ought to work.

Regardless of checking the type, the type information has to be accessible somewhere so we can have first-class deserialization support.

yunstanford:
+1

toumorokoshi:
This would be awesome! Is the goal to enable type checking within attrs itself? and to support types as listed in typing.py? There's a few caveats that we've discovered, diving in.

1 is there's no standardized type checking behavior for collections in typing. From my understanding, this is still an area that needs more work:

https://github.com/python/typing/issues/136

Most people have been writing their own type validation libraries to compensate:

https://pypi.python.org/pypi/typeguard

All in all, I'm a huge fan of this direction. Just wondering what the choice is for type checker.

hynek:
We don’t have concrete plans *yet*!  Mostly because we want to make types & serialization first class concepts in attrs while doing this. :)

hynek:
Wait, wrong ticket! :) Yeah, no still no idea yet!

toumorokoshi:
I have a couple thoughts on this, if you're looking for ideas.

The type parameters should probably be equivalent to adding two behavior: validate and convert:

```
new_style = attr.ib(type=int)
old_style = attr.ib(validator=validators.instance_of(int), converter=int)
```
This works pretty well for primitive types. There's some nastiness with more complex types, which attrs probably shouldn't solve. For example, I'd expect to be able to do something like:

```
from datetime import daytime

@attr.s
class X:
    date = attr.ib(type=datetime)

# epoch
x = X(date=1092183940129)
# ISO 8601 string
x = X(date="2017-06-30T14:14:57Z")
```
but these would fail, because the datetime constructor takes none of these.

I would argue probably just not handling this case, and letting people override with whatever complex converters they prefer:

```
attr.ib(type=datetime, converter=my_complex_converter)
```

The other part is using or supporting typing modules. this will require some choices in behavior (e.g. validating types recursively, or just validating the top level collection for collection types). I prefer recursive validation and conversion, personally.

hynek:
Yes, feel  free to dump all your thoughts.  The more input we get beforehand, the less mistakes we’re likely to make. :)

Tinche:
OK, let's split the proposal into sub-features for easier commenting.

### Create the type metadata

[...]

PR #238, opened 2017-08-27
Add PEP484 stubs

chadrik:
Here's a first take at PEP484-compatible pyi stubs.  The discussion in #215 has been mostly focused on runtime type checking, so I wanted to make sure that the plan discussed there will work with static analysis as well.

Here's a quick test of the stubs that I put together to run against mypy:

```python
from typing import cast
import attr

@attr.s
class C(object):
    x : int = attr.ib(convert=int)
    y: str = attr.ib()

c = C('1', 'foo')  # E: Too many arguments for "C"
c.x = 2
c.y = 'bar'
c > C()  # E: Unsupported left operand type for > ("C")
str(c)

# errors below are correctly detected:
c.x = 'foo'  # E: Incompatible types in assignment (expression has type "str", variable has type "int")
c.y = 1  # E: Incompatible types in assignment (expression has type "int", variable has type "str")

D = attr.make_class('D', {'x': cast(attr._CountingAttr, attr.ib())})
d = D(2)
# errors below are not correctly detected (mypy knows nothing about D other than that it's a type)
d.x ='foo'
```

mypy knows nothing about the dunder methods added by attrs, so cmp methods and `__init__` fail.  This will have to be solved with a mypy plugin, which I'll look into next. Same goes for `attr.make_class` though that's likely to be more difficult.

One important thing to note: in order to ensure that `y: str = attr.ib()` does not trigger a mypy error, the return type of `attr.ib` is `Any` instead of `_CountingAttr`.   A side effect of this is that it complicates passing the result of `attr.ib` to functions that expect `_CountingAttr` instances.  There are a handful of strategies to solve this:

1. use `typing.cast`:  ugly and verbose
2. create, or choose an existing, alias for `attr.ib` that will return `_CountingAttr`, to use with functions like `make_class`. 
3. add another argument (`make=True`?) whose presence invokes an `@overload` of `attr.ib` that returns `_CountingAttr`
4. slight variation on 2.  create a public alias for `_CountingAttr` and copy the keyword defaults from `attr.ib` to `_CountingAttr.__init__`.  In other words, if you use static type checking and are working with a function that expects `_CountingAttr` instances, use `_CountingAttr()`  directly instead of `attr.ib()`

~~I favor the 3rd solution.~~  After some consideration I favor 4.

If you're interested in accepting this, the next things to figure out are:
- how to make tests for this, and run them on travis
- how best to install the stubs (should they be moved to [python-typeshed](https://github.com/python/typeshed)?)

Let me know what you think.

[...]

PR #239, opened 2017-08-28
Basic type support

chadrik:
Added `Attribute.type`, which gets its value from attribute `__annotations__` in python 3.6+ or a newly added `type` argument to `attr.ib()`. 

This is a prerequisite for runtime and static type-checking, as discussed in #215.

### DIFF
--- a/changelog.d/239.change.rst
+++ b/changelog.d/239.change.rst
@@ -1,3 +1,3 @@
-Added ``type`` argument to ``attr.attr()`` and corresponding ``type`` attribute to ``attr.Attribute``.
-This value can be inspected by third-party tools for type checking and serialization.
+Added ``type`` argument to ``attr.ib()`` and corresponding ``type`` attribute to ``attr.Attribute``.
+This change paves the way for automatic type checking and serialization (though as of this release attrs does not make use of it).
 In Python 3.6 or higher, the value of ``attr.Attribute.type`` can also be set using variable type annotations (see `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_).
--- a/changelog.d/239.change.rst
+++ b/changelog.d/239.change.rst
@@ -1,3 +1,3 @@
 Added ``type`` argument to ``attr.ib()`` and corresponding ``type`` attribute to ``attr.Attribute``.
-This change paves the way for automatic type checking and serialization (though as of this release attrs does not make use of it).
-In Python 3.6 or higher, the value of ``attr.Attribute.type`` can also be set using variable type annotations (see `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_).
+This change paves the way for automatic type checking and serialization (though as of this release ``attrs`` does not make use of it).
+In Python 3.6 or higher, the value of ``attr.Attribute.type`` can alternately be set using variable type annotations (see `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_).
--- a/changelog.d/239.change.rst
+++ b/changelog.d/239.change.rst
@@ -1,0 +1,3 @@
+Added ``type`` argument to ``attr.attr()`` and corresponding ``type`` attribute to ``attr.Attribute``.
+This value can be inspected by third-party tools for type checking and serialization.
+In Python 3.6 or higher, the value of ``attr.Attribute.type`` can also be set using variable type annotations (see `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_).
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -127,7 +127,8 @@ def attr(default=NOTHING, validator=None,
         components.  See :ref:`extending_metadata`.
     :param type: The type of the attribute.  In Python 3.6 or greater, the
         preferred method to specify the type is using a variable annotation
-        (see PEP-526 ).  This argument is provided for backward compatibility.
+        (see `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_).  
+        This argument is provided for backward compatibility.
         Regardless of the approach used, the type will be stored on
         ``Attribute.type``.
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -62,7 +62,7 @@ Sentinel to indicate the lack of a value when ``None`` is ambiguous.
 
 def attr(default=NOTHING, validator=None,
          repr=True, cmp=True, hash=None, init=True,
-         convert=None, metadata={}):
+         convert=None, metadata={}, type=None):
     """
     Create a new attribute on a class.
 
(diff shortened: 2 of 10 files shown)


src/attr/_make.py
```diff
@@ -125,10 +125,16 @@ def attr(default=NOTHING, validator=None,
         value is converted before being passed to the validator, if any.
     :param metadata: An arbitrary mapping, to be used by third-party
         components.  See :ref:`extending_metadata`.
+    :param type: The type of the attribute.  In python 3.6 or greater, the
```
hynek: **P**ython please :)

src/attr/_make.py
```diff
@@ -191,8 +198,11 @@ def _transform_attrs(cls, these):
                    for name, ca
                    in iteritems(these)]
 
+    ann = getattr(cls, '__annotations__', {})
```
hynek: double quotes please

src/attr/_make.py
```diff
@@ -871,22 +882,31 @@ def __init__(self, name, default, validator, repr, cmp, hash, init,
...
         raise FrozenInstanceError()
 
     @classmethod
-    def from_counting_attr(cls, name, ca):
+    def from_counting_attr(cls, name, ca, type=None):
+        # type holds the annotated value. deal with conflicts:
+        if type is None:
+            type = ca.type
+        elif ca.type is not None and type is not ca.type:
+            raise ValueError(
```
hynek: I think we should raise an error regardless whether they differ?  A tighter API envelope is usually better.

tests/test_make.py
```diff
@@ -406,6 +406,19 @@ def __attrs_post_init__(self2):
...
+    def test_types(self):
+        """
+        Sets the `Attribute.type` attr from type argument.
+        """
+        @attributes
+        class C(object):
+            x = attr(type=int)
+            y = attr(type=str)
+            z = attr()
+        assert int is C.__attrs_attrs__[0].type
```
hynek: I think it's better to use `attr.fields(C).x` etc here

tox.ini
```diff
@@ -18,7 +18,7 @@ commands = coverage run --parallel -m pytest {posargs}
 
 [testenv:py36]
 deps = -rdev-requirements.txt
-commands = coverage run --parallel -m pytest {posargs}
+commands = coverage run --parallel -m pytest {posargs} tests/_test_annotations.py
```
hynek: The `_test_` approach is not ideal because it relies on using tox to run the tests.

Please use the approach I took in https://github.com/hynek/prometheus_async/blob/d82f527c905cd581411007ccdbdaca396f0bad77/conftest.py#L23-L24 instead.

tests/_test_annotations.py
```diff
@@ -0,0 +1,67 @@
+"""
+Tests for python 3 type annotations.
```
hynek: well, it's not python 3, it's Python 3.6…let's call it PEP 526 instead?

src/attr/_make.py
```diff
@@ -125,10 +125,16 @@ def attr(default=NOTHING, validator=None,
         value is converted before being passed to the validator, if any.
     :param metadata: An arbitrary mapping, to be used by third-party
         components.  See :ref:`extending_metadata`.
+    :param type: The type of the attribute.  In Python 3.6 or greater, the
+        preferred method to specify the type is using a variable annotation
+        (see PEP-526 ).  This argument is provided for backward compatibility.
```
hynek: PEPs are written without dashes and maybe this should point to some docs?  in any case there shouldn't be a blank before the paren :)

chadrik:
All notes addressed (as of 11 minutes ago), and added the changelog fragment.

changelog.d/239.change.rst
```diff
@@ -0,0 +1,3 @@
+Added ``type`` argument to ``attr.attr()`` and corresponding ``type`` attribute to ``attr.Attribute``.
```
hynek: `attr.ib()` :)  it can be admittedly confusing when coming from the actual source.  I’ve been thinking to refactor the implicit names for a while…

changelog.d/239.change.rst
```diff
@@ -0,0 +1,3 @@
...
+This value can be inspected by third-party tools for type checking and serialization.
```
hynek: I think this should be more general – we’ll probably use it ourselves too, right?  Maybe just scratch ` by third-party tools` altogether.

OTOH we should stress that nothing happens just by setting it.

src/attr/_make.py
```diff
@@ -871,22 +883,31 @@ def __init__(self, name, default, validator, repr, cmp, hash, init,
...
 
     @classmethod
-    def from_counting_attr(cls, name, ca):
+    def from_counting_attr(cls, name, ca, type=None):
+        # type holds the annotated value. deal with conflicts:
+        if type is None:
+            type = ca.type
+        elif ca.type is not None:
+            raise ValueError(
+                "Type annotation and type argument are both present: "
```
hynek: No need to add the type names into the message anymore.

Tinche:
@hynek Roger, I'll put this into my queue.

changelog.d/239.change.rst
chadrik: How about this:

This change paves the way for automatic type checking and serialization features (though as of this release there are no built-in features that make use of it).

src/attr/_make.py
chadrik: I thought it could be useful in more complex cases where it's not immediately obvious from the class definition how the type argument was set.  But I'm happy to drop it.

changelog.d/239.change.rst
chadrik: In changelogs should the project name be monospace?  attrs or `attrs`?

src/attr/_make.py
```diff
@@ -125,10 +125,17 @@ def attr(default=NOTHING, validator=None,
...
+        preferred method to specify the type is using a variable annotation
+        (see `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_).
+        This argument is provided for backward compatibility.
+        Regardless of the approach used, the type will be stored on
+        ``Attribute.type``.
 
     ..  versionchanged:: 17.1.0 *validator* can be a ``list`` now.
     ..  versionchanged:: 17.1.0
         *hash* is ``None`` and therefore mirrors *cmp* by default .
+    ..  versionadded:: 17.3.0 *type*
```
chadrik: @Tinche What version will this be released in?

changelog.d/239.change.rst
hynek: Always monospaced `attrs` everywhere when you talk about the project.

Well let's not overthink it, we don't know whether or not we'll make use of it but I can still edit the fragment before the release.

src/attr/_make.py
Tinche: I would say the next one, so 17.3.0 is right.

Tinche:
It's still showing @hynek as requesting some changes, but I can't figure out exactly what. I think all comments have been addressed?

wsanchez:
@Tinche: Github is kinda lame about review process workflow, and this looks a lot like one of the ways it's lame.

This is great, BTW.

Tinche:
In you go!

Tinche:
Thank you for your contribution.

chadrik:
It's been a pleasure.  You all run a very tight ship here.  I like it.

Head up: I'll be rebasing #238 against this change soon, and pushing that PR along.

Issue #174, opened 2017-03-30
slots=True classes are not weak-referenceable


gmacon:
This code:

```
import weakref
import attr

@attr.s(slots=True)
class Foo(object):
    x = attr.ib()

weakref.ref(Foo('x'))
```

produces this error:

```
Traceback (most recent call last):
  File "slotsref.py", line 8, in <module>
    weakref.ref(Foo('x'))
TypeError: cannot create weak reference to 'Foo' object
```

In the docs for *\_\_slots\_\_*, it says "Without a *\_\_weakref\_\_* variable for each instance, classes defining *\_\_slots\_\_* do not support weak references to its instances. If weak reference support is needed, then add `'__weakref__'` to the sequence of strings in the *\_\_slots\_\_* declaration."

Adding `'__weakref__'` to the list isn't hard, but I wanted to see if people had any opinions about whether this should happen transparently, or if you should have to opt-in to it by saying `@attr.s(weakref=True)`.

Tinche:
It should be opt-in, just like it's opt-in for normal slot classes. No sense in wasting a slot by default for something that's pretty niche. I agree we should support it somehow.

Not a fan of `attr.s(weakref=True)`. The parameter doesn't make sense for dict classes and it makes our main user-facing API more complex for, again, arguably a niche feature.

You can work around the limitation today, by defining `__weakref__` yourself, explicitly.
```
import weakref
import attr


@attr.s(slots=True)
class A:
    a = attr.ib()
    __weakref__ = attr.ib(init=False, hash=False, repr=False, cmp=False)

weakref.ref(A(1))
```
This is ugly but it works fine. It's also kind of logical I think, because this is exactly the `attrs` equivalent of adding `__weakref__` to slots directly like you'd do if you weren't using attrs. The first step could be just documenting this in the docs for slot classes.

We could make it a little prettier with a wrapper:
```
import weakref
import attr


@attr.s(slots=True)
class A:
    a = attr.ib()
    __weakref__ = attr.weakref_attrib()

weakref.ref(A(1))
```

Another thing we could do is enable users to define their own `__slots__`, and just append to that, instead of generating the entire `__slots__` ourselves. Then you could do:
```
import weakref
import attr


@attr.s(slots=True)
class A:
    __slots__ = ('__weakref__')
    a = attr.ib()

weakref.ref(A(1))
```

This is completely in the spirit of `attrs`, I think. You define the class just the way you want it, and `attrs` adds to that.

[...]

PR #253, opened 2017-09-28
Stop setting attributes on class bodies

hynek:
This behavior has been deprecated since 16.1 and can now be removed in
accordance with our backward-compatibility policy.

Interestingly, this simplified a lot of our code and might even improve performance of class creation.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -505,22 +505,6 @@ Slot classes are a little different than ordinary, dictionary-backed classes:
          ...
      AttributeError: 'Coordinates' object has no attribute 'z'
 
-- Slot classes cannot share attribute names with their instances, while non-slot classes can.
-  The following behaves differently if slot classes are used:
-
-  .. doctest::
-
-    >>> @attr.s
-    ... class C(object):
-    ...     x = attr.ib()
-    >>> C.x
-    Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, convert=None, metadata=mappingproxy({}), type=None)
-    >>> @attr.s(slots=True)
-    ... class C(object):
-    ...     x = attr.ib()
-    >>> C.x
-    <member 'x' of 'C' objects>
-
 - Since non-slot classes cannot be turned into slot classes after they have been created, ``attr.s(.., slots=True)`` will *replace* the class it is applied to with a copy.
   In almost all cases this isn't a problem, but we mention it for the sake of completeness.
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -10,7 +10,6 @@ from ._compat import (
     PY2,
     iteritems,
     isclass,
-    iterkeys,
     metadata_proxy,
     set_closure_cell,
 )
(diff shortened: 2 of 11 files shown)


docs/examples.rst
```diff
@@ -505,22 +505,6 @@ Slot classes are a little different than ordinary, dictionary-backed classes:
          ...
      AttributeError: 'Coordinates' object has no attribute 'z'
 
-- Slot classes cannot share attribute names with their instances, while non-slot classes can.
```
hynek: I wasn't sure what to replace this with.

src/attr/_make.py
```diff
@@ -1068,7 +1064,14 @@ def make_class(name, attrs, bases=(object,), **attributes_arguments):
...
         raise TypeError("attrs argument must be a dict or a list.")
 
-    return attributes(**attributes_arguments)(type(name, bases, cls_dict))
+    post_init = cls_dict.pop("__attrs_post_init__", None)
+    return attributes(
+        these=cls_dict, **attributes_arguments
+    )(type(
+        name,
+        bases,
+        {} if post_init is None else {"__attrs_post_init__": post_init}
```
hynek: This is needed *at least* for our hypothesis strategy…

Tinche:
Yas! I will look at this over the weekend.

njsmith:
Huh, I haven't checked but it's possible this PR will also fix the PyPy incompatibility described here: https://github.com/python-attrs/attrs/issues/174#issuecomment-333335667

src/attr/_make.py
```diff
@@ -184,28 +183,33 @@ class MyClassAttributes(tuple):
 
 def _transform_attrs(cls, these):
     """
-    Transforms all `_CountingAttr`s on a class into `Attribute`s and saves the
-    list in `__attrs_attrs__`.
+    Transform all `_CountingAttr`s on a class into `Attribute`s and save the
+    list in `__attrs_attrs__` while potentially deleting them from *cls*.
```
Tinche: Do we need the 'potentially' there?

src/attr/_make.py
Tinche: Hm, this is technically a new feature?

So we should either document it, or if we don't want it, I can change our Hypothesis strategies.

Tinche:
I plopped this branch into my work code base, all tests still pass! :D

src/attr/_make.py
hynek: I think yes because in some cases they’re not there. :)

src/attr/_make.py
hynek: Good Question!  It seems useful?

hynek:
I think it would be a nice touch to get the ball rolling on slots-style creation and I’d like to have it in the same release as this but it’s a different PR.

src/attr/_make.py
Tinche: I think feature parity is good even though using it like this is kinda awkward. Just document that it should be a callable that takes one argument, self.

hynek: I think we should make it a proper argument to make_class, agreed?  (I’m gonna land this as it is but we should make it nicer so I'm not documenting it)

PR #257, opened 2017-10-02
Use new serious biz names in internal code

hynek:
This is a purely internal change.

The old internal names (that were also the old serious biz names) were kind of non-sensical so I’m moving the APIs to our newer ones that mirror the official names.

This PR doesn’t change any public APIs but makes sure that we have more consistent naming throughout the project.

This is something that has been annoying me for quite a while. *shiver*

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,12 +9,12 @@ def C():
     """
     Return a simple but fully featured attrs class with an x and a y attribute.
     """
-    from attr import attrs, attrib
+    import attr
 
-    @attrs
+    @attr.s
     class C(object):
-        x = attrib()
-        y = attrib()
+        x = attr.ib()
+        y = attr.ib()
 
     return C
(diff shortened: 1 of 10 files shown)


Tinche:
Hm why not refactor to "attr.s" and "attr.ib" where possible? Those are the "defaults" right?

hynek:
It can't be `*attr*.s` because then we'd have to import from the package.  would you be comfortable to use `@s` and `x = ib()`?  I think *I* could live with it.

Tinche:
Good point. I had a feeling there was a technical hurdle but was a little lazy to think about it.

I think things outside of the main source tree (i.e. tests - conftest.py) should use "attr.s", but inside "attrs" is better to me than "s".

hynek:
Done and ready for your discerning eyes. :)

Issue #202, opened 2017-06-02
Add a more idiomatic way to define hashing


hynek:
The current way of using True/False/None is confusing even to me.

I propose we use an Enum (probably just class on legacy Python) a la:

```python
class Hashing(Enum):
    SMART = "smart"  # None → new default
    BY_ID = "by_id" # False
    WRITE_FOR_ALL = "write_for_all" # True
    WRITE_FOR_NONE = "write_for_none" # True, but the default on attr.ibs is False
```

The last two names are terrible of course.

***

I probably could be talked into depending on https://pypi.org/project/enum34/ on legacy Python.

Opinions?

P.S. Now while typing this out I realized that this could solve a long-standing complaint both I and many users have:  assuming you want an attrs-generated method based on a small subset of attr.ibs, you have to write a lot of `hash|init|repr=False`.  Ditching bools for enums would allow for that. 🤔   But let’s talk about that in a different ticket.

Tinche:
We should definitely do something to make the hashing behavior clearer.

I'm currently working on a PR to "fix" slots hashing behavior, and ``attr.s(hash=False)`` is really throwing me off.

The docs say nothing about what should happen when ``hash=False``.

My personal understanding would be for attrs to do exactly nothing, we're saying do not touch ``__hash__``. However this is complicated. I will demonstrate.

Assume a very simple class, with hashing disabled:

```
@attr.s(hash=False)
class A:
    a = attr.ib()
```
The equivalent of this class is: https://gist.github.com/Tinche/d8c1dadfaa3aa4618b736a61e752b3e1

Try pasting them both. The attrs one will be hashable, due to falling back to `__object__.__hash__`. The normal one will not be.

This is because in the normal class, the interpreter will see the `__cmp__` and insert `__hash__ = None`. In the attrs class, the `__cmp__` method isn't there when the class is defined; we put it into the class after the fact. So the interpreter won't touch `__hash__` and it will fall through to `object`.

This is the root cause of why our slots behavior is different.

hynek:
Yeah so that’s settled.

Ideas for better names for `WRITE_FOR_ALL/NONE`?  @glyph? @njsmith?

P.S.  the replacement/slots thingie explains a lot of my own confusion.

Tinche:
What functionalities are we aiming for exactly?

I'm guessing:
* make hashable by generating a hash function based on attributes
* make hashable, but attributes are opt-in instead of opt-out
* make hashable by using identity-based hashing
[...]

PR #260, opened 2017-10-06
Stop modifying classes and always create new ones

hynek:
This allows us to harmonize class creation for dict and slots classes.

Interestingly this seems to break our Hypothesis strategies.  I can't dig any further right now but I wanted to put it out there.  Helpful comments welcome.

Ref #223 #202


Tinche:
I’m happy to take a look

hynek:
A few more notes:

1. I'm not married to the naming.
2. I’m not raising a warning because it seems unfortunate to force people into setting the argument and makes it kind of impossible to remove it later.  My hope is that most people don't ever realize that we changed something underneath.
3. I hate Python 2.

Tinche:
Hm after a 5 minute investigation, there seems to be a problem with `copy.copy` on classes that are `replace_class=True, frozen=True` (and only those, slot classes are OK). Will look into it more later.

On a side note, is it possible to overwrite the class `__repr__`? As in, not the instance repr. Our Hypothesis classes are called `HypClass`, would be nicer if their repr had the flags too. I mean, I could always just change the name...

hynek:
🤦‍♂️

changelog.d/260.deprecation.rst
```diff
@@ -0,0 +1,8 @@
+Added new option to ``@attr.s()``: *replace_class* will make ``attrs`` stop attaching methods to the existing class.
+Instead it will create a new class with the methods attached.
+This is consistent with how ``slots``-classes have *always* been created.
+
+For the next year, the default value is ``replace_classes=False`` which is the current behavior.
+
+The default value will change to ``True`` no sooner than one year after this release.
+There are no concrete plans to remove ``replace_classes=False`` altogether but its use is discouraged.
```
Tinche: After a year, if no problems present themselves, I say we remove the option altogether. Mainly in the interest of simplicity, `attr.s` seems to be getting options all the time.

hynek: The problem is that we’d break everyone’s software who decided to do the right thing and set it to true.

hynek: I guess what we could do is to add a `**kw` and spit out DeprecationWarnings on certain arguments that have been removed.  But I really don’t want `attrs` to become notorious for breaking backward-comparability all the time.  We’re already getting comments like https://www.reddit.com/r/Python/comments/74fro5/attrs_iii_frozen_classes/dnzjn6d/ :|

Tinche: Why don't we just do this: default to False for one year, default to True for one year, remove if no problems. I mean, the alternative is to carry this forever :/

I'm ok either way btw, this is definitely not a hill I want to die on :) I'm just worried about attr.s having a gazillion args down the line, which will make it harder to use.

hynek: Yeah well I’m deeply unhappy with every option, hence trying to source opinions. :)  The two years plan is something I considered too.  Let's see we can get some opinions out of glyph and Nate.

hynek:
So um @glyph and @njsmith, any ideas/opinions on how to make this thing happen without pissing people off unnecessarily?

hynek:
Just to add another thought: my wish is that *ideally*, nobody notices a difference at all.  So a three stage approach with a `*kw`-warning in two years seems best:

1. now: add option, default is False
2. Oct 2018: default is True
3. Oct 2019: option becomes a nop and starts raising a warning

Now, this kind of costs us a pointless year because people are not gonna try out that option unless they have a problem.  It's like with pre-releases, nobody tests them. 

Ergo, in Oct 2018, there will be people yelling at us in any case and we've maybe saved <0.01% of our users some headaches.

***

OTOH, I think I’ll do it anyway because I'm totally not ready to break hashing a second time this year. :|  Also we'll be able to point to this: yes we did the right thing.

I really hope this is the last fundamental screwup in `attrs`. :|

***

I’d really like more input on this…happily also from @wsanchez @Lukasa @DRMacIver @wbolster @hawkowl

hynek:
I have added the current plan to the news fragment.

DRMacIver:
I don't think I actually understand the problem well enough to comment yet. What actually is the interaction with hashing and equality here that causes the problem?

I assume the "super" problem is that if I have something like:

```
@attr.s()
class Foo(Bar):
    def stuff(self):
        return super(Foo, self).stuff()
```

Then the "Foo" in the method is different from the Foo outside the method and this causes a supertype mismatch of some sort? I assume the class you're creating *can't* be a subclass of the existing Foo. Are you copying over the namespace and is that what causes you problems?

DRMacIver:
OK. I've now done the reading of the news fragment and understand some of this better.

On the hashing thing... I think I'd personally be inclined to be less nice than you. I would argue that getting a hash when passing `hash=False` is just a bug and people relying on it are relying on a bug, and that just fixing that bug would be a non-breaking change. As you say, people are going to suffer that change at some point anyway, so I think you might as well just rip off the band-aid.

Personally the idea of adding a `replace_class` argument to `attr.s` doesn't sit well with me, especially when defaulting it to false. One of the rules we have in Hypothesis code review is that when adding a setting you're supposed to ask whether the user is ever likely to actually have a reason to care about the value of this setting. When adding a new scary change the user *might* care about being able to back it out, but they're certainly not going to care to opt into it because `replace_class` is about as internal to attrs as it gets.

Were it me I would adopt the following plan:

* No optional behaviour. Just opt everyone into this straight up.
* Think real hard about possible gotchas and how to fix them.
* Before releasing, try running the test suite of a few of the larger open source users of attrs and see if anything hilarious happens.
* Be on the ball for a certain amount of user screaming.
* In the unlikely event that there are any actual unfixable bugs caused by this, add a `replace_class` option defaulting to True that emits a warning when you set it to False.

Note however this is coming from the guy who thought "Lets get real drunk and turn coverage on for every Hypothesis user" was a good idea, so I may have more of a YOLO attitude to feature evolution than you are comfortable with. 😉

wsanchez:
I tend to agree with @DRMacIver, though it's possible that all the wine I had over the weekend is influencing me into associating with the "Lets get real drunk and…" camp.

I certainly agree regarding the hashing behavior; that's a bug and it should just get fixed.  I don't see a strong argument for bending over to provide compatibility there; at a minimum, I'd have `replace_class` default to `True`, but as a user I'd rather fix the bug if this effects me than opt into buggy behavior.  (But, to be fair, I also don't expect his to effect me.)

hynek:
I’d like to stress one thing: `hash=False` is supposed to give you hash-by-id, if**f**, `cmp=False` on Python 3.  It will *always* give you one on Python 2.

What we *could* do is using this as a unification/simplification for creating classes and attach `object.__hash__` for non-slot classes to not breaking any compatibility.

***

Then continue with #202 and use an enum to control what actually happens and deprecate using bool for the hash argument. 🤔

DRMacIver:
@hynek that sounds sensible to me.

hynek:
OK, I’ve spent yesterday basically rewriting the class generation code because the old one was all about tacking on and then copying or removing.

Now it's much cleaner.  There might be some doc inconsistencies I guess?  I’m much happier with how it works now.

changelog.d/223.change.rst
```diff
@@ -0,0 +1,4 @@
+``attrs``-decoratorated classes are never modified now.
+Instead a new class is created with the appropriated methods and attributes attached from the get go.
+
+For backward-compatibility, the behavior of getting hashing by ID on Python 3 when passing ``@attr.s(hash=False)`` is manually retained by setting ``__hash__`` to ``object.__hash__``.
```
wsanchez: Something is grammatically off here… maybe:

For backward-compatibility, the prior behavior of hashing by ID on Python 3 when passing ``@attr.s(hash=False)`` may be retained retained by setting ``__hash__`` to ``object.__hash__``.

changelog.d/260.change.rst
```diff
@@ -0,0 +1,4 @@
+``attrs``-decoratorated classes are never modified now.
+Instead a new class is created with the appropriated methods and attributes attached from the get go.
+
+For backward-compatibility, the behavior of getting hashing by ID on Python 3 when passing ``@attr.s(hash=False)`` is manually retained by setting ``__hash__`` to ``object.__hash__``.
```
wsanchez: See above.  Also, I'm confused about why there are two of these, but I bet there's a reason.

DRMacIver:
> What we could do is using this as a unification/simplification for creating classes and attach object.__hash__ for non-slot classes to not breaking any compatibility.

Sorry, I somehow misparsed what you originally said which was totally unambiguous but not quite what I thought the good idea was.

What I would actually recommend here is that the default behaviour is to attach a `__hash__` which calls `object.__hash__` and emits a deprecation warning saying that if this is the behaviour you want then you should be explicit about wanting it.

hynek:
> What I would actually recommend here is that the default behaviour is to attach a __hash__ which calls object.__hash__ and returns its value, but also emits a deprecation warning saying that if this is the behaviour you want then you should be explicit about wanting it.

The problem is that as of now, there is no explicit way of defining it.  You totally can ask for unhashable but with cmp and the behavior also differs between Python versions.  That’s why I want to start raising a warning as soon as there is actually a good way.

changelog.d/260.change.rst
hynek: Yeah there is: identical news fragments are merged into one changelog entry by town crier with multiple links to the issues.

changelog.d/223.change.rst
hynek: Oh dear, so my wording was so bad that it confused you what I mean. 🙈

*We* do set `__hash__` to `object.__hash__` automatically, to not break compat.

English is hard man.  Let me try again…

> The previous (incorrect) behavior of hashing by object ID on Python 3 when passing `@attr.s(hash=False)` (which implies `cmp=True`), is retained  for backward compatibility reasons.
>
> A better -- and more explicit -- way to control object hashing is in the pipeline.

Better/clearer?

wsanchez: Ah, yes, OK.  :-)

hynek:
OK folks, I get it: this PR is fucking scary.

So I’m not gonna ask you to review and risk to share the blame.  Mut what I *am* gonna ask you is to run it against your code bases and report back any breakage you encounter.  Technically, nothing should break.

I just did that and the result is 6dd4b1d.  I’ll take all the blame, but please help me not have an emergency 17.4 by running your test suite once. :D

src/attr/_make.py
```diff
@@ -181,22 +181,24 @@ class MyClassAttributes(tuple):
...
-    Return a list of tuples of (attribute name, attribute).
+    Return an `_Attributes`.
     """
     if these is None:
         ca_list = [(name, attr)
                    for name, attr
                    in cls.__dict__.items()
                    if isinstance(attr, _CountingAttr)]
-        for name, _ in ca_list:
-            delattr(cls, name)
```
DRMacIver: To check my understanding: This was only needed because the old class was still hanging around, and now it's not because they're just not copied to the newly created one anyway class?

DRMacIver:
FWIW I've run the Hypothesis test suite with these changes and it seems fine, but I also don't think Hypothesis makes interesting enough use of attrs that that's a meaningful test of anything.

Tinche:
I got one:
```
import attr
from typing import TypeVar, Generic

G = TypeVar('G', bound=int)


@attr.s
class A(Generic[G]):
    a = attr.ib()
```

Works with PyPI, not with this branch (on 3.6). GL :D

src/attr/_make.py
```diff
@@ -263,6 +272,161 @@ def _frozen_delattrs(self, name):
...
+        # compiler will bake a reference to the class in the method itself
+        # as `method.__closure__`.  Since we replace the class with a
+        # clone, we rewrite these references so it keeps working.
+        for item in cls.__dict__.values():
+            closure_cells = getattr(item, "__closure__", None)
+            if not closure_cells:  # Catch None or the empty list.
+                continue
+            for cell in closure_cells:
+                if cell.cell_contents is self._cls:
+                    set_closure_cell(cell, cls)
```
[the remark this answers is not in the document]
hynek: It is, isn't it?  I believe @Tinche sold his soul for it. :D

src/attr/_make.py
```diff
@@ -206,35 +208,42 @@ def _transform_attrs(cls, these):
...
     for c in reversed(cls.__mro__[1:-1]):
         sub_attrs = getattr(c, "__attrs_attrs__", None)
         if sub_attrs is not None:
-            super_cls.extend(
-                a for a in sub_attrs
-                if a not in super_cls and a.name not in non_super_names
+            super_attrs.extend(
+                a
+                for a in sub_attrs
+                if a not in super_attrs and a.name not in non_super_names
```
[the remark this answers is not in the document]
hynek: Well, we need to add each only once and we can’t use a set, because the default might not be hashable. :|

Suggestions/recommendations?

src/attr/_make.py
hynek: Yeah, the original approach was to stick everything on the old class and if it’s a slots class, copy the result to a new one.

hynek:
Thanks @DRMacIver, your look is very much appreciated!

@Tinche not sure if trolling? :)

src/attr/_make.py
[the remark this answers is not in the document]
DRMacIver: Well ignoring the O(n^2) thing, I think an explicit loop would be *much* clearer here:

```python
for a in sub_attrs:
    if a not in super_attrs and a.name not in non_super_names:
        super_attrs.append(a)
```

That way you at least don't have to deal with the question of how iteration and updating the list interact (which is the bit that confused me).

In terms of suggesting how to improve this complexity wise, I don't think it matters much (there aren't *that* many attrs and it only runs once), but I don't think I understand the code well enough. Why are there duplicates in `sub_attrs` at all?

Tinche:
I'm not trolling, this was a failure from my work codebase. I've simplified it of course.

hynek:
Related to the type work, not this PR right?  Maybe open new issue and set milestone?

wsanchez:
Well so I was wrong… this did break my code, specifically, [this test](https://github.com/burningmantech/ranger-ims-server/blob/master/src/ims/model/test/test_cmp.py#L49).

This isn't my favorite code, so please don't judge me.  :-)

I have a hack mix-in class I use for attrs-classes where I want some simple control over how comparisons are done… anyway it fails [here](https://github.com/burningmantech/ranger-ims-server/blob/master/src/ims/model/test/test_cmp.py#L55).

wsanchez:
(I'm 100% OK with changing this code, by the way, just reporting on test results, as requested.)

src/attr/_make.py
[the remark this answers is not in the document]
hynek: > Why are there duplicates in sub_attrs at all?

I think something something inheritance?  I think I’ll do the loop thing and investigate, thanks!  Your “naive” questions are *very* helpful.

src/attr/_make.py
```diff
@@ -347,75 +511,39 @@ def attrs(maybe_cls=None, these=None, repr_ns=None,
...
+            builder.add_cmp()
 
         if hash is not True and hash is not False and hash is not None:
+            # Can't use `hash in` because 1 == True for example.
             raise TypeError(
                 "Invalid value for hash.  Must be True, False, or None."
             )
         elif hash is False or (hash is None and cmp is False):
-            pass
+            builder.add_hash_by_id()
```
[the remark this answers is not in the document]
hynek: No, because the whole point of this patch is that dict classes will behave like slots classes now.  A comment should be added tho indeed!  (your question genuinely confused me and made me think it through one more time, so thanks again for that!)

src/attr/_make.py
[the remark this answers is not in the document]
hynek: Turns out™, you’ve spotted a bug!  Check out 507d9d5baae3871ff0392cf7794165b45c1e6fac for details if you want to know more.

Does it make more sense now?

hynek:
Great news everyone, we’ve got ourselves another edge case:

```python
import attr

@attr.s
class BaseNoSlots:
  def __init_subclass__(cls, param, **kwargs):
    super().__init_subclass__(**kwargs)
    cls.param = param

@attr.s    
class Foo(BaseNoSlots, param='foo'):
  pass

print(Foo.param)
```

This is taken from #269 and https://repl.it/Mmkb/0

Since we didn’t intervene into the class generation for non-slots before, it just worked.

Now we get:

```pycon
>>> import attr
...
... @attr.s
... class BaseNoSlots:
...   def __init_subclass__(cls, param, **kwargs):
...     super().__init_subclass__(**kwargs)
...     cls.param = param
...
... @attr.s
... class Foo(BaseNoSlots, param='foo'):
...   pass
  File "<stdin>", line 10, in <module>
  File "/Users/hynek/Projects/attrs/src/attr/_make.py", line 573, in attrs
    return wrap(maybe_cls)
  File "/Users/hynek/Projects/attrs/src/attr/_make.py", line 566, in wrap
    return builder.build()
  File "/Users/hynek/Projects/attrs/src/attr/_make.py", line 379, in build
    cd,
TypeError: __init_subclass__() missing 1 required positional argument: 'param'
__init_subclass__() missing 1 required positional argument: 'param'
```

hynek:
@RonnyPfannschmidt sooo…how could metaclasses help here without making stuff runtime-dynamic?

hynek:
In any case: I’m going to make this PR only about the refactoring which is a net win.  Then we’ll get attrs 17.3 out of the door and then we‘ll tackle it again.  Maybe we’ll find a better way to attach __slots__.

PR #272, opened 2017-10-19
Refactor class creation

### LINKED ISSUE #269: slots=True do not play nice with __init_subclass__

hynek:
This is a sad remnant of #260 but at least it makes the code much nicer.

Fixes #269


hynek:
To be clear: this should be 100% backward compatible; pls report any breakage.

wsanchez:
Tests pass for me.

PR #274, opened 2017-10-21
Refactor and polish docs

hynek:
`attrs` has grown out of its doc structure long time ago and it’s time to acknowledge that we need narrative docs.  Docs that maybe will prevent people from opening the daily issue about underscore prefix-handling in `__init__` and docs that give us better space to explain concepts in general.

This isn’t those docs yet.  Instead it makes the index more beginner friendly and lays out the space for narrative docs: the `Day-to-Day Usage` section (better names welcome).

It’s against #272 because I hope that’s gonna get merged soon (*wink* *wink*).

To give you an idea, how the new layout looks like, I’ve activated it in readthedocs for your viewing pleasure: http://www.attrs.org/en/narrative-docs/

Please let me know what ya’ll think!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -230,7 +230,7 @@ Changes:
 ^^^^^^^^
 
 - ``__slots__`` have arrived!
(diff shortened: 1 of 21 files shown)


hynek:
(maybe it’s also time to get rid of the testimonials, at least in the docs.  they take *a lot* of space)

PR #277, opened 2017-10-27
Add auto_attribs

### LINKED ISSUE #262: Add option to collect annotated fields
Since I’m sick of hearing that “[PEP 557](https://www.python.org/dev/peps/pep-0557/) is like attrs, but using variable annotations for field declarations”, I’d like to have an option to collect annotated fields that have no attr.ib definition.

ie.

```python
@attr.s(collect_bare=True)
class C:
    x: int
```

should be equivalent to:

```python
@attr.s
class C:
    x: int = attr.ib()
```

I’m open to better/shorter names.

Volunteers?  @chadrik maybe? :)

hynek:
Well, here we go.  I’ve also added examples for the *type* argument.

Fixes #262 

Final step before 17.3.0.


hynek:
Also I wonder if we should troll a bit and add a `dataclass = attr.s(auto_attribs=True)` alias. ;)

hynek:
People who might be interested to give some feedback on the API because they expressed interest in this stuff before: @agronholm @wsanchez @ambv @DRMacIver @chadrik 

I’m not as afraid to ship a bug as locking us into a conceptually wrong direction.

wsanchez:
The [auto_attrib example](https://github.com/python-attrs/attrs/pull/277/files#diff-266b5b92397656d625732c1aa3db0b85R483) looks great to me.

docs/examples.rst
```diff
@@ -461,6 +461,51 @@ The metadata dictionary follows the normal dictionary rules: keys need to be has
...
+.. doctest::
+
+   >>> import typing
+   >>> @attr.s(auto_attribs=True)
+   ... class AutoC:
+   ...     x: int
+   ...     y: int
+   ...     foo: typing.Any = attr.ib(
+   ...          default="every attrib needs a type if auto_attribs=True"
+   ...     )
```
wsanchez: I'd consider making this a `str`, since that's the type of the default, and absent some reason not given here, it just looks like a bad practice to use `Any`.

wsanchez:
The `dataclass` trolling alias is kind of a great idea but it they really want to be compatible, so [this example from PEP 557](https://www.python.org/dev/peps/pep-0557/#abstract) should work, for example:

```python
@attr.s(auto_attribs=True)
class InventoryItem:
    quantity_on_hand: int = 0
```

agronholm:
Is my hunch correct that it does not properly handle class variable annotations (`ClassVar[....]`)? Those must be ignored for this to work properly.

hynek:
> Is my hunch correct that it does not properly handle class variable annotations (ClassVar[....])? Those must be ignored for this to work properly.

Could you give a full example of what wouldn’t work please?

hynek:
OMG @wsanchez thanks for reminding me!  Implemented in 0ea9b00635cdac543d4e1bfdf86099b7f9db3af1 and it looks epic!

This works now as expected:

```python
@attr.s(auto_attribs=True)
class C:
    a: int
    x: typing.List[int] = attr.Factory(list)
    y: int = 2
    z: int = attr.ib(default=3)
    foo: typing.Any = None
```

agronholm:
> Could you give a full example of what wouldn’t work please?

```python
@attr.s(auto_attribs=True)
class Foo:
    var: ClassVar[int]
    foo: str
    bar: bool
```
Here `foo` and `bar` should be recognized as attrib()utes but not `var`.

hynek:
@agronholm hm that’s interesting in combination with #220.

Doesn’t a ClassVar have to have a value too?  We could treat `var: ClassVar[int] = 42` like `attr.ib(init=False)`, or am I missing something?

agronholm:
> Doesn’t a ClassVar have to have a value too?

Why? You could always set it later. 

> We could treat var: ClassVar[int] = 42 like attr.ib(init=False), or am I missing something?

No, because when combining with `slots=True`, it would add a `__slots__` entry for it, which is wrong.

hynek:
> Why? You could always set it later.

Hm yeah I guess.  It’s probably not too common (except for subclassing) but it’s possible.

> No, because when combining with slots=True, it would add a __slots__ entry for it, which is wrong.

Well yeah, black magic around slots is implied.

***

The big question now is if we can merge this without blocking ourselves from handling #220 nicely and the answer seems to be “no”.

So what would be a good way to support this?

- allow a way to mark attributes as class variables (either `x: ClassVar[int]` or `attr.ib(class_var=True)`)
- make those always `init=False`
- for slots:
   - set the attribute to the value on the new class
   - don't add them to `__slots__`

Missing something?

***

As for next steps, if we can agree on what I’ve just outlined, I’d take the discussion to #220, promote it to blocking for 17.3 and get *this* merged.

It’s frustrating that 17.3.0 moved into a far future again but I really don’t want to paint myself into a corner with this because I really care about #220.

agronholm:
What is the benefit of including class variables in the attrs system in the first place?

hynek:
In non-slots you can overwrite class variables and they become instance variables, see #220 for a discussion.  It kind of gives you “free” defaults.

agronholm:
I've moved the discussion to the other thread.

hynek:
From the docs:

> ClassVar is not a class itself, and should not be used with isinstance() or issubclass(). ClassVar does not change Python runtime behavior, but it can be used by third-party type checkers.

Ehhh…so what is the clean way here?  I just went for basically `isinstance(a, ClassVar.__class__)` which appears to work.

agronholm:
Another question: how will you deal with `foo: int = 4` and `slots=True`?

agronholm:
Also, have you considered that the user might be using `backports.typing`?

hynek:
For now it’s treated as a default as per documentation:

```pycon
>>> i = C()

>>> i.foo
4

>>> i.foo = 5

>>> i.foo
5
```

Are you insinuating we should make them class variables?  I feel like it’s more useful this way and if you want a class var, you use a `ClassVar`.  (it also copies  PEP 557 semantics)

hynek:
> Also, have you considered that the user might be using `backports.typing`?

Since we only annotations on Python 3.6, I see no value in adding that complexity.

agronholm:
> Are you insinuating we should make them class variables? I feel like it’s more useful this way and if you want a class var, you use a ClassVar. (it also copies PEP 557 semantics)

I'm just trying to make sure that attrs removes them as class variables so they don't conflict with `__slots__`.

hynek:
When you say “removes them as class variables” you mean that they don’t exist on the class body, correct?

And yes, they don’t, because we have to create slots classes by hand anyways.  I’ve added regression tests to ensure that contract.

agronholm:
Ok, good, then it seems my concerns have been addressed.

hynek:
> Ok, good, then it seems my concerns have been addressed.

Thank you *very* much for your feedback, you’ve saved me a bunch of embarrassment.

chadrik:
I have a question about this:

```python
x: typing.List[int] = attr.Factory(list)
```

Technically with static type checking this is an error since we're assigning a `Factory` instance to a `List[int]` variable. 

Over in #238 I'm working on writing PEP484 stubs for attrs.   I'm already using the stubs to lie to the type system, e.g. we make `x: int = attr.attrib(default=0)` possible even though `attr.attrib` returns a `_CountingAttr` instance.  

I  can do the same for `Factory`, with a stub function that lies about the return type of `Factory`.  Instead of a class, it becomes a function that pretends to return the type of the callable given to it.

```python
def Factory(Union[Callable[[], _T], Callable[[Any], _T]], takes_self: bool = ...) -> _T: ...
```

Now if you ran the following through mypy, it would tell you that the type of `x` is `list`.

```python
x = attr.Factory(list)
reveal_type(x)
# Revealed type is 'builtins.list
```

The main drawback of this from a technical standpoint is that accessing instance variables of `Factory`, such as `x.factory` or `x.takes_self` will raise an error in mypy, since mypy will think that `x` is a `list`.

Lying about the return type of `attr.ib` is a necessary evil if we want type checking to work at all, but this lie feels a bit less justified.  When it comes to factories I think I'd rather be slightly more verbose, but more correct:

```python
@attr.s(auto_attribs=True)
class C:
    x: typing.List[int] = attr.ib(attr.Factory(list))
```

Thoughts?

agronholm:
`x: typing.List[int] = attr.Factory(list)` is correct. The point of the type hinting system is that the developer knows better than the static typechecker what the type will be, so the type information is treated as gospel by the checker. Wrapping it with `attr.ib()` seems pointless to me not only because that, but also because it won't give the typechecker any better information either.

hynek:
The good news is that everyone can decide that for themselves if they want wrap or not.  We just set whatever is assigned as a default.

Tinche:
OK, I went through the PR. I don't have any complaints on the implementation.

A few thoughts on the API follow.

Non-annotated fields with attr.ib() are silently ignored. I know this is documented, but it still might trip users up. I know the technical reason we can't support them. It should be an error - it's almost certainly not what a user would want and we can detect it with certainty. Just print out what to do and maybe the reason we can't support it (so people don't whine at the library "but you knew what I meant" :)

Do we need to document there's no runtime validation going on? Users might assume there is?

@chadrik brings up a good point with regards to static typing. We're just going to have to lie through our teeth in our stubs, but we already do that.

chadrik:
> The good news is that everyone can decide that for themselves if they want wrap or not.

Static type checking is not that flexible: we have to choose one type as the result of `attr.Factory()`. If we want to allow the brief form then the following becomes an error in static type checkers like mypy, PyCharm, or SublimeJEDI:

```python
factory = attr.Factory(list)
print("pass_self is %s" % factory.pass_self)  # error: "list" has no attribute "takes_self" 

@attr.s(auto_attribs=True)
class C:
    x: typing.List[int] = factory
```

I don't have a very strong opinion either way.  I favor correctness where possible, but acknowledge that the desire to access attributes of `Factory` is a niche case.  

Before making a decision on this, I think it's important to first consider that most users are exposed  to static type checking via their IDE, such as PyCharm or sublime with [jedi](https://github.com/srusskih/SublimeJEDI). That means that most users of static type checking have not *knowingly* adopted it, and may not even know that the suggestions and warnings within their IDE are the result of PEP484 and pyi stubs.  Our goal should be to make their experience as rewarding as possible, which in my mind means supporting, via attr's pyi stubs, the idioms put forth in the attrs documentation.

That in turn means that this project should begin taking static type checking into account when designing features.  If a feature is released that cannot be accurately described in the [attrs pyi stubs](https://github.com/python-attrs/attrs/pull/238), then you may be introducing undue frustration and confusion to users of PyCharm, sublime, etc.

With all that said, if everyone feels that it is a fair trade-off to lie to type checkers about the result of `attr.Factory` in order to allow `x: List[int] = attr.Factory(list)`, then I wiIll make the necessary adjustments to the stubs.

hynek:
@Tinche I’ve added errors on unannotated fields in 71f3ddf

***

And I really don’t know what to say about the static typing thing.  What I meant with “everyone can decide” is that the fact it works with bare Factories is kind of coincidental.

I’m not sure but did someone suggest something actionable about this? :)  I have to admit that this whole topic is rather gibberish to me and I’d prefer if someone else made the call on it.

What are our options here?

chadrik:
@hynek  there are 3 options:
1. encourage the long form `x: List[int] = attr.ib(Factory(list))` by presenting this in the docs, and optionally making the short form raise an error.
2. make the stubs to support `x: List[int] = Factory(list)`, and accept the cons presented above.
3. encourage the short form via the docs for average users, while supporting the long form in the stubs: users of static type checking can use the long form, since they've opted in to a more restrictive toolset.

- 1 is the most correct, but it's less convenient for users.
- 2 seems to be where people are leaning, but first I'd like someone connected to the project evaluate the cons that I presented above.  @Tinche is probably in the best position to make this call.
- I'm opposed to 3, because the adoption of PEP484 by IDEs means that static type checking is no longer opt-in: almost everyone is exposed to it (or will be soon), so the stubs should support the idioms used in the docs.

FWIW, `dataclasses` will not support anything like `x: List[int] = Factory(list)` because they do not have a `Factory` object.  They will require `x: List[int] = field(default_factory=list)`.  That could easily be used as an argument for *and against* the short form.

wsanchez:
I do think that `x: List[int] = Factory(list)` is sufficiently nice syntactic sugar to justify lying to the type checker…  but that may be because I think that the fact that my classes use `attrs` is an implementation detail, so that `x` is a `Factory` isn't something I want documented outside of the class implementation, where I'm OK with having the cast `x` to `Factory` in the few times I'd need to use `Factory` attributes.

ambv:
I like this.

hynek:
Ugh looks like we need to rethink the handling of ClassVar: https://github.com/ericvsmith/dataclasses/issues/61

hynek:
As for the factory drama: I think I’ll go with option 1 for now unless @Tinche can shed some different light on this.

***

That said, is there *some* way to make it work with a different function that works somehow differently?  `x: List[int] = factory(list)` just looks too cute to be passed on. :|

Tinche:
Ugh. I say let's go with @chadrik 's option two and allow it. "Useful now" trumps "maybe easier for mypy" later.

hynek:
Allowing was never the question, since we just apply the default…it works by accident. :)  The question is what we put into the docs. :|

***

But I really need to do something clever about the typing.ClassVar thingie for the reason alone, that typing imports are slow. *Sigh*

src/attr/_make.py
```diff
@@ -191,16 +191,14 @@ class MyClassAttributes(tuple):
...
-        pass
+def _is_class_var(annot):
+    """
+    Check whether *annot* is a typing.ClassVar.
 
-    _CLASS_VAR_CLS = FakeClassVar
+    The implementation is gross but importing `typing` is slow and there are
+    discussions to remove it from the stdlib alltogether.
+    """
+    return str(annot).startswith("typing.ClassVar")
```
Tinche: Naughty 😙

chadrik:
> That said, is there some way to make it work with a different function that works somehow differently?

There is a fourth option, which I did not see previously because I was blinded by my own bias.  We can stub `Factory` so that it returns `Any`, which disables type checking for this object completely and would therefore allow it to be used in the short form or the long form without static type check errors.  The downside is that you are no longer warned about *actual* errors, such as:

```python
x: typing.List[int] = attr.Factory(str)
```

I prefer option 2 over this. With option 2 there is a low chance of false errors, but with option 4 there is a high chance of undetected errors.

wsanchez:
@chadrik: I agree.

The false errors in option 2 are rare and easily worked around with a cast.

wsanchez:
Actually, the more I think about it, the more Option 2 seems correct to me.  `x` is a `List[int]` in the intended API for class `C`.  That `Factory` is involved at all is only because I'm using `attrs`; that is: `Factory` is an artifact of `attrs`'s API, not of `C`'s, so `C`'s annotations ideally shouldn't leak that out.

No?

agronholm:
> Actually, the more I think about it, the more Option 2 seems correct to me. x is a List[int] in the intended API for class C. That Factory is involved at all is only because I'm using attrs; that is: Factory is an artifact of attrs's API, not of C's, so C's annotations ideally shouldn't leak that out.

I just tried, and PyCharm does not make the type of `x` the union of `List[int]` and `Factory`. MyPy does. But if `Factory()` was marked as returning `object`, it would not pollute the type information, right?

chadrik:
>  I just tried, and PyCharm does not make the type of x the union of List[int] and Factory. MyPy does. 

I'm not entirely sure what you mean.  Without any special attrs pyi stubs, this is an error in mypy:

```python
x: typing.List[int] = attr.Factory(list)
```

Based on PEP484, when you annotate a variable with a type, the value assigned to it must be "compatible" with that type.   Nothing here becomes the union of anything else.  A `Factory` is not a `list`, so it's an error.  I'm not sure what PyCharm does here, but it still has a lot of catching up to do wrt PEP484.

> But if Factory() was marked as returning object, it would not pollute the type information, right?

Still not sure what you  mean.   If you want the above code to work without an error in mypy, `Factory` must return a value that is compatible with `List[int]`.  `Any` is compatible with everything (it's the union of all possibilities), so as I said earlier, that is one option, but not one that I like.  `object` is the intersection of all possibilities: i.e. the lowest common denominator.  Only operations that are possible on all types are valid with `object`.  It's not what you want here.

agronholm:
> Still not sure what you mean. If you want the above code to work without an error in mypy, Factory must return a value that is compatible with List[int]. 

Could you show me a simple script with such an annotation that would result in a mypy error on inspection?

chadrik:
> Could you show me a simple script with such an annotation that would result in a mypy error on inspection?

```python
import typing
import attr
x: typing.List[int] = attr.Factory(list)
```

Make sure that you set `MYPYPATH` to the <attrs_repo>/src directory (do not set it to your site-packages directory).

agronholm:
Wait, are you saying that I need to have a checked out copy of the attrs git repo whenever I run mypy against my code??

hynek:
That or get the stubs into typeshed IIUC.  It’s kind of bonkers.

chadrik:
> That or get the stubs into typeshed IIUC. It’s kind of bonkers.

Yeah, it's very annoying, but there's [a pep](https://github.com/ethanhs/peps/blob/typedist/pep-0561.rst) in progress to solve it at the distutils/setuptools level.

> Alright, no wonder I didn't get any errors before then.

Well, you would have gotten an error like ` Cannot find module named 'attr'`

agronholm:
> Well, you would have gotten an error like Cannot find module named 'attr'

I routinely skip missing imports with MyPy because most libraries don't have stubs.

hynek:
Since everyone seems to love option 2...do *I* have to do something a out it or am I done?

chadrik:
> Since everyone seems to love option 2...do I have to do something a out it or am I done?

I handled it over on #238

chadrik:
> Since everyone seems to love option 2...do I have to do something a out it or am I done?

I handled it over on #238

PR #282, opened 2017-11-04
Fix pickling for instances of classes created using make_class()

trendels:
Currently, trying to pickle an instance of a class created using `attrs.make_class` fails (tested on Python 2.7 and 3.5):

    >>> import attr
    >>> import pickle
    >>> C = attr.make_class('C', ['x'])
    >>> obj = C(1)
    >>> pickle.dumps(obj)
    _pickle.PicklingError: Can't pickle <class 'attr._make.C'>: attribute lookup C on attr._make failed

It has something to do with pickle looking for the class/type in the wrong module. I was curious how `collections.namedtuple` solves this and found this (also present in Python 2.7):

https://github.com/python/cpython/blob/v3.5.4/Lib/collections/__init__.py#L436-L445

This PR adds the same code (with comment) verbatim to attrs, to make pickling work. As a side effect, the `repr` of classes created this way now also contain name of the module where `make_class()` was called instead of "attr._make". I'm not sure if this needs a Changelog/Documentation update?

I also noticed that Python 3.6 added an explicit `module` parameter to `collections.namedtuple`, to allow manually overriding the module name to fix issues with pickling on other Python implementations such as IronPython (https://bugs.python.org/issue17941). Maybe this would also be useful to add to `attrs` in the future.

### DIFF
--- a/changelog.d/282.change.rst
+++ b/changelog.d/282.change.rst
@@ -1,0 +1,1 @@
+Instances of classes created using ``attr.make_class()`` can now be pickled.
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -601,7 +601,7 @@ You can still have power over the attributes if you pass a dictionary of name: `
    ...                     repr=False)
    >>> i = C()
    >>> i  # no repr added!
-   <attr._make.C object at ...>
+   <__main__.C object at ...>
    >>> i.x
    42
    >>> i.y
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1248,7 +1248,8 @@ def make_class(name, attrs, bases=(object,), **attributes_arguments):
     # sys._getframe is not defined (Jython for example) or sys._getframe is not
     # defined for arguments greater than 0 (IronPython)
     try:
-        type_.__module__ = sys._getframe(1).f_globals.get('__name__', '__main__')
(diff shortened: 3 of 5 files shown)


hynek:
That looks like a good last-minute change for 17.3 (famous last words). 

Would you mind completing coverage and add a changelog entry?  There’s a bunch of cosmetic problems but I can deal with them myself once it’s merged.

Thanks!

trendels:
That would be awesome.

I've added a changelog entry. I'm not sure what do do about the coverage though, other than using `# pragma: no cover`. [The line that is missing coverage](https://github.com/python-attrs/attrs/pull/282/commits/86ed904e5e94b1435b42cbaa5b28a9c77bcf88ae#diff-21756223332b04173cec8910fda6655bR1252) is only there to silence exceptions on more "esoteric" Python interpreters (IronPython, Jython) where `sys._getframes` is not available or works differently than in CPython, according to the comment from `collections.namedtuple`.

hynek:
Maybe try `monkeypatch.delattr(sys, "_getframe)"`?

If that doesn’t work because it’s too invasive, just add a no cover and call it a Saturday.

trendels:
Ah yes, that should work. TIL about pytest monkeypatch :)

hynek:
Wonderful, so you got something out of it too. ;)  Thank you for your contribution!

PR #294, opened 2017-11-15
@validate and @default decorators return self instead of meth, fixes #264

### LINKED ISSUE #264: Name of default factory method does matter
Name of default factory method does matter. It shouldn't be the same as the name of attribute:

```python
>>> @attr.s
... class C(object):
...     x = attr.ib(default=1)
...     y = attr.ib()
...     @y.default
...     def y(self):
...         return self.x + 1
>>> C()
C(x=1)
>>> C().y
<bound method C.y of C(x=1)>
```

malinoff:


changelog.d/294.change.rst
```diff
@@ -0,0 +1 @@
+``@attr.ib().default`` and ``@attr.ib().validate`` decorators now return ``self`` instead of ``meth``
```
hynek: It’s validator, not validate.  We’ll fix convert eventually to achieve consistency to avoid this exact confusion.

hynek:
Heh so I played with it and I don’t think that returning self is gonna work because in the moment you set a different name to an `attr.ib()`, you define a new attribute:

```
File "examples.rst", line 280, in default
Failed example:
    C()
Expected:
    C(x=1, y=2)
Got:
    C(x=1, y=2, name_does_not_matter=2)
```

hynek:
So I’m gonna close this because I think it can’t be made to work. :(

If someone knows a way to make this work, please speak out.

PR #297, opened 2017-11-26
Use isort so we can stop thinking about imports

hynek:
We add two tox environments for it: one that just checks and one that formats.
The checker runs as part of Travis, the formatter runs locally if you run `tox`
without arguments.

It gives us a bit of churn on the imports but I’ve been meaning to do this for a while.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,8 @@
+[settings]
+atomic=true
+lines_after_imports=2
+lines_between_types=1
+multi_line_output=5
+not_skip=__init__.py
+
+known_first_party=attr
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,8 +1,0 @@
-[settings]
-atomic=true
-lines_after_imports=2
-lines_between_types=1
-multi_line_output=5
-not_skip=__init__.py
-
-known_first_party=attr
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,6 @@ matrix:
     - python: "3.6"
       env: TOXENV=flake8
     - python: "3.6"
-      env: TOXENV=isort-check
-    - python: "3.6"
       env: TOXENV=manifest
     - python: "3.6"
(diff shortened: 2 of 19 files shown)


.isort.cfg
```diff
@@ -0,0 +1,8 @@
+[settings]
+atomic=true
+lines_after_imports=2
+lines_between_types=1
+multi_line_output=5
+not_skip=__init__.py
+
+known_first_party=attr
```
blueyed: Could also go into `setup.cfg`.

.travis.yml
```diff
@@ -25,6 +25,8 @@ matrix:
     # Meta
     - python: "3.6"
       env: TOXENV=flake8
+    - python: "3.6"
+      env: TOXENV=isort-check
```
blueyed: Is it worth a single job?
There could be one `checkqa` job.
And since there is flake8 already, what about using https://github.com/gforcada/flake8-isort ?

tox.ini
```diff
@@ -21,12 +21,42 @@ deps = -rdev-requirements.txt
...
+    coverage report
+
+
 [testenv:flake8]
 basepython = python3.6
 deps = flake8
 commands = flake8 src tests setup.py conftest.py docs/conf.py
 
 
+# Runs when invoking tox without arguments and formats the files.
```
blueyed: A bit unexpected, isn't it?
Also, doesn't `tox` run all of `envlist` then?

.isort.cfg
hynek: Hm I used to have issues with it but it seems to work now!

.travis.yml
hynek: Excellent point, fixed.

tox.ini
hynek: 1. Maybe?  But in my own practice, I found it very useful.
2. It does, but the now removed `isort-check` was not part of the `envlist` – hence the distinction.

Issue #285, opened 2017-11-09
17.3.0 changes the order how inherited class attributes are defined if they have the same base class


Emantor:
Minimal working example:
```
import attr

@attr.s()
class A:
    name = attr.ib()
    cookie = attr.ib(default="Cookie")

@attr.s()
class B:
    cake = attr.ib(default="Cake")
    cereal = attr.ib(default="Cereal")

@attr.s()
class C(B, A):
    breakfast = attr.ib(default="All")

@attr.s()
class D(A):
    food = attr.ib(default="food")
    melon = attr.ib(default="melon")

@attr.s()
class E(D, C):
    apple = attr.ib(default="apple")
```
This works with attrs==17.2.0 and breaks with attrs==17.3.0. In 17.3.0 the list of non_super_names in base.py: `['cake', 'cereal', 'breakfast', 'name', 'cookie', 'food', 'melon', 'apple']` breaks because attr is trying to create a non-default attribute in between default attributes.

jluebbe:
When looking at what _transform_attrs is doing for class E, we see what's going wrong. The MRO for E is E, D, C, B, A.
It first collects all attributes from D.__attrs_attrs__ in reverse order: melon, food, cookie, name. Then it looks at C.__attrs_attrs__ and collects the attributes which are not yet defined: breakfast, cereal, cake.
When reversing this again, the name attribute ends up in the middle.
It seems that this should work if it only looked at the locally defined attributes when iterating over the MRO, as A.name would be processed last.

jluebbe:
As a minimal workaround, this change fixes the order for our usecase:
``` diff
@@ -278,7 +278,9 @@
             # list in the end and get all attributes in the order they have
             # been defined.
             for a in reversed(sub_attrs):
-                if a.name not in taken_attr_names:
+                if a.name in super_cls.__super_names__:
+                    continue
+                elif a.name not in taken_attr_names:
                     super_attrs.append(a)
                     taken_attr_names.add(a.name)
 
@@ -354,6 +356,7 @@
         self._has_post_init = bool(getattr(cls, "__attrs_post_init__", False))
 
         self._cls_dict["__attrs_attrs__"] = self._attrs
+        self._cls_dict["__super_names__"] = self._super_names
 
         if frozen:
             self._cls_dict["__setattr__"] = _frozen_setattrs
```
The downside is obviously cluttering the generated classes with the ``__super_names__`` attribute. Perhaps each Attribute could keep a reference to the class where it was defined.
[...]

PR #299, opened 2017-11-27
BF: reverse MRO as well if we are reversing the attrs at the end

### LINKED ISSUE #298: order of attributes is 'backwards' in case of multiple inheritance
consider following code.  I also added a dummy method `m` to show that precedence is "correct"ly follows for a basic method

```python

@attributes
class A(object):
    a = attr()
    def m(self):
        print("A.m")

@attributes
class B(object):
    b = attr()
    def m(self):
        print("B.m")

@attributes
class C(A, B):
    c = attr()

    def m(self):
        super(C, self).m()

    @classmethod
    def supers(cls):
        return cls.__bases__

c = C(1, 2, 3)

print(c)
c.m()
print("supers=%s" % str(c.supers()))
```
running it results in 
```
C(b=1, a=2, c=3)
A.m
supers=(<class '__main__.A'>, <class '__main__.B'>)
```
so the first argument is (IMHO) incorrectly assigned to attribute b not a, and then 2nd one to a...

Before recent 3040bdabbc627fc6a4d151a9cb2e6f9d4177b6f2  quick and dirty (and possibly incomplete) fix was to remove `reversed` around mro within src/attr/_make.py  but neither I have looked yet into how logic should be adjusted in the current state of affairs, nor I analyzed for why reversed was needed to start with

yarikoptic:
Changes traversal of MRO to be done without any reverse-twice dance -- not really sure why it is needed since attributes simply "sink down" into derived classes quite naturally, or tests do not capture the intricacies enough.  

- [x] It does resolve incorrect (IMHO) ordering of attributes in case of multiple inheritance (Closes #298) 
       This changed behavior from what was discussed/fixed in #285 .  To me, the order enforced by the test `test_multiple_inheritance` did not make much sense, so I might be utterly wrong!
- [x] Added **tests** for changed code and adjusted (as stated above) one previous test

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -279,13 +279,12 @@ def _transform_attrs(cls, these, auto_attribs):
     # redefinitions deeper in the hierarchy.
     super_attrs = []
     taken_attr_names = {a.name: a for a in non_super_attrs}
-    for super_cls in reversed(cls.__mro__[1:-1]):
+    for super_cls in cls.__mro__[1:]:
         sub_attrs = getattr(super_cls, "__attrs_attrs__", None)
         if sub_attrs is not None:
-            # We iterate over sub_attrs backwards so we can reverse the whole
-            # list in the end and get all attributes in the order they have
-            # been defined.
-            for a in reversed(sub_attrs):
+            # We iterate over sub_attrs forward and populate them as soon
+            # as we see it
+            for a in sub_attrs:
                 prev_a = taken_attr_names.get(a.name)
                 if prev_a is None:
                     super_attrs.append(a)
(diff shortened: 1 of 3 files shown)


src/attr/_make.py
```diff
@@ -279,27 +279,25 @@ def _transform_attrs(cls, these, auto_attribs):
     # redefinitions deeper in the hierarchy.
     super_attrs = []
     taken_attr_names = {a.name: a for a in non_super_attrs}
-    for super_cls in cls.__mro__[1:-1]:
+    for super_cls in cls.__mro__[1:]:
```
yarikoptic: comment about this change: what if the last one isn't object, i.e. why to impose assumptions without validation?  If `object`, it would be happily skipped since wouldn't have any `__attrs_attrs__`

tests/test_make.py
```diff
@@ -240,8 +240,8 @@ class E(D, C):
             e2 = attr.ib(default="e2")
 
         assert (
-            "E(a1='a1', a2='a2', b1='b1', b2='b2', c1='c1', c2='c2', d1='d1', "
-            "d2='d2', e1='e1', e2='e2')"
+            "E(a1='a1', a2='a2', d1='d1', d2='d2', b1='b1', b2='b2', "
+            "c1='c1', c2='c2', e1='e1', e2='e2')"
```
yarikoptic: Why was d args following a args???  E is first subclassing from D, which sub-classes A.  So it is only logical and sensible to have A and D args first!  or what am I missing?

yarikoptic: ah, may be because there is another route to A through C, so expectation was that they would somehow "squeeze" (overload) previously defined ones out?  but then still then `d`s should then lead the list of args.

yarikoptic:
Please provide guidance where/if my logic is flawed

hynek:
Fixed by #304

PR #304, opened 2017-12-03
Fix MRO traversal with multiple inheritance

### LINKED ISSUE #298: order of attributes is 'backwards' in case of multiple inheritance

hynek:
So the double reversal we used to do screwed up the MRO traversal for forever (i.e. this is not a 17.3 regression, I had to fix examples.rst!).

This fixes it – the question is how we go about it.  A proper deprecation cycle seems impossible again, so I guess we can push it and ask for forgiveness?

This fixes #298 and #299 cc @yarikoptic

### DIFF
--- a/changelog.d/298.breaking.rst
+++ b/changelog.d/298.breaking.rst
@@ -1,0 +1,4 @@
+The traversal of of MROs when using multiple inheritance was backward:
+If you defined a class ``C`` that subclasses ``A`` and ``B`` like ``C(A, B)``, ``attrs`` would have collected the attributes from ``B`` *before* those of ``A``.
+
+This is fixed now however due to its nature, it may require fixes on your side too unfortunately.
--- a/changelog.d/299.breaking.rst
+++ b/changelog.d/299.breaking.rst
@@ -1,0 +1,4 @@
+The traversal of of MROs when using multiple inheritance was backward:
+If you defined a class ``C`` that subclasses ``A`` and ``B`` like ``C(A, B)``, ``attrs`` would have collected the attributes from ``B`` *before* those of ``A``.
+
+This is fixed now however due to its nature, it may require fixes on your side too unfortunately.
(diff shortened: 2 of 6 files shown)


yarikoptic:
Well, +1 from me ;-)

hynek:
Paging @glyph for wise opinions. :|

glyph:
Somewhat belatedly (although it appears in time for a release?) I did have an idea about this.

The deprecation cycle here *could* go like this:

- X.Y emit a warning when calling multiply-inherited `__init__` with positional arguments
- X.Y+1 raise an error when calling multiply-inherited `__init__` with positional arguments
- X.Y+2 add back this positional ordering.

Alternately, if we just want to bite the bullet, we could switch to this behavior, but still emit a warning for a release or two about using positional args in the multiple-inheritance case.

This might help folks with deep inheritance trees at least *notice* that there might be a problem if their test coverage isn't great.

Calling `__init__` with kwargs is generally clearer anyway, *especially* in the presence of potentially confusing orderings as with e.g. diamond inheritance.  So if this pushed pretty much 100% of code using `attrs` to always use kwargs maybe that would be a good idea.

hynek:
Hm that seems a rather harsh thing to do, but an interesting idea.

I’m not ready to go that far, but I think I’ll add a blurb to the newsfragments, encouraging people to use kwargs.

Issue #305, opened 2017-12-03
Equality by comparing tuples is not always correct


hynek:
Another Hynek-didn’t-quite-fully-understand-Python bug: while I maintain that establishing order using tuples makes totally sense, Hypothesis just taught me about an invariant in container equality I wasn’t fully aware of:

```python
>>> from math import nan

>>> nan == nan
False

>>> (nan,) == (nan,)
True
```

The problem here is that tuples compare by identity first and by value second.  And nan is identical but unequal.

I’ve noticed this when I tried to implement our `__eq__` using a string of `and` s instead of building two tuples which I’m sure is faster.

My first gut instinct was to use tuples again.  My second one is that we’re probably masking some bugs somewhere by this behaviour.  And it totally is a bug in attrs.

What to do?

hynek:
Obligatory ping to @ericvsmith since they seem to do it the same way in data classes.

ericvsmith:
I think the conflict is unresolvable. As long as you define the behavior, the user has to live with the results of depending on it. Using nans as keys in a dict or in sets has the same issue. And it's all made worse by float('nan') returning a new nan:
```
>>> s = {nan}
>>> nan in s
True
>>> float('nan') in s
False
>>> float('nan') == nan
False
>>> float('nan') is nan
False
```
I do agree it's unfortunate that changing to `and` chaining might be prevented by the wacky nan behavior.

Do you have any numbers on `and` chaining speed vs. tuple creation? I might change `dataclasses` to define comparison by `and` chaining if it's demonstrably faster.

Tinche:
My initial reaction: f*** floats (not helpful, I know). Is the chain of ands approach faster?

ericvsmith:
I sort of agree: I don't think you should do anything special for anything where `is` is `True`, but `==` is `False`. nan is the only thing that falls in this category that I know of.

There are many instances of this being raised on bpo. Here's an interesting one https://bugs.python.org/issue4296 with a quote by Mark Dickinson: 

> But in general I don't think the fact that NaNs are weird
should get in the way of optimizations.  In other words,
I'm not sure that Python should be asked to guarantee
anything more than "b == b" returning False when b is
a NaN.  It wouldn't seem unreasonable to consider
behaviour of nans in containers (sets, lists, dicts)
as undefined when it comes to equality and identity
checks.

Although he seems to walk back part of his comment, so I don't know where it stands in the end, unfortunately. Also, it's old.

hynek:
[...]

PR #306, opened 2017-12-03
Speed up __eq__ by generating code

hynek:
So this is a 1:1 implementation of the old behaviour, ignoring #305 for now.

Improvements:

```python
import attr


@attr.s(hash=True)
class C:
    x = attr.ib()
    y = attr.ib()
    z = attr.ib()


c1 = C(1, "2", 3.0)
c2 = C(1, "2", 3.0)
```

attrs 17.3.0:
```
➤ pyperf timeit --rigorous -s "from t import c1, c2" "c1 == c2"
Mean +- std dev: 3.24 us +- 0.13 us
```

New:
```
➤ pyperf timeit --rigorous -s "from t import c1, c2" "c1 == c2"
Mean +- std dev: 538 ns +- 29 ns
```

For sets:

attrs 17.3.0:
```
➤ pyperf timeit --rigorous -s "from t import c1, c2" "s = {c1, c2}; c1 in s"
Mean +- std dev: 8.56 us +- 0.32 us
```

New:
```
➤ pyperf timeit --rigorous -s "from t import c1, c2" "s = {c1, c2}; c1 in s"
Mean +- std dev: 1.82 us +- 0.09 us
```

This looks nice. :)

### DIFF
--- a/changelog.d/306.change.rst
+++ b/changelog.d/306.change.rst
@@ -1,0 +1,1 @@
+Equality tests are *much* faster now.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -701,7 +701,7 @@ def _make_hash(attrs):
         if a.hash is True or (a.hash is None and a.cmp is True)
     )
 
-    # We cache the generated init methods for the same kinds of attributes.
+    # We cache the generated hash methods for the same kinds of attributes.
     sha1 = hashlib.sha1()
     sha1.update(repr(attrs).encode("utf-8"))
     unique_filename = "<attrs generated hash %s>" % (sha1.hexdigest(),)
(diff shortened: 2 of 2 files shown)


Tinche:
The speed up looks excellent. So, an evil thought: if the class in question is a dict class, does `self.__dict __ == other.__dict__` fulfills all our requirements? 😈

hynek:
Where are you getting your ideas? ಠ_ಠ 

(For posterity: no it doesn’t :) — unless class is frozen
And all attributes are part of eq and probably a few more invariants.)

Tinche:
I admit the idea is probably not usable in practice but I'd like to get to why exactly? All attributes are in `instance.__dict__`. And what does the frozenness of the class have to do with it?

Also consider this:
```
@attr.s
class C:
    a = attr.ib()

>>> a =  C(1)
>>> b = C(1)
>>> a == b
True
>>> a.b = 2
>>> a == b
True
```

Tinche:
Just to clarify, I'm considering the eligibility of the following `__eq__` implementation (but just for dict classes, obviously):
```
def __eq__(self, other):
    return self.__dict__ == other.__dict__
```

hynek:
The main reason is that it changes the whole contract between attrs and its users.   Currently we promise we only care about attributes that are explicitly defined.   This approach would compare everything.  Frozenness technically prevents  people from tacking more attributes on instances that we shouldn’t know about.

hynek:
Now that I’m not on a phone anymore: what I mean is something like this:

```pycon
>>> import attr

>>> @attr.s(cmp=False)
... class C:
...     x = attr.ib()
...     def __eq__(self, other):
...         return self.__dict__ == other.__dict__

>>> C(1)
C(x=1)

>>> c1 = C(1)

>>> c2 = C(1)

>>> c1 == c2
True

>>> c2._cache = {}

>>> c1 == c2
False
```

Leaving aside better ways to achieve that, I’m sure people have done that and breaking backward compat for no real gain doesn’t seem a great way forward to me…

Tinche:
Yeah, yeah, point taken, I recant my suggestion. Damn users, making it hard for us :(

At this point I'd really like to know if there's any speed difference between a chain of ands and comparing tuples; think this is easy to hack together on this branch?

Tinche:
Ah I see you did the ands benchmark in another thread, and it's slower? Hm. I guess tuples are the way to go then.

hynek:
Yeah it’s def slower if the classes are equal.  The difference is in any case marginal, so we can choose what we do.  For now I’d prefer to not break backward compatibility. :)

hynek:
I’ve stopped using `"".format()` because it’s slow and ugly.  f-strings for lyfe! :)

PR #315, opened 2017-12-16
Rename convert to converter

### LINKED ISSUE #307: convert should be converter
We use proper nouns everywhere else too and this is kind of a sore thumb to me.  `attr.ib(convert=int)` should be deprecated and replaced by `attr.ib(converter=int)`.

hynek:
Fixes #307

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -90,7 +90,7 @@ Core
       ... class C(object):
       ...     x = attr.ib()
       >>> attr.fields(C).x
-      Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, convert=None, metadata=mappingproxy({}), type=None, converter=None)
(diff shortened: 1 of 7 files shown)


hynek:
anyone? 😇

btw @RonnyPfannschmidt pretty sure pytest will have to update – I hope you have activated warnings? ;)

docs/api.rst
```diff
@@ -90,7 +90,7 @@ Core
       ... class C(object):
       ...     x = attr.ib()
       >>> attr.fields(C).x
-      Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, convert=None, metadata=mappingproxy({}), type=None)
+      Attribute(name='x', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, convert=None, metadata=mappingproxy({}), type=None, converter=None)
```
wsanchez: It seems unfortunate that the `repr` string emits both `convert` and `converter` here…

hynek:
Yeah, done!  Also I’ve added a warning if someone accesses that attribute.

PR #316, opened 2017-12-17
Add __module__ and __qualname__ to methods, fix __name__

### LINKED ISSUE #309: Generated methods like __init__ lack the __module__ attribute
This issue was initially reported in https://github.com/Stewori/pytypes/issues/18.
Applying ``__module__`` on e.g. the ``__init__`` function generated by attrs raises a ``KeyError``. For more information please refer to the link above.

Being able to retrieve the module for a given function/method is relevant to locate the function or method in the class hierarchy from perspective of a decorator or profiler. Searching a module top-down for a function/method is often the only way to access its enclosing class, including nesting structure of classes.

A workaround that sometimes worked but not in this case is to get the module via ``__code__.co_filename``. However, any valid approach for identifying the corresponding module and/or location in class hierarchy for a generated function would be fine.

Please apologize if I should overlook some approach that is already viable.

hynek:
Fixes #309

Does this look good to you @Stewori?

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -468,17 +468,22 @@ class _ClassBuilder(object):
         return cls
 
     def add_repr(self, ns):
-        self._cls_dict["__repr__"] = _make_repr(self._attrs, ns=ns)
+        self._cls_dict["__repr__"] = self._add_method_dunders(
+            _make_repr(self._attrs, ns=ns)
+        )
         return self
 
     def add_str(self):
-        repr_ = self._cls_dict.get("__repr__")
-        if repr_ is None:
+        repr = self._cls_dict.get("__repr__")
+        if repr is None:
             raise ValueError(
                 "__str__ can only be generated if a __repr__ exists."
             )
 
-        self._cls_dict["__str__"] = repr_
+        def __str__(self):
+            return self.__repr__()
+
+        self._cls_dict["__str__"] = self._add_method_dunders(__str__)
         return self
 
     def make_unhashable(self):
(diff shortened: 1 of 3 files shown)


Stewori:
It looks good and thanks for solving this! However, I don't know enough about attrs to fully assess its workability from reading the diff. I suppose that if the dunders are created like expected (and like the tests ensure), this would allow me to fix https://github.com/Stewori/pytypes/issues/18 (not out of the box though, because of  ``__code__.co_filename``, see note 2).

Note 1: In ``_add_method_dunders`` you should maybe put access to ``__qualname__`` and ``__module__`` into try/catch, because I'm not sure if there can be circumstances under which these attributes are not properly populated in a class (e.g. C-extensions can be odd). Maybe this is not necessary, but from my experience strange things can always happen in Python. Remember: I once assumed a function would always have a ``__module__`` attribute. I would not want this fix to destabilize attrs under some circumstances. However, I don't know an actual example or if it can happen at all. Decide yourself ;)

Note 2: Consider if also ``__code__.co_filename`` might better link to the right module name as this is the only way to find the module for a given code object. However, I suppose this is not strictly required to fix https://github.com/Stewori/pytypes/issues/18, because the profiler has access to the function object and thus to ``__module__``. This is just a "nice to have".

src/attr/_make.py
```diff
@@ -468,43 +468,68 @@ def slots_setstate(self, state):
...
+            cd["__ge__"] = (
+                self._add_method_dunders(meth)
+                for meth in _make_cmp(self._attrs)
+            )
 
         return self
 
+    def _add_method_dunders(self, meth):
+        """
+        Add __module__ and __qualname__ to a method *method*.
```
blueyed: `s/*method*/*meth*/` ?

hynek:
>  I once assumed a function would always have a __module__ attribute.

Excellent point. :D

I wonder, if the class hasn't those attributes, is it better to leave them gone or to set them to None?  The later sounds cleaner but like something that might mask subtle bugs.

Is there any precedent?  I’ve chosen to leave it off for now.  Turns out, testing that behaviour is harder than it sounds. :)

> Note 2: Consider if also __code__.co_filename might better link to the right module name as this is the only way to find the module for a given code object. However, I suppose this is not strictly required to fix Stewori/pytypes#18, because the profiler has access to the function object and thus to __module__. This is just a "nice to have".

I’m not sure I 100% follow here.  Do you want me to *set* co_filename or to use it?  The generated methods do have it but since they’re not in actual files, it’s set to something like `'<attrs generated init 97d170e1550eee4afc0af065b78cda302a97674c>'` which is used in linecache to look them up if a user wants to pdb thru.

Stewori:
Regarding ``co_filename`` it's okay like it is, if it would clash with another use case otherwise :)
Being able to conclude the module from a code object would be nice, but is not crucial.

PR #318, opened 2017-12-27
Use extras_require instead of dev-requirements.txt

hynek:
It’s time for attrs to stop pretending we're an application. :)

I’m just not sure about the name; what is more common?  `test` or `tests`?


Tinche:
Someone contributed a change like this to pytest-asyncio, they called the extra 'testing'. pipenv uses '--dev'. Flit uses 'dev-requires'. Maybe standardize on 'dev'?

hynek:
- cryptography has `test`
- Twisted has `dev`
- pip has `testing`
- Pyramid has `testing`
- Flask has `dev`

***

I tend to like `dev`, but it doesn’t really reflect the fact, that it’s dependencies needed to run tests.  It’s rather implied.  🤔

I’ve invited a public bike shed: https://twitter.com/hynek/status/946022597224955904 :)

hynek:
JFTR, I remember why I disliked `dev`: sometimes it makes sense to install packages in a dev-mode: like `structlog` had forever with colorama, a web framework may have a debug toolbar etc.

That’s also stuff that belongs to pipenv’s dev, but it relates to the user’s concept of dev, not of the package.

Does it make any sense?

Tinche:
On the other hand, things like Sphinx and maybe something for benchmarking don't directly correspond to 'testing'. Sure Sphinx runs your doctests but its primary purpose is generating documentation. Linters also don't fall directly in the category of testing. Ideally your dev environment should contain everything we expect for development. If your contribution passes tests but crashes the docs, it's not going to get looked at.

Tinche:
On the third hand we might be bikeshedding this way too much. I'm basically OK with any of your Twitter poll options.

Tinche:
Also I see now we have a special docs-requirements.txt. Do we really need to be so granular with our deps?

hynek:
The docs thing is for RTD mainly.

hynek:
So my current inclination is to call it `tests` and `docs` to mirror the names of the directories.  Maybe plus `dev` that contains both.

PR #332, opened 2018-01-26
Make NOTHING's hash vegan

### LINKED ISSUE #331: Hashing NOTHING in CPython is 60% slower than it should be on 32 bit builds
```
class DeadBeef:
    def __hash__(self):
        return 0xdeadbeef

class LeetBeef:
    def __hash__(self):
        return 0x5eadbef0

a = Timer('hash(self)', 'from __main__ import DeadBeef;self=DeadBeef()')
b = Timer('hash(self)', 'from __main__ import LeetBeef;self=LeetBeef()')

print('Dead:', hex(hash(DeadBeef())), min(a.repeat()))
print('Leet:', hex(hash(LeetBeef())), min(b.repeat()))

```

Don't know how/want to learn how to contribute to projects on here, but I would feel bad if I didn't at least mention on here an oversight of this magnitude.

hynek:
...and fit into a 32bit integer.

Fixes #331

@mr-nfamous I do not have access to a 32bit Python build – could you verify whether this does what you want? Thank you.

### DIFF
--- a/changelog.d/331.change.rst
+++ b/changelog.d/331.change.rst
@@ -1,0 +1,1 @@
+The hash of ``attr.NOTHING`` is now vegan and faster on 32bit Python builds.
--- a/changelog.d/332.change.rst
+++ b/changelog.d/332.change.rst
@@ -1,0 +1,1 @@
+The hash of ``attr.NOTHING`` is now vegan and faster on 32bit Python builds.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -45,7 +45,7 @@ class _Nothing(object):
         return "NOTHING"
 
     def __hash__(self):
-        return 0xdeadbeef
+        return 0xc0ffee
 
 
 NOTHING = _Nothing()


Tinche:
You can do something like 'docker run --rm -it -v (pwd):/attrs i686/ubuntu bash' and test it that way (assuming you have Docker for Mac). I can run this later if you don't get the chance to.

hynek:
Right.

```
$ /app/bin/python --version
Python 2.7.6
$ file /app/bin/python
/app/bin/python: ELF 32-bit LSB  executable, Intel 80386, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.24, BuildID[sha1]=9494a18dfca1d00187b5b87406cbe1fd7b205e3a, stripped
```

Old:

```
$ /app/bin/pyperf timeit --rigorous -s 'from attr import NOTHING' 'hash(NOTHING)'
Mean +- std dev: 141 ns +- 12 ns
```

New:

```
$ /app/bin/pyperf timeit --rigorous -s 'from attr import NOTHING' 'hash(NOTHING)'
Mean +- std dev: 139 ns +- 6 ns
```

Meh.

***

```
$ /app/bin/python3 --version
Python 3.4.3
$ file /app/bin/python3
/app/bin/python3: ELF 32-bit LSB  executable, Intel 80386, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.24, BuildID[sha1]=fbc4179873a2c79e95c065734797516db21fb00e, stripped
```

Old:

```
Mean +- std dev: 245 ns +- 8 ns
```

New:

```
Mean +- std dev: 169 ns +- 4 ns
```

Uh.

***

```
$ /app/bin/python --version
Python 3.6.4
/usr/bin/python3.6: ELF 32-bit LSB  executable, Intel 80386, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.24, BuildID[sha1]=a2e460c7e48bd5ac714d4acefcb01d1091bb8ef6, stripped
```

Old:

```
Mean +- std dev: 314 ns +- 20 ns
```

New:

```
Mean +- std dev: 212 ns +- 11 ns
```

Oh.

***

So I guess this is it?

(I’m gonna pretend the fact that hashing got much slower with Python 3.6 is only true for 32bits. #denial)

ambv:
A confused and embarassed man saying "OK" repeatedly

PR #337, opened 2018-01-29
Explain our position on Data Classes

hynek:
Since I get asked almost daily whether `attrs` will survive Python 3.7, I guess it’s time for a section on why we dare to think it’s worthwhile to not delete `attrs` from PyPI the moment 3.7 drops.

I’d like to ask @ericvsmith to rubber stamp this (or suggest changes) so it’s obvious it’s not some nefarious fake news propaganda.

If anyone can think of more reasons, please comment.

### DIFF
--- a/docs/why.rst
+++ b/docs/why.rst
@@ -147,10 +147,9 @@ Nevertheless, there are still reasons to prefer ``attrs`` over Data Classes whos
 
 - ``attrs`` supports all maintream Python versions, including CPython 2.7 and PyPy.
 - Data Classes are intentionally less powerful than ``attrs``.
-  There is a long list of features that were sacrificed for the sake of simplicity.
-  The most obvious ones are validators, converters, and metadata but it permeates throughout all APIs.
+  There is a long list of features that were sacrificed for the sake of simplicity and while the most obvious ones are validators, converters, and metadata, it permeates throughout all APIs.
 
-  On the other hand, Data Classes currently do not offer any significant feature that ``attrs`` doesn't already already have.
+  On the other hand, Data Classes currently do not offer any significant feature that ``attrs`` doesn't already have.
 - ``attrs`` can and will move faster.
   We are not bound to any release schedules and we have a clear deprecation policy.
 
--- a/docs/why.rst
+++ b/docs/why.rst
@@ -135,6 +135,29 @@ With ``attrs`` your users won't notice a difference because it creates regular, 
 .. _behaving like a tuple: https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences
 
 
+…Data Classes?
+--------------
+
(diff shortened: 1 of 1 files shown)


ericvsmith:
This generally looks good to me.

Just a few comments:

- `dataclasses` does support per-field metadata. I don't know if this puts it on par with `attrs` or not.

- Does `attrs` support the equivalent of `dataclasses`'s `InitVar` or `ClassVar`? My point being that I'm not sure it's true that `dataclasses` has no significant features that `attrs` doesn't have. I guess it depends on your definition of significant: these won't be commonly used features.

I think it's useful to have this all written down, and I agree with the overall sentiment. Thanks for writing it up!

docs/why.rst
```diff
@@ -135,6 +135,28 @@ With ``attrs`` your users won't notice a difference because it creates regular,
 .. _behaving like a tuple: https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences
 
 
+…Data Classes?
+--------------
+
+`PEP 557 <https://www.python.org/dev/peps/pep-0557/>`_ added Data Classes to `Python 3.7 <https://docs.python.org/3.7/whatsnew/3.7.html#pep-557-data-classes>`_ that resemble ``attrs`` in many ways.
+
+They are the result of the Python community's `wish <https://mail.python.org/pipermail/python-ideas/2017-May/045618.html>`_ to have to have an easier way to write classes in the standard library that doesn't carry the problems of ``namedtuple``\ s.
```
blueyed: s/to have to have/to have/

hynek:
> dataclasses does support per-field metadata. I don't know if this puts it on par with attrs or not.

Oh, the PEP says it doesn't.

> Does attrs support the equivalent of dataclasses's InitVar or 

Hmm, that’s basically

```python
>>> @attr.s(auto_attribs=True)
... class C:
...     i: int
...     j: int = None
...     database: DataBaseType = attr.ib(default=None, repr=False, cmp=False)
...
...     def __attrs_post_init__(self):
...        if self.j is None and self.database is not None:
...             self.j = self.database.lookup('j')
...        del self.database
```

right? I grant you it’s not super elegant we could consider something similar but it seems at least *possible* to me?

> ClassVar?

The point of this one it to exclude an annotated variable, right? Like this:

```python
>>> import attr

>>> @attr.s(auto_attribs=True)
... class C:
...     x: typing.ClassVar = 42
...     y: int

>>> C(5)
C(y=5)
```

? I seem to remember that we shared ways to implement while avoiding to import typing. :)

ericvsmith:
Correct on ClassVar. `dataclasses` still a problem with PEP 563 and `ClassVar`.

Looks like my points are covered.

Oh, just thought of another one: you should mention that `dataclasses` doesn't support `__slots__`.

hynek:
OK cool, I’ll just replace `metadata` with `__slots__` and call it a day.

PR #345, opened 2018-02-05
attr.asdict(recurse=True) now can handle nested lists

Pr0Ger:
Right now if field contains nested lists it will not properly encoded to dict:
```python
>>> @attr.s
... class Foo():
...     a = attr.ib()
...
>>> a = Foo(a=[[Foo(a=1)]])
>>> attr.asdict(a)
{'a': [[Foo(a=1)]]}
```

I added support for this case (only for lists and only for `asdict` function) and also updated testing strategy

### DIFF
--- a/changelog.d/345.change.rst
+++ b/changelog.d/345.change.rst
@@ -1,0 +1,1 @@
+``attr.asdict(recurse=True)`` now can handle nested lists
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -38,6 +38,16 @@ def asdict(inst, recurse=True, filter=None, dict_factory=dict,
     """
     attrs = fields(inst.__class__)
     rv = dict_factory()
+
+    def encode_sequence_item(item):
+        if has(item.__class__):
+            return asdict(item, recurse=True, filter=filter,
+                          dict_factory=dict_factory)
+        elif isinstance(item, (tuple, list, set)):
+            cf = item.__class__ if retain_collection_types is True else list
+            return cf(map(encode_sequence_item, item))
+        return item
+
     for a in attrs:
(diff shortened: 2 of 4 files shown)


Tinche:
Hi,

I see two problems with this solution.

The first is we're making a new function every asdict invocation, which slows things down. The second is, it doesn't solve the problem fully. Consider this:

```
>>> b = Foo([{1: Foo(1)}])
>>> asdict(b)
{'a': [{1: Foo(a=1)}]}
```

So if we're gonna solve it, let's solve it right. I've pushed a branch with what I think is a more correct solution: https://github.com/python-attrs/attrs/tree/feature/recursive-asdict. It makes asdict in general a little slower, but it's faster than this solution.

```
Before:
> pyperf timeit -g -s "from attr import asdict; from random import Random; from tests.utils import nested_classes, simple_classes; r = Random(1); C = simple_classes().example(r); c=C()" "asdict(c)"
Mean +- std dev: 2.43 us +- 0.09 us

After:
> pyperf timeit -g -s "from attr import asdict; from random import Random; from tests.utils import nested_classes, simple_classes; r = Random(1); C = simple_classes().example(r); c=C()" "asdict(c)"
Mean +- std dev: 2.51 us +- 0.10 us

This branch:
> pyperf timeit -g -s "from attr import asdict; from random import Random; from tests.utils import nested_classes, simple_classes; r = Random(1); C = simple_classes().example(r); c=C()" "asdict(c)"
Mean +- std dev: 2.95 us +- 0.10 us

```

Please add the tests to that branch and we can get this merged.

hynek:
Ping? 😇

hynek:
Fixed in #395 – thank you for the inspiration.

PR #349, opened 2018-02-27
Add attr.fields_dict()

### LINKED ISSUE #290: Possible to add a "field_names" helper function?
In a specific use case, I need to get all field names that are defined in an attrs class, which is as simple as
```
[f.name for f in attr.fields(MyClass)]
```
Do you think it's OK to add a public API for this?

rouge8:
This patch adds `attr.fields_dict()`, which returns an `OrderedDict` of `{attribute_name: Attribute}` for all attrs attributes of a class.

Closes #290.

I'm bad at describing dictionaries in prose, so feel free to improve. :)

 # Pull Request Check List

- [x] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/utils.py).

### DIFF
--- a/changelog.d/290.change.rst
+++ b/changelog.d/290.change.rst
@@ -1,0 +1,1 @@
+Added ``attr.field_dict()`` to return an ``OrderedDict` of ``attrs`` attributes for a class, whose keys are the attribute names.
--- a/changelog.d/290.change.rst
+++ b/changelog.d/290.change.rst
@@ -1,1 +1,1 @@
-Added ``attr.field_dict()`` to return an ``OrderedDict` of ``attrs`` attributes for a class, whose keys are the attribute names.
+Added ``attr.field_dict()`` to return an ordered dictionary of ``attrs`` attributes for a class, whose keys are the attribute names.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1068,6 +1068,32 @@ def fields(cls):
     return attrs
 
 
+def fields_dict(cls):
+    """
+    Returns an OrderedDict of ``attrs`` attributes for a class, whose keys are
+    the attribute names.
+
+    :param type cls: Class to introspect.
+
+    :raise TypeError: If *cls* is not a class.
+    :raise attr.exceptions.NotAnAttrsClassError: If *cls* is not an ``attrs``
+        class.
+
+    :rtype: OrderedDict where keys are attribute names and values are
+        :class:`attr.Attribute`\ s.
+
+    .. versionadded:: 18.1.0
+    """
+    if not isclass(cls):
+        raise TypeError("Passed object must be a class.")
+    attrs = getattr(cls, "__attrs_attrs__", None)
+    if attrs is None:
+        raise NotAnAttrsClassError(
+            "{cls!r} is not an attrs-decorated class.".format(cls=cls)
+        )
+    return ordered_dict([(a.name, a) for a in attrs])
+
+
 def validate(inst):
     """
     Validate all attributes on *inst* that have a validator.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1042,7 +1042,7 @@ def _add_init(cls, frozen):
 
 def fields(cls):
     """
-    Returns the tuple of ``attrs`` attributes for a class.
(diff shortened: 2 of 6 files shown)


src/attr/_make.py
```diff
@@ -1068,6 +1068,32 @@ def fields(cls):
...
+    .. versionadded:: 18.1.0
+    """
+    if not isclass(cls):
+        raise TypeError("Passed object must be a class.")
+    attrs = getattr(cls, "__attrs_attrs__", None)
+    if attrs is None:
+        raise NotAnAttrsClassError(
+            "{cls!r} is not an attrs-decorated class.".format(cls=cls)
+        )
+    return ordered_dict([(a.name, a) for a in attrs])
```
hynek: Shouldn't it be possible to avoid the list and just make it a generator? (iow drop the square brackets)

changelog.d/290.change.rst
```diff
@@ -0,0 +1 @@
+Added ``attr.field_dict()`` to return an ``OrderedDict` of ``attrs`` attributes for a class, whose keys are the attribute names.
```
hynek: I think we shouldn’t speak of ``OrderedDict`` since that’s not true. Just say “ordered dict” or “ordered dictionary”. I’m not gonna point it out in every instance of your PR to avoid clutter, but please get rid of it. :)

src/attr/_make.py
```diff
@@ -1068,6 +1068,32 @@ def fields(cls):
     return attrs
 
 
+def fields_dict(cls):
+    """
+    Returns an OrderedDict of ``attrs`` attributes for a class, whose keys are
```
hynek: Ugh I’ve just noticed the “Returns” in `fields`! Please make both a “Return”, thanks!

tests/test_make.py
```diff
@@ -676,6 +677,40 @@ def test_fields_properties(self, C):
...
+            fields_dict(C(1, 2))
+
+        assert "Passed object must be a class." == e.value.args[0]
+
+    def test_handler_non_attrs_class(self, C):
+        """
+        Raises `ValueError` if passed a non-``attrs`` instance.
+        """
+        with pytest.raises(NotAnAttrsClassError) as e:
+            fields_dict(object)
```
hynek: empty line before assert block would be nice

tests/test_make.py
```diff
@@ -676,6 +677,40 @@ def test_fields_properties(self, C):
...
+            "{o!r} is not an attrs-decorated class.".format(o=object)
+        ) == e.value.args[0]
+
+    @given(simple_classes())
+    def test_fields_dict(self, C):
+        """
+        Returns an ordered dict of ``{attribute_name: Attribute}``.
+        """
+        d = fields_dict(C)
+        assert isinstance(d, ordered_dict)
```
hynek: empty line before assert block would be nice

src/attr/_make.py
```diff
@@ -1068,6 +1068,32 @@ def fields(cls):
...
+    the attribute names.
+
+    :param type cls: Class to introspect.
+
+    :raise TypeError: If *cls* is not a class.
+    :raise attr.exceptions.NotAnAttrsClassError: If *cls* is not an ``attrs``
+        class.
+
+    :rtype: OrderedDict where keys are attribute names and values are
```
hynek: In this case it’s worth explaining what happens (something along of “a plain dict if it's naturally ordered like in Python 3.6+, `OrderedDict` otherwise”)

tests/test_make.py
rouge8: want me to clean these up in the tests for `fields` too? that's where I copy/pasted from 😉

rouge8:
@hynek updated 🎉

tests/test_make.py
hynek: Sure, #boyScout. :)

rouge8: (done ✔️)

hynek:
Have you pushed? 😇

rouge8:
I think so? The review comments show as outdated except for https://github.com/python-attrs/attrs/pull/349#discussion_r170957703 which is on the line above the newline

hynek:
Thanks!

PR #350, opened 2018-03-03
Ignore build/

ngie-eign:
This change ignores the `build/` directory, created by `python setup.py build`, with `git status`, etc.

While here, sort .gitignore entries.


hynek:
I guess it’s not in there because we don’t build anything so it never came up? Anyways sorting is probably a good idea so thanks. :)

PR #351, opened 2018-03-03
Fix warning from py3 regarding "improper regex escapes"

ngie-eign:
py3 emits warnings about strings which contain improper character escapes. This silences one of those warnings, while keeping the desired documentation output.


hynek:
I’m kind of baffled I haven’t caught that warning before. :|. But this doesn’t work. I think the correct the approach is to make the string raw?

PR #358, opened 2018-03-16
No recursion error for cycles in `repr`

### LINKED ISSUE #95: Infinite Recursion in repr with bi-directional relationships
Take a fairly trivial example (which has obviously been simplified to the point of absurdity):

``` python
import attr

@attr.s
class Test():
    attr = attr.ib(default=None)

test = Test()
other = Test()
test.attr = other
other.attr = test

print(repr(test))
```

This will result in infinite recursion. This can be worked around by setting `repr=False` in the attr.s decorator.

I've sat on this bug for a week trying to come up with a good solution, and I can't think of anything good. I'm starting to think the best solution is to add a warning to the documentation of this behavior and calling it good.

glyph:
Fixes #95.

> - [x] Added **tests** for changed code.

Done.

There are no tests for the multithreaded case because it's literally impossible to test that, you just have to think really hard.  I could try harder if the reviewer thinks that this implementation strategy is risky but it seems pretty boring; just using threading.local to avoid sharing state between threads.

> - [ ] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/utils.py).

I don't *think* I need to do this here because I don't see anything relevant, but, I am not a Hypothesis expert so I might be missing something.

> - [ ] Updated **documentation** for changed code.

It's a bugfix, so I don't think it calls for updated docs.

Done.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -995,19 +995,15 @@ def _make_repr(attrs, ns):
         # like weakref- or hash-ability.
         working_set.add(id(self))
         try:
-            result = class_name
-            result += "("
+            result = [class_name, "("]
             first = True
             for name in attr_names:
                 if first:
                     first = False
                 else:
-                    result += ", "
-                result += name
-                result += "="
-                result += repr(getattr(self, name, NOTHING))
-            result += ")"
-            return result
+                    result.append(", ")
+                result.extend((name, "=", repr(getattr(self, name, NOTHING))))
+            return "".join(result) + ")"
         finally:
             working_set.remove(id(self))
     return __repr__
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -4,6 +4,7 @@ import hashlib
 import linecache
 import sys
 import warnings
+import threading
 
 from operator import itemgetter
 
@@ -952,6 +953,7 @@ def _add_cmp(cls, attrs=None):
 
     return cls
 
+_already_repring = threading.local()
 
 def _make_repr(attrs, ns):
     """
@@ -967,6 +969,14 @@ def _make_repr(attrs, ns):
         """
         Automatically created by attrs.
         """
+        try:
+            working_set = _already_repring.working_set
+        except AttributeError:
+            working_set = set()
+            _already_repring.working_set = working_set
+
+        if id(self) in working_set:
+            return "..."
         real_cls = self.__class__
         if ns is None:
             qualname = getattr(real_cls, "__qualname__", None)
@@ -977,13 +987,21 @@ def _make_repr(attrs, ns):
         else:
             class_name = ns + "." + real_cls.__name__
 
-        return "{0}({1})".format(
-            class_name,
-            ", ".join(
-                name + "=" + repr(getattr(self, name, NOTHING))
-                for name in attr_names
+        # Since 'self' remains on the stack (i.e.: strongly referenced) for the
+        # duration of this call, it's safe to depend on id(...) stability, and
+        # not need to track the instance and therefore worry about properties
+        # like weakref- or hash-ability.
+        working_set.add(id(self))
+        try:
+            return "{0}({1})".format(
+                class_name,
+                ", ".join(
+                    name + "=" + repr(getattr(self, name, NOTHING))
+                    for name in attr_names
+                )
             )
-        )
+        finally:
+            working_set.remove(id(self))
     return __repr__
 
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -995,13 +995,19 @@ def _make_repr(attrs, ns):
         # like weakref- or hash-ability.
         working_set.add(id(self))
         try:
-            return "{0}({1})".format(
-                class_name,
-                ", ".join(
-                    name + "=" + repr(getattr(self, name, NOTHING))
-                    for name in attr_names
-                )
-            )
+            result = class_name
+            result += "("
+            first = True
+            for name in attr_names:
+                if first:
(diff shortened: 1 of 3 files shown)


asvetlov:
Is `repr` speed the bottleneck?
Usually `repr` is used for logging, errors handling etc. but never for not fast paths if successful code execution.

Guys, it is your library, you establish the rule.
But for example if comprehensive repr in aiohttp takes extra time -- I don't care, correctness first.

glyph:
Re: logging attrs classes: yes, this is precisely why I really want this to happen :).  Sometimes having `repr` in a hot logging loop be kinda slow is vastly preferable to sometimes entirely losing a log message because an object got into a cycle somehow.

glyph:
Anyone I can motivate to give me a review? :)

glyph:
@asvetlov Thanks for your input, your review is appreciated! But per project policy, I do need another project member to sign off :).

Tinche:
So show us some performance numbers?

Tinche:
BTW I log attrs classes all the time on the happy path. Not in hot loops of course, but once per request for sure. A slow repr isn't prohibitive, but maybe it should be opt-in. Personally I have never had a cycle, why pay the price?

glyph:
Preliminary testing suggests a 20% slowdown on both CPython and PyPy.

Test:

```python
import attr
from textwrap import dedent

@attr.s
class Foo(object):
    bar = attr.ib()
    baz = attr.ib()
    boz = attr.ib()

import timeit
print(timeit.timeit(
    'repr(foo)',
    dedent('''
    from __main__ import Foo
    foo = Foo(1, "hello", ["some", "values"])
    '''),
    number=1000000
))
```

glyph:
@Tinche On the one hand, I like the philosophy of "don't pay for what you don't use".  On the other hand, what would you call this if you had to opt in to it?  `dont_crash_on_repr`?  I don't like "behave correctly" flags.  Not sure how to resolve that here, and on balance I feel like this is acceptable overhead for starters, and if we wanted to optimize repr there are a few other things which might be a good idea to investigate as well.

glyph:
How do you feel about `frozen=` being the opt-in?

glyph:
Actually I guess since `frozen=False` is the default that would make it an opt-*out* not an opt-in.

glyph:
@Tinche With the micro-optimization I just committed, this change is about 25% faster on PyPy and about 10% faster on CPython.

It's true that if we removed the code that makes it correct now it could probably be *even faster*, but presumably if the previous performance was acceptable, faster performance is also acceptable :).

glyph:
(25%/10% faster than `master`, I mean, not than the previous implementation of this fix)

mahmoud:
Restraining the optimization golfer in me, I think eliminating  the `.format()` call and generator expression is a cromulent and uncontroversial way to address repr performance concerns. +1

hynek:
Let’s get some numbers.

# Hynek’s Molasses

```
➤ pyperf timeit --rigorous -s 'import attr; C = attr.make_class("C", ["a", "b", "c"]); x = C(1, "2", 3.0)' 'repr(x)'
Mean +- std dev: 3.22 us +- 0.09 us
```

# Glyph’s Vroom-Vroom

```
➤ pyperf timeit --rigorous -s 'import attr; C = attr.make_class("C", ["a", "b", "c"]); x = C(1, "2", 3.0)' 'repr(x)'
Mean +- std dev: 3.12 us +- 0.08 us
```

# Hawkie’s Turbo

However, looking at the code reminded me of @hawkowl ’s [blog post](https://atleastfornow.net/blog/not-all-bytes/) on string concats (well, she wrote about byte strings but I figured the lessons are transferable) and I did some more micro optimizations that even made the code shorter and nicer. And lo and behold:

```
pyperf timeit --rigorous -s 'import attr; C = attr.make_class("C", ["a", "b", "c"]); x = C(1, "2", 3.0)' 'repr(x)'
Mean +- std dev: 2.99 us +- 0.07 us
```

So I’m not sure what the etiquette here is but I’ve just pushed my changes against Glyph’s branch. If you (Glyph) are OK with it, we can merge it as soon as CI passes.

src/attr/_make.py
```diff
@@ -3,6 +3,7 @@
 import hashlib
 import linecache
 import sys
+import threading
```
hynek: I still hate you for this import.

glyph: Not as much as I hate myself.

hynek:
Although I have one last question: does this work with greenlets too? Or do we have to do something like I did in [structlog](https://github.com/hynek/structlog/blob/38841503567bd892fb643f5ff53cfd454006d125/src/structlog/threadlocal.py#L17-L47)?

glyph:
It should work well enough with greenlets - in principle it's possible to suspend and reentrantly repr something in a different greenlet but the same thread but:

- you'd have to do something *truly* horrible to make that happen
- if you're doing that reentrantly in a paused greenlet then you might have the same issues with recursion anyway so this logic is not *wrong* exactly.

glyph:
If we're happy with the perf numbers, I'm going to go ahead and merge.

hawkowl:
I helped, without even having to help. Woo!

hynek:
“Hawkie’s Turbo” is a thing now

njsmith:
In 3.7 you could use `contextvars` instead, to avoid saying the Forbidden Word :-).

(I haven't managed to convince Yury to make a backport yet though.)

Tinche:
Yay, everyone's happy!

PR #363, opened 2018-03-26
Annotate __init__ with type hints

### LINKED ISSUE #249: Annotate __init__ with type hints?
So I'm not sure this makes 100% sense, but technically we could start adding type hints to the signature of our `__init__`.

I guess there would be some runtime introspection benefits?

euresti:
This just adds the annotations found at run-time to the
`__annotations__` attribute of the created `__init__` function

Fixes #249

# Pull Request Check List

### DIFF
--- a/changelog.d/363.change.rst
+++ b/changelog.d/363.change.rst
@@ -1,1 +1,1 @@
-Generated `__init__` methods now have an `__annotations__` attribute derived from the types of the fields.
+Generated ``__init__`` methods now have an ``__annotations__`` attribute derived from the types of the fields.
--- a/changelog.d/363.change.rst
+++ b/changelog.d/363.change.rst
@@ -1,0 +1,1 @@
+Generated `__init__` methods now have an `__annotations__` attribute derived from the types of the fields.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1261,7 +1261,7 @@ def _attrs_to_init_script(attrs, frozen, slots, post_init, super_attr_map):
     # This is a dictionary of names to validator and converter callables.
     # Injecting this into __init__ globals lets us avoid lookups.
     names_for_globals = {}
-    annotations = {}
+    annotations = {'return': None}
 
     for a in attrs:
         if a.validator:
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1034,7 +1034,7 @@ def _make_init(attrs, post_init, frozen, slots, super_attr_map):
         sha1.hexdigest()
     )
 
-    script, globs = _attrs_to_init_script(
+    script, globs, annotations = _attrs_to_init_script(
         attrs,
         frozen,
         slots,
@@ -1063,7 +1063,9 @@ def _make_init(attrs, post_init, frozen, slots, super_attr_map):
         unique_filename,
     )
 
-    return locs["__init__"]
+    __init__ = locs["__init__"]
+    __init__.__annotations__ = annotations
+    return __init__
 
 
 def _add_init(cls, frozen):
@@ -1259,6 +1261,7 @@ def _attrs_to_init_script(attrs, frozen, slots, post_init, super_attr_map):
     # This is a dictionary of names to validator and converter callables.
     # Injecting this into __init__ globals lets us avoid lookups.
     names_for_globals = {}
+    annotations = {}
 
     for a in attrs:
         if a.validator:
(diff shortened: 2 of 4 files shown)


src/attr/_make.py
```diff
@@ -1259,6 +1261,7 @@ def fmt_setter_with_converter(attr_name, value_var):
     # This is a dictionary of names to validator and converter callables.
     # Injecting this into __init__ globals lets us avoid lookups.
     names_for_globals = {}
+    annotations = {}
```
euresti: Do we want to also add the return annotation i.e. `{'return': None}`

src/attr/_make.py
```diff
@@ -1349,6 +1352,9 @@ def fmt_setter_with_converter(attr_name, value_var):
             else:
                 lines.append(fmt_setter(attr_name, arg_name))
 
+        if a.init is True and a.converter is None and a.type is not None:
```
euresti: If there's a converter I don't include an annotation.  I don't think it's wise to try to follow the annotation chain.  Mypy itself has a hard time doing it too.

hynek:
I’m vacationing until Easter.

🇮🇹 is pretty awesome. #protip #lifehack

Maybe you can bribe Tin to have a look. 😇 

I trust the approach is sound.

src/attr/_make.py
```diff
@@ -1063,7 +1063,9 @@ def _make_init(attrs, post_init, frozen, slots, super_attr_map):
         unique_filename,
     )
 
-    return locs["__init__"]
+    __init__ = locs["__init__"]
+    __init__.__annotations__ = annotations
```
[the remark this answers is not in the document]
Zac-HD: FWIW Hypothesis propagates `__annotations__` on any version of Python, and will try to use them for type-driven strategy inference on Python 2 too.  So we'd appreciate the compatible version :smile:

src/attr/_make.py
euresti: Well you don't want the return value. You want the type of the first argument.  

I played around with it and it's pretty complicated because some converters don't have `__annotations` (e.g. `int`) and some have an implicit first argument like `self` or `cls`.  `inspect.signature` does the right thing but that call is pretty expensive.

Zac-HD: Fair enough - if performance is a consideration here I agree that skipping annotations is the best option when a converter is present.

hynek:
Also obligatory ping @ambv

src/attr/_make.py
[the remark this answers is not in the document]
ambv: I think doing it on both Python versions is fine.

src/attr/_make.py
ambv: Yes, please include 'return', too.

src/attr/_make.py
ambv: I agree with euresti, converters are hard. We can add those later, on the basis whether the **converter** is a type. If it is, just put the converter directly as an annotation. If it's not a type, then try to get the annotation of the first argument to the callable. But it's going to be a fun pull request on its own.

ambv: But FWIW I'm not worried about performance here, it's essentially just a few additional dict lookups. To notice any adverse effect, a codebase would have to have thousands of fields with converters. And even then, we're only talking about slight increase in startup time.

euresti: FYI `converter=<some type>` is actually pretty hard because what you actually want are all the types that the type accepts as a first argument.   (e.g. `int` would be `Union[float, str, int, bool]` and probably a bunch of other things.)

ambv: Well, now we're opening a can of worms, which is asymmetric types, e.g. the setter accepts a broader spectrum of types than the getter. In __annotations__, there is no way to fully express this. The fields will always have the final type (the one after conversion). You can make unions for __init__'s arguments but that won't cover the case where users modify fields by assignment.

And, as you said, plucking the unions from types is going to require a registry for C types, including builtins. And if user classes aren't annotated, we know nothing besides the final type.

My intuition is to just handle final types all around and see what users say later.

changelog.d/363.change.rst
```diff
@@ -0,0 +1 @@
+Generated `__init__` methods now have an `__annotations__` attribute derived from the types of the fields.
```
hynek: This is markdown, not rst. :)

hynek:
So um, is everyone happy with the current state?

ambv:
Yes, this is good to go. The only minor disagreement we have is that David decided not to fill in `__annotations__` if any converter is used, based on the fact that converters can either be arbitrary callables (which might themselves not be annotated), or better yet, C types that don't have annotations at all or for which they would be very complicated.

I think we could do better than giving up in this scenario but we can iterate on this in a future version. I agree with David that doing it comprehensively is a rather complex project.

So, +1 from me as is.

PR #364, opened 2018-03-27
Suppress HealthCheck.too_slow for test_not_none_metadata

mattst88:
On very slow hardware, these may time out.

Bug: https://bugs.gentoo.org/651794


wsanchez:
It might be better to have a global thing like:

```python
settings.register_profile(
    "patience", settings(suppress_health_check=[HealthCheck.too_slow])
)
settings.load_profile("patience")
```

And make the `load_profile` call conditional on an environment variable, rather than trying to suppress this for specific tests.  (All tests will time out on slow enough hardware.)

hynek:
Yeah I think I agree with @wsanchez here. Trying to suppress only certain tests is a game of whac-a-mole. It should be all or none.

leio:
Then push the relevant changes please?

wsanchez:
@hynek: Should we suppress these all the time or try to limit that to CI environments (eg. "CI" env var is set)?  I kinda lean towards the former; wall-clock timeouts like these aren't my favorite.

hynek:
Hm ISTM that it’s kind of a feature in local dev that some tests are not optimal.

So I’d tend to an env var.  But would like  input from @Tinche and @DRMacIver on it.

PR #367, opened 2018-04-11
Support `from typing import ClassVar`

### LINKED ISSUE #361: Support ClassVar string annotations
The following doesn't work in Python 3.7

```
import attr
from typing import ClassVar

@attr.dataclass
class A:
    x: ClassVar[int]

a = A()
```

Because `ClassVar[int]` will be `'ClassVar[int]` in `A.__annotations__` and so `return str(annot).startswith("typing.ClassVar")` returns `False`.

You can repro without 3.7 using this:

```
import attr
from typing import ClassVar

@attr.dataclass
class A:
    x: 'ClassVar[int]'

a = A()
```

Note: It's weird I'm not giving it a value but if I do then attrs will think it's a default value and the code won't error out.

Oh and the following works fine:
```
import attr
import typing

@attr.dataclass
class A:
    x: 'typing.ClassVar[int]'

a = A()
```

ambv:
Also updated the docstring to reflect why exactly we're doing what we're doing.
The previous comment was incorrect (`typing` is already imported in
applications using annotations).

I added `t.ClassVar` as well which is the only third alternative import I found
in use for typing-related classes.

Fixes #361

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -23,6 +23,8 @@ _obj_setattr = object.__setattr__
 _init_converter_pat = "__attr_converter_{}"
 _init_factory_pat = "__attr_factory_{}"
 _tuple_property_pat = "    {attr_name} = property(itemgetter({index}))"
+_classvar_prefixes = ("typing.ClassVar", "t.ClassVar", "ClassVar")
+
 _empty_metadata_singleton = metadata_proxy({})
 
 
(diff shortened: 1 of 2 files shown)


src/attr/_make.py
```diff
@@ -229,13 +229,13 @@ class MyClassAttributes(tuple):
 
 
 def _is_class_var(annot):
-    """
-    Check whether *annot* is a typing.ClassVar.
+    """Check whether *annot* is a typing.ClassVar.
```
hynek: This is not how we write docstrings around here! 👨‍🏫

src/attr/_make.py
```diff
@@ -229,13 +229,13 @@ class MyClassAttributes(tuple):
...
 
-    The implementation is gross but importing `typing` is slow and there are
-    discussions to remove it from the stdlib alltogether.
+    The string comparison hack is used to avoid evaluating all string
+    annotations which would put attrs-based classes at a performance
+    disadvantage compared to plain old classes.
     """
-    return str(annot).startswith("typing.ClassVar")
+    return str(annot).startswith(("typing.ClassVar", "t.ClassVar", "ClassVar"))
```
hynek: I think we should amend our tests to cover the new cases?

ambv: Yes.

hynek:
🏓

ambv:
(updated)

hynek:
Thanks everyone! At least we can feel bad about that hack *together*. ;)

PR #369, opened 2018-04-13
Add a narrative chapter on object hashing

hynek:
It would be great if our resident hashing expert @njsmith could have a look too. :)

### DIFF
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -1,0 +1,46 @@
+Object Hashing
+==============
+
+.. warning::
+
+   The overarching theme is to never set the ``@attr.s(hash=X)`` parameter yourself.
+   Leave it at ``None`` which means that ``attrs`` will do the right thing for you, depending on the other parameters.
+
+   Setting ``hash`` yourself can have unexpected consequences so we recommend to tinker with it only if you know exactly what you're doing.
+
+Under certain circumstances, it's necessary for objects to be *hashable*.
+For example if you want to put them into a :class:`set` or if you want to use them as keys in a :class:`dict`.
+
+The *hash* of an object is an integer that represents the contents of an object.
+It can be obtained by calling :func:`hash` on an object and is implemented by writing a ``__hash__`` method for your class.
+
+``attrs`` will happily write a ``__hash__`` method you [#fn1]_, however it will *not* do so by default.
+Because according to the definition_ from the official Python docs, the returned hash has to fullfill two constraints:
+
+#. The hash of an object *must not* change.
+   This is the reason why mutable structures like lists, dictionaries, or sets aren't hashable while immutable ones like tuples or frozensets are.
+
+   If you create a class with ``@attr.s(frozen=True)`` this is fullfilled by definition therefore ``attrs`` will write a ``__hash__`` function for you automatically.
+   You can also force it to write one with ``hash=True`` but then it's *your* responsibility to make sure that the object is not mutated.
+
+#. Two objects that are equal, must have the same hash. This means that if ``x == y``, it *must* follow that ``hash(x) == hash(y)``.
+
+   By default, Python classes are compared *and* hashed by their :func:`id`.
+   That means that every instance of a class has a different hash, no matter what attributes it carries.
+
+   It follows that the moment you (or ``attrs``) change the way equality is handled by implementing ``__eq__`` which is based on attribute values, this constraint is broken.
+   For that reason Python 3 will make a class that has customized equality unhashable.
+   Python 2 on the other hand will happily let you shoot your foot off.
+   Unfortunately ``attrs`` currently mimics Python 2's behavior for backward compatibility reasons if you set ``hash=False``.
+
+   The *correct way* to achieve hashing by id is to set ``@attr.s(cmp=False)``.
+   Setting ``@attr.s(hash=False)`` (that implies ``cmp=True``) is almost certainly a *bug*.
+
+
+For a more thorough explanation of this topic, please refer to this blog post: `Python Hashes and Equality`_.
+
+
+.. [#fn1] The hash is computed by hashing a tuple that consists of an unique id for the class plus all attribute values.
+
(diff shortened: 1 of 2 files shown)


docs/hashing.rst
```diff
@@ -0,0 +1,46 @@
...
+For example if you want to put them into a :class:`set` or if you want to use them as keys in a :class:`dict`.
+
+The *hash* of an object is an integer that represents the contents of an object.
+It can be obtained by calling :func:`hash` on an object and is implemented by writing a ``__hash__`` method for your class.
+
+``attrs`` will happily write a ``__hash__`` method you [#fn1]_, however it will *not* do so by default.
+Because according to the definition_ from the official Python docs, the returned hash has to fullfill two constraints:
+
+#. The hash of an object *must not* change.
+   This is the reason why mutable structures like lists, dictionaries, or sets aren't hashable while immutable ones like tuples or frozensets are.
```
njsmith: The way this is ordered seems confusing for those who don't already know about hashes... if the first thing you say about a hash is that it can't change, then it's not clear why it shouldn't be `def __hash__(self): return 0` or what mutability has to do with it.

Maybe: (1) If objects are `==`, then their hashes MUST be the same. (2) If objects are not `==`, then their hashes SHOULD be different – technically this isn't a requirement, but it has to be true most of the time or sets and dicts will become horribly inefficient. (3) The hash MUST NOT change.

So (2) means that if you mutate an object in a way that changes `==` comparisons, then the hash has to change, but (3) says it can't change => that's why only immutable objects are hashable.

docs/index.rst
```diff
@@ -36,6 +36,8 @@ Day-to-Day Usage
 ================
 
 - Once you're comfortable with the concepts, our :doc:`api` contains all information you need to use ``attrs`` to its fullest.
+- If you want to put objects into sets or use them as keys in dictionaries, they have to be hashable.
+  Unfortunately the topic is more complex than it seems but :doc:`hashing` will give you a primer on what to look out for.
```
njsmith: Suggestion: "... they have to be hashable. The simplest way to do that is to use `frozen=True`. Unfortunately ..."

docs/hashing.rst
```diff
@@ -0,0 +1,46 @@
...
+   It follows that the moment you (or ``attrs``) change the way equality is handled by implementing ``__eq__`` which is based on attribute values, this constraint is broken.
+   For that reason Python 3 will make a class that has customized equality unhashable.
+   Python 2 on the other hand will happily let you shoot your foot off.
+   Unfortunately ``attrs`` currently mimics Python 2's behavior for backward compatibility reasons if you set ``hash=False``.
+
+   The *correct way* to achieve hashing by id is to set ``@attr.s(cmp=False)``.
+   Setting ``@attr.s(hash=False)`` (that implies ``cmp=True``) is almost certainly a *bug*.
+
+
+For a more thorough explanation of this topic, please refer to this blog post: `Python Hashes and Equality`_.
```
njsmith: Maybe put a little "but what do I do?" faq here for those who didn't follow all the details? Like:

If you want to make objects hashable by value: use `@attr.s(frozen=True)`

If you want hashing and comparison by object identity: use `@attr.s(cmp=False)`

If you want bugs: use `@attr.s(hash=False, cmp=True)`

hynek:
I *think* I’ve addressed your comments?

docs/hashing.rst
```diff
@@ -15,15 +18,10 @@ The *hash* of an object is an integer that represents the contents of an object.
 It can be obtained by calling :func:`hash` on an object and is implemented by writing a ``__hash__`` method for your class.
 
 ``attrs`` will happily write a ``__hash__`` method you [#fn1]_, however it will *not* do so by default.
-Because according to the definition_ from the official Python docs, the returned hash has to fullfill two constraints:
-
-#. The hash of an object *must not* change.
-   This is the reason why mutable structures like lists, dictionaries, or sets aren't hashable while immutable ones like tuples or frozensets are.
+Because according to the definition_ from the official Python docs, the returned hash has to fullfill certrain constraints:
```
njsmith: spelling: certrain → certain

docs/hashing.rst
```diff
@@ -0,0 +1,56 @@
...
+   Setting ``hash`` yourself can have unexpected consequences so we recommend to tinker with it only if you know exactly what you're doing.
+
+Under certain circumstances, it's necessary for objects to be *hashable*.
+For example if you want to put them into a :class:`set` or if you want to use them as keys in a :class:`dict`.
+
+The *hash* of an object is an integer that represents the contents of an object.
+It can be obtained by calling :func:`hash` on an object and is implemented by writing a ``__hash__`` method for your class.
+
+``attrs`` will happily write a ``__hash__`` method you [#fn1]_, however it will *not* do so by default.
+Because according to the definition_ from the official Python docs, the returned hash has to fullfill certrain constraints:
```
Tinche: fullfill -> fulfill

docs/hashing.rst
```diff
@@ -0,0 +1,56 @@
...
+   For that reason Python 3 will make a class that has customized equality unhashable.
+   Python 2 on the other hand will happily let you shoot your foot off.
+   Unfortunately ``attrs`` currently mimics Python 2's behavior for backward compatibility reasons if you set ``hash=False``.
+
+   The *correct way* to achieve hashing by id is to set ``@attr.s(cmp=False)``.
+   Setting ``@attr.s(hash=False)`` (that implies ``cmp=True``) is almost certainly a *bug*.
+
+#. If two object are not equal, their hash **should** be different.
+
+   While this isn't a requirement from a standpoint of correctness, sets and dicts become less effective if there are a lot of identical hashes.
```
Tinche: This isn't always possible technically. Reading up on __hash__, in practice Python limits hash values to 64-bit integer ranges. And imagine a class with even a single int, there are more different values of this int than there are distinct values in a 64-bit integer, so collisions will happen. But the idea is to minimize them.

PR #371, opened 2018-04-21
Add glossary & make slotted classes official

### LINKED ISSUE #338: Class nomenclature
It’s time for everyone’s favorite pastime: bikeshedding!

I’ve noticed that we’re a bit inconsistent with how we call __dict__ classes and __slots__ classes throughout the documentation.

I would like to unify it and add a glossary.

I’m pretty happy with “dict classes” but I’m not sure whether it should be “slot classes” or “slots classes”? The former sounds a bit better while the latter is technically more correct I guess?

Opinions especially from native Python and native English speakers? ;)

hynek:
Fixes #338

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -84,7 +84,7 @@ Changes
   The singleton empty ``dict`` is still enforced.
 
   `#280 <https://github.com/python-attrs/attrs/issues/280>`_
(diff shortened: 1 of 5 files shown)


hynek:
Final PR before 18.1! We’re sooo close folks!

PR #370, opened 2018-04-20
Add narrative chapter about initialization

hynek:
So it’s time to depend on examples even less.

Although the sections on validators and converters are copied mostly verbatim from there. I’m not sure how to cull examples in a meaningful way? It could be a little briefer and just refer to the new chapter.

But it finally gives a good place to explain intricacies of initialization.

Feedback welcome!

Fixes #289 #314 #264 #255 #263 #171 #107 #30

Issue #301, opened 2017-11-28
Support for automatic runtime type-checking


anthrotype:
This has been proposed and discussed in https://github.com/python-attrs/attrs/issues/215, as a possible use case for the newly added `type`  argument to `attr.ib()` #239 

quoting @hynek https://github.com/python-attrs/attrs/issues/215#issuecomment-347529479

> a good first step would be to add generic class-wide validators (@attr.s(validator=whatever)) and then make type checking a special case of it, possibly with some syntactic sugar.

aldanor:
Ok, as a continuation of previous discussion in #215... :)

> @glyph I was trying to find https://github.com/aldanor/typo today but instead I ran across https://github.com/RussBaz/enforce which looks like it may be a fairly full-featured version of the same thing.

> @chadrik There is also https://github.com/Stewori/pytypes May the best project win.

Just to clarify, the sole reason I've started one of my own was that all existing solutions (those including enforce and pytypes) were (a) slow and (b) wrong -- although both are very good attempts and good inspiration. That being said, my version is not fully type-correct either when it comes to sum types (see examples below), but 'less wrong' if I may; on the bright side, it's fast. I haven't spent any time on finishing it due to lack of motivation and time lately, but it could be done, maybe with some help. If anyone knows any other comparable projects - shout away.

TL;DR: it's hard to write a runtime type checker that's both fast and correct, especially if it aims to handle both typevars and sum types; although not impossible (I'm not sure one already exists at this moment however). Details below.

---

```python
# pip install git+https://github.com/RussBaz/enforce.git
import enforce
# pip install git+https://github.com/aldanor/typo.git  (3.5 only; needs a few fixes)
import typo
# pip install git+https://github.com/Stewori/pytypes.git
import pytypes  
```

---

Simple example:

```python
def simple(x: int, y: str): pass

simple_pytypes = pytypes.typechecked(simple)
simple_enforce = enforce.runtime_validation(simple)
simple_typo = typo.type_check(simple)

args = 1, 'foo'
%timeit -r7 -n1000 simple_pytypes(*args)
%timeit -r7 -n1000 simple_enforce(*args)
%timeit -r7 -n1000 simple_typo(*args)
```

```
314 µs ± 23 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
155 µs ± 2.14 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
[...]

PR #372, opened 2018-04-22
Add instance-level validators

hynek:
This is a prerequisite for #301 

I have not update examples because this belongs into #370 and I will update the respective docs once this is merged.

### DIFF
--- a/changelog.d/372.change.rst
+++ b/changelog.d/372.change.rst
@@ -1,0 +1,1 @@
+``@attr.s`` now also has a *validator* argument that takes a callable and is run with the fully initialized instance.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -714,7 +714,8 @@ def attrs(maybe_cls=None, these=None, repr_ns=None,
 
         .. _`PEP 526`: https://www.python.org/dev/peps/pep-0526/
     :param callable validator: Run after a new instance is fully initialized
-        with the instance as its only argument.
+        with the instance as its only argument.  *validator* runs *after*
+        the attribute validators.
 
     .. versionadded:: 16.0.0 *slots*
     .. versionadded:: 16.1.0 *frozen*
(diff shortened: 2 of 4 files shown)


Tinche:
First thought:

why aren't we supporting this:

```
@attr.s
class C:
    a = attr.ib()
    b = attr.ib()

    @attr.validator
    def _validate(self):
        if self.a > self.b:
            raise ValueError()
```

Then I went and read the linked issue.

I think runtime type validation is a cool feature that we should definitely support. However the interface of the proposed solution is ugly to me. If you are an advanced attrs user, and you want to extend attrs, you don't actually need this functionality - wrap attr.s or place a class decorator underneath that would wrap an existing __attrs_post_init__ or add a new one if none found, and do your logic in there.

hynek:
Hm that seems a bit complicated to me. The idea is that you can write a reusable validator that looks at  `__annotations__` and uses one of the many available type checkers. That sounds like something that should be possible without wrapping `@attr.s`?

Tinche:
First of all, I'm not going to stop you merging this in, just thinking out loud. :)

We already have a mechanism for executing code after init is done, it's the post_init.

What if you want an additional instance validator after or before typechecking?

Also note:

```
@attr.s(validator=typecheck)
class C:
    a: int = attr.ib()
```

In this scenario, typecheck will have to be a generic function, right? What I mean by this is, all instances will share a single function, and this function will probably call attr.fields and iterate over them. If I was doing this, I would want to eval a specialized function to avoid the overhead. And in that case, you need to either wrap attr.s or do your own class decorator. Since I appreciate code aethethics, I would do (hypothetically):

```
@typechecked.attr.s
class C:
    a: int = attr.ib()
```

hynek:
Well, thinking is what I’m (not – but I totally would) paying you for, so please keep doing that. ;)

I guess it’s true that it might get too slow to do it dynamically. 🤔

Maybe you should hopp over to #301 and make some suggestions (after reviewing my PRs natch), I’m gonna close this for now.

PR #383, opened 2018-05-23
Improve error for in_(a_string) with a non-string value

### LINKED ISSUE #382: Poor error message/type for validators.in_ with a string
```python
import attr

@attr.s
class C:
   s = attr.ib(validator=attr.validators.in_('abc'))

C(s=1)  # TypeError: 'in <string>' requires string as left operand, not int
```

`__contains__` behaves a little weirdly for strings, but I still think this error could be improved.

Obvious options would be to catch exceptions and re-raise something with a clearer error message, and/or to deprecate use of strings as the collection here.

Found in attrs=18.1.0 while working on HypothesisWorks/hypothesis#954.

Zac-HD:
- [x] Is my first attrs PR :heart_eyes:

Closes #382.

### DIFF
--- a/changelog.d/383.change.rst
+++ b/changelog.d/383.change.rst
@@ -1,2 +1,2 @@
 ``attr.validators.in_()`` now raises a ``ValueError`` with a useful message even if the options are a string and the value is not a string.
-Previously, this raised a ``TypeError`` (matching e.g. ``1 in "abc"``).
+This previously raised a ``TypeError``, like ``1 in "abc"``.
--- a/changelog.d/383.change.rst
+++ b/changelog.d/383.change.rst
@@ -1,3 +1,2 @@
-``in_()`` validators now raise a ValueError with a useful message
-even if the options are a string and the value is not a string.
-By contrast, e.g. ``1 in "abc"`` raises an unhelpful TypeError.
+``attr.validators.in_()`` now raises a ``ValueError`` with a useful message even if the options are a string and the value is not a string.
+Previously, this raised a ``TypeError`` (matching e.g. ``1 in "abc"``).
--- a/changelog.d/383.change.rst
+++ b/changelog.d/383.change.rst
@@ -1,0 +1,3 @@
+``in_()`` validators now raise a ValueError with a useful message
+even if the options are a string and the value is not a string.
+By contrast, e.g. ``1 in "abc"`` raises an unhelpful TypeError.
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -137,7 +137,7 @@ class _InValidator(object):
     def __call__(self, inst, attr, value):
         try:
             in_options = value in self.options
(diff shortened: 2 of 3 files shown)


src/attr/validators.py
```diff
@@ -135,7 +135,12 @@ class _InValidator(object):
     options = attrib()
 
     def __call__(self, inst, attr, value):
-        if value not in self.options:
+        try:
+            in_options = value in self.options
+        except Exception as e:  # e.g. `1 in "abc"`
```
wsanchez: I think we want `TypeError`, not `Exception` here.  Other exceptions don't necessarily imply that `value` isn't one of the items in `self.options`.

tests/test_validators.py
```diff
@@ -243,6 +243,18 @@ def test_fail(self):
             "'test' must be in [1, 2, 3] (got None)",
         ) == e.value.args
 
+    def test_fail_with_string(self):
+        """
+        Same error for string options as for list options, not TypeError.
```
wsanchez: Don't like "Same error" referencing… probably the prior test.  I'd change to:

```
Raise ValueError if the value is outside our options when the options are specified as a string and the value is not a string.
```

Zac-HD:
@wsanchez - done :smile:

changelog.d/383.change.rst
```diff
@@ -0,0 +1,3 @@
+``in_()`` validators now raise a ValueError with a useful message
```
hynek: It would be faster to fix it myself, but for the next time:

- We use [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).
- We use fully qualified names for functions (e.g. `attr.validators.in_()`) which simplifies the sentences – “`attr.validators.in_()` now raises…”.
- “raises an unhelpful TypeError” makes it sound like it’s present behavior which it’s not. :)
- Please wrap symbols like ValueError or TypeError in double ticks.

Yeah, sorry I care about changelogs. 😇

Zac-HD: No worries, I care about them too!  Fixed ❤️

changelog.d/383.change.rst
```diff
@@ -0,0 +1,2 @@
+``attr.validators.in_()`` now raises a ``ValueError`` with a useful message even if the options are a string and the value is not a string.
+Previously, this raised a ``TypeError`` (matching e.g. ``1 in "abc"``).
```
hynek: I cannot quite parse the “matching” part – any chance to rephrase that? I genuinely don’t know what that means and it almost looks like a typo to me. :)

Zac-HD: How about "(just like `1 in "abc"`)" ?

hynek: ah yeah maybe drop the `just` and we’re golden

hynek:
Thanks!

Issue #146, opened 2017-02-12
Generalizing ad-hoc attribute callbacks


hynek:
While 37d38c4 is pretty cool, when working on it it was very obvious to me, that it should be united with convert.  After @Tinche pointed it out too, there’s no way back.

IOW, something like this:

```
@attr.s
class C:
    x = attr.ib()

   @x.callback_but_a_better_name:
   def take_number_make_it_a_hex_string(self, attribute, value):
       if not isinstance(value, int):
           raise TypeError(f"{attribute.name} must be an int")
       return hex(value)
```

Any other ideas?

The downside that it breaks the simplicity of the current implementation and we’d have to treat those  callbacks in a special way (instead of just like any other validator).

***

Thinking of it, it might make sense to make it a different feature altogether (i.e. having a lightweight validator and something more substantial) because requiring to return the new value is a source for errors…in any case it mustn’t be called validator because the semantics would be different.

I’d tend to say that people can just do a `self.x = hex(value)` but that breaks frozen classes and we have that problem already in #120.

hynek:
Maybe `@x.pipe_through`?

Tinche:
My line of thinking was basically that:
* converters are a superset of validators (they can do everything validators do, and more)
* converters have a less powerful interface than validators, but it's more convenient

Thus, if we can come up with a more sophisticated interface for converters, we could unify converters and validators and get rid of validators. I think the ``@x.validator`` syntax is nifty but of limited use (I almost always refactor my converters/validators out and reuse them), we should definitely support it, but this wasn't the primary motivation for my thinking; this idea is more to simplify the core.

So here is my idea, which may be different than @hynek's in scope.

We define a new concept. I'll call it hook, although I'm not really happy with that name, but let's go with it for now.

A simple class, to illustrate the concept.
```
def positive_int(inst, att, value):
    if value < 1:
        raise ValueError

@attr.s
class A:
    a = attr.ib(convert=int, validator=positive_int)
```
(I think it's telling I had to open attrs docs to look up the exact validator signature to write this.)

This results in the following `__init__`, basically:
```
class A:
    def __init__(self, a):
        self.a = int(a)
        positive_int(self, A.a, a)
```
This can be simplified by having a hook that does both.
```
def ensure_positive_int(inst, att, value):
    v = int(value)
    if v < 1:
        raise ValueError
    return v

@attr.s
class A:
    a = attr.ib(hook=ensure_positive_int)

# Resulting in:

class A:
[...]

PR #404, opened 2018-07-02
Add decorator option for converters

altendky:
Issue #240

Draft for:
- [ ] Maybe update the mypy plugin
  - https://github.com/python/mypy/blob/13ae58ffe8bedb7da9f4c657297f0d61e681d671/mypy/plugins/attrs.py
  - `tests/typing_example.py:126: error: Unsupported converter, only named functions and types are currently supported`
- [x] Test against the new `setattr` feature
- [x] ~~Consider eliminating the `Converter` class~~
- [x] Test the passed `Attribute`
- [ ] Look over again for any more doc changes
- [ ] Resolve the circular import due to `attr.setters.converter()` needing `attr._make.Converter` (but `attr._make` imports `attr.setters` so...)


hynek:
I see it’s WIP, just a bunch of random notes to save you some time:

- are you sure we need `takes_self` in converters? I don’t think so, we don’t have anything similar for the converter option either and you hard-code it too. The diff gets a lot simpler if you “just” use the decorator to set the option and be done with it.
- [test docstrings](https://jml.io/pages/test-docstrings.html)
- newsfragment :)

altendky:
That's why I put it up here sooner than later, thanks.

I made `Converter` to be like `Factory` is for `default=`.  Though the calling side didn't end up so similar I don't think, I need to review that to see about removing the `Converter.__call__` part that `Factory` doesn't have.  I'll see how I can simplify that as I learn `Factory`'s implementation better but it is not presently clear to me how the diff would be much different without `Converter.takes_self`.  It seems like just assuming that to be true would only alleviate half of a boolean check/assignment.

```python3
converter_self = (
    isinstance(a.converter, Converter) and a.converter.takes_self
)
```
would roughly become:
```python3
converter_self = isinstance(a.converter, Converter)
```

altendky:
@hynek, I figured it would make sense to clarify the logic before writing up docs etc.  I didn't quite understand how your suggestion would make the diff much simpler.  Do you have time to offer a bit more explanation?

I looked at making it so that internally `converter` is always a `Converter` but that seems questionable because `attr.fields(MyClass).my_attribute.converter` would not provide what the developer expects when have `attr.ib(converter=my_function)`.  With this change, `TestConverter.test_deprecated_convert` failed for this reason.  Otherwise, this did make for a significant simplification of the diff.  I wouldn't expect we would want to hold up progress much based on a deprecated parameter name, but I do expect that aside from that we want to get the same object out of `attr.fields(MyClass).my_attribute.converter` that we put into `converter=`.

Then again, perhaps I have just misrepresented my needs.  I want both the decorator syntax for converter and also the `takes_self` aspect similar to `default=Factory()`.  I see how the former does not imply the latter and I'm sorry if that caused confusion.  The issue that brought me to this topic this week is https://github.com/ebroecker/canmatrix/commit/a8893523319c4ede32e59ee19bf8420095945e41#r29555978.

altendky:
I'm not sure if storing a `Converter()` regardless in `names_for_globals` is a tidy way to increase consistency in this code while retaining the outward appearance of whatever the caller passed to `convert=`...  or if it's really ugly.

hynek:
So um I’m confused by your use case, it looks like you really just want a default? What do you need `self` in the converter for? What I meant is that you should “just” set the converter attribute of CountingAttr and be done with it, which would allow you to touch a lot less code.

So could you elaborate *why* you think you need `self` in a converter? Because semantically, it doesn’t make any sense, but I guess I’m missing something?

P.S.  The rst in the news fragment is broken. :)

altendky:
One attribute is `float_factory` which defaults to `float` but may also be `decimal.Decimal` or such. Whatever is passed in for the `offset` attribute should be converted using the `float_factory` attribute. Does that explain the desire for `self` in the converter?  It seems possibly a little dirty, but I haven't decided it's bad yet.

As to just making every `_CountingAttr` just have a `Converter`?  It seems it would be nice if an attribute defined as `class C: a = attr.ib(converter=f)` would return `f` from `attr.fields(C).a.converter`.  Hopefully I'm remembering my functions etc correctly at the moment.

altendky:
https://travis-ci.org/python-attrs/attrs/jobs/401284405#L510
```
would reformat /home/travis/build/python-attrs/attrs/tests/test_make.py
```

I don't get this locally and it's not telling me what line etc...  but apparently I guessed correctly.

altendky:
040f2b978147837f8f0f0b0880012da2f43904a6 adds a test for my desired behavior of `attr.fields(C).a.converter` being the object passed via `attr.ib(converter=f)`.  I figure that as long as my code is targeting that, it should have a test.  If attrs doesn't care about that functionality then code and tests will both get updated.

hynek:
- a working `attr.fields(C).a.converter` is definitely a must
- I can sympathize with needing self for having a factory since I use that pattern in `structlog` although I’m still not a fan of the complexity it introduces. 🤔 I’d like a weigh in by @Tinche, @wsanchez, and/or @glyph if there’s something missing from the picture.

glyph:
I honestly don't understand the use-case at all here.

Access to partially-initialized `self` was a nightmare in the cases where I used it even with validators, for converters this seems worse.  Every time I think I want a converter that accesses `self`, what I actually want is a `@classmethod` constructor that does the I/O before attempting to construct an instance.

hynek:
I think what Kyle wants it this:

```python
@attr.s
class C:
    type = attr.ib(default=float)
    value = attr.ib(default=0.0). # NB: you can’t even have mandatory attribs with this approach

   @value.converter
    def _(self, val):
        return type(val)
```

And I’m gonna side with @glyph here: passing around half-initialized `self`s is something I rather regret so I would strongly prefer to not add more of them. I also feel like converters per se are being overused – at least I noticed it for myself. It really shouldn’t do anything more than int -> float or similar. Once you want logic, you should put that into a `@classmethod` with a meaningful name.

This is really complicated. 🤔

glyph:
@altendky So to be as specific as possible, the question is: why don't you do this instead?

```python
@attr.s
class C:
    type = attr.ib()
    value = attr.ib()

    @classmethod
    def new(cls, type, value):
        return cls(type, type(value))
```

altendky:
I don't have a problem doing 'something else' like suggested.  I implemented the way it is because there was a ticket open for the decorator and because I figured symmetry with the `default` decorator taking self made sense.  Since `default` taking self is viewed as an error, I understand not repeating it with `converter`.  (I'll have to think about it more before I am sure if I agree it is an error :] )  

So, no `self`...  Do we even want a decorator?  Aside from my sample use case which is easy enough to find an alternative to (such as in the last comment)

Overall I thought the relationship between an attribute and it's default or converter code seemed nice. Better than a totally arbitrary `__attrs_post_init__` or `@classmethod` but it sounds like I should reconsider that perspective.

hynek:
Yeah so to be clear, we don't blame you for tackling an open issue – we’re still figuring this stuff out in-flight ourselves and I’m very much heartbroken that you probably wasted your time. However it kinda prompted a good discussion and is a good opportunity for some introspection.

I’m gonna share my experience with the related issues over the years:

- `converter` is almost never what you want. It makes sense for the original intent of ensuring that a number is a float or int or vice versa. I also like to use it to convert strings into URL objects but even that is kinda sketchy. The problem is that it generally becomes less clear how to instantiate the object and has all kind of weirdnesses around itself.
- Passing around half-initialized objects was a bad idea and it already fell on @glyph's foot once, so I don’t want to add any more of that. It *sort of* makes sense for validators because they tend to be read only and it allows to validate items in relation to each other but I also honestly have to admit that I don't use validators myself a lot. For typing I now have mypy and for everything more complicated it's usually a better idea to take the `@classmethod` route.
- Obviously a decorator makes no sense if you don't want to pass around `self`. You can as well write a function; maybe scope it inside the class body.
- `__attrs_post_init__` *can* be quite useful but probably not for what most people think. So far I’ve identified two useful things: 1. initialising things like metrics objects or loggers that live on a class and depend in some way on data on the class: e.g. labels.  2. calling `super()` if you have to subclass and need to call it. The second use case is really a fringe issue that shouldn't affect to many people but those who need it, *really* need it. Although it would be nice to know how many people actually use it and how many decided that it’s not worth it in the end.

***

So with a heavy heart, I propose to close this and #240 as won't fix and instead add more documentation to http://www.attrs.org/en/latest/init.html – basically summing up what glyph and  I wrote above.

I welcome your comments first though!

altendky:
My comments about expecting this (concept) to just be accepted weren't meant to express frustration.  Rather just that I hadn't put a lot of thought into the concept making sense or not because it seemed to fit into the mold of other pieces etc.  :]  Closing is fine.

For the discussion though, I still don't see the issue with passing around a partially initialized object to things that are clearly part of initialization.  It's just breaking up an __init__ into pieces and just like __init__, order matters.  Also like __init__, you get an attribute error if you access another attribute before you define it.  Could people get confused about this? Sure.  I've gotten confused about much more obvious errors (and maybe this one as well the first time).  I guess I see attrs as providing a framework for describing these things (default, converter, etc) rather than totally freeform code like a class method.  I am probably also biased because I have built my own things on top of this structure.  Serialization, automatic on change signals for PyQt, a PyQt model from attrs classes which uses the converter method for user input validation.  attrs reduces boilerplate...  and also offers this framework for easily introspecting attributes to build large frameworks.

As to inheriting?  I do that some, I should do it less, but still some.  I try not to let Qt's inheritance leak too far but sometimes a bit seems appropriate.

jimbo1qaz:
Why was this rejected?

I load config files from YAML, where user input is generally in the form of `str`. I would like to use a "converter decorator" to convert `Class.field` from `str` to `EnumSubclass[field]`.

But when I pass a Class object to `attr.evolve`, `field` is already `EnumSubclass`, so I must skip `EnumSubclass[field]` to prevent crashing. (I don't need access to half-initialized `self`.)

I feel defining an external converter outside the class is ~~unnecessarily bloated...~~ hmm I could just add a `EnumSubclass.from_str(Union[str, EnumSubclass])` and use it as a `attr.s(converter=)`.

hynek:
It wasn't definitely rejected, since it's still open. :) The problem is that it's dawning on me, that converters weren't the best idea in the first place. They are _nice_ if you just need to fix up some simple value, but they get weird really fast, especially in concert with defaults. A classmethod is almost always the better solution.

***

I _think_ – and I'd love input from others on this too – that the better way would be to focus on figuring out #146?

altendky:
What if attrs created the `@classmethod`?  As another option, I joked recently about `attr.with_converter(TheClass)(x, y)` as well.  Or the opposite such as `attr.raw_init(TheClass)`.

The basic point is that you should be able to create the class without all the conversion and validation, even if you do usually want those activities to occur?

I haven't thought about it much but the hybrid attribute 'handler' seems reasonable. To toss another point into the mix, having the handlers (or converter/validator) applied even after init is sometimes useful.  I've got a probably messy thing that creates properties for that.

glyph:
> What if attrs created the `@classmethod`?

I wouldn't say that this is *necessarily* a bad idea, but in my mind there are two cases:

1. `attrs` initializes some attributes directly, and the constructor has no logic beyond that (with the proviso that converters can massage a few things from strings by parsing them or somesuch)
2. there's some complex, interdependent logic for initializing the attributes from some inputs

for the first case, converters are already adequate; for the second, I don't know if there's really much attrs can do to help.  When writing `__init__`s all day long I was constantly thinking "this is tedious, I wish it could be automated!".  Now that I write `@classmethod`s on `attrs` classes all day long, in the simple case, I just… don't write the classmethod at all and don't worry about it.

I'd be interested to hear if anyone else has a *specific* sense that there's a tedious or automatable portion of writing classmethod constructors.

glyph:
> hmm I could just add a `EnumSubclass.from_str(Union[str, EnumSubclass])` and use it as a `attr.s(converter=)`

We do this in several places in our codebase and it comes out reasonably nicely.

Maybe there's room for a generic `if_not_already(Type[O], Callable[[T], O]) -> Callable[[Union[T, input], O]` to be somewhere in attrs

altendky:
@glyph, my point was that presently you can't bypass the converters (correct?).  Having attrs provide a way to bypass those steps or putting them in a class method rather than init would make it possible.

glyph:
> my point was that presently you can't bypass the converters (correct?).

As far as I know, yes.

> Having attrs provide a way to bypass those steps or putting them in a class method rather than init would make it possible.

Hm.  This strikes as squarely within the domain of metaprogramming, ("if you want this, you probably want to call `__new__`, not `__init__`").  The given concrete use-case so far (serialization) is definitely just such a thing; you don't want to have to explicitly opt in to having serialization skip initialization steps with optional methods, you want a thing that is going to call `attr.fields` and have some very defined semantics with respect to what it does with the information it finds there, which either does or doesn't respect the constructor.

hynek:
> > my point was that presently you can't bypass the converters (correct?).
> 
> As far as I know, yes.

Which is one of the main reason why I think this approach is flawed and shouldn't be pursuit any further.

> What if attrs created the `@classmethod`?

I don't think this is useful TBH. I understand your desire for a raw `__init__` (which is what I usually do nowadays) but I don't think code is gonna get any easier if we try to write some meta language to define factory methods. I think that'd just introduce a lot of complexity, make it less readable, and still not fix all use cases.

Sometimes it's OK to write some code. :)

***

The question is whether or how to simplify @glyph's first use case: give some power to `__init__` but don't make it too awkward. And I truly believe that #146 is the better way and the solution should put emphasis on code re-use.

I'm very open to suggestions on how to achieve that.

meereeum:
I would like to echo the request for more fully-featured converters (whether via this or #146) - especially the ability to take `self` as an arg. (while acknowledging the tradeoff involved in adding another way to interact with a half-baked `self`..)

Consider this use-case: the initial `val` is some `str`. The converter is like a case statement -- has a dictionary of keys (of which `val` must be one) and values that are functions of previously passed `attrs` (hence need for `self`).

I could do this in `__attrs_post_init__` but seems clumsy to separate the `attr.ib` definition from its actual effect.

Not sure if this issue has fallen by the wayside because of lack of contributors or perceived lack of interest, but just wanted to register definite *interest* !

I have found `attrs` in general to be very useful and concise, but this would make it feel more natural rather than like I'm sometimes fighting against the varying capabilities of `factory` vs `validator` vs `converter`...

BrendanSimon:
> I would like to echo the request for more fully-featured converters (whether via this or #146) - especially the ability to take `self` as an arg. (while acknowledging the tradeoff involved in adding another way to interact with a half-baked `self`..)
> 
> Consider this use-case: the initial `val` is some `str`. The converter is like a case statement -- has a dictionary of keys (of which `val` must be one) and values that are functions of previously passed `attrs` (hence need for `self`).
> 
> I could do this in `__attrs_post_init__` but seems clumsy to separate the `attr.ib` definition from its actual effect.

My use case is to do some processing on the input value.  e.g. trim the size of a string to a maximum length.  The length is determined by another attribute in the class (`self.max_length`).  I can't do this with a global function as self is not passed to it (with a coverter callable).

Using `__attrs_post_init__` can be done but only for instanace initialisation.  It wont work if you have a setter for the attribute.
e.g. `attr.ib( ..., on_setattr=attr.setters.convert ).

Actually that does work for calling the converter, but I need access to other attributes of the instance.

That's where an @setter.converter would be handy (well as long as the API passes a reference to the instance, like the validator decorator does).

altendky:
I don't know if it's really better but at this point I pretty much don't use converters or validators (or `__attrs_post_init__`).  If I want to do anything beyond accepting a value and putting it in the attribute I write an `@classmethod` builder.  Or a regular function because I ran into issues with proper type hinting and `cls` and just trying to allow inheritance without even actually leveraging it.

I kind of feel like there's a way to get a 'just construct the class' thing and a 'validate and convert when constructing' thing built by attrs (I made some kinda crazy examples before concluding I shouldn't).  The 'partially constructed `self`' concerns never completely settled with me.  They make sense but at some point there is code that executes in an order and which builds the pieces.  Though I guess tying that order to `__init__` parameter order which ties it to defaults-first and the order from `attrs.to_dict()` etc etc might just be too much.  Maybe that justifies simply not passing around 'partially constructed `self`', maybe not.

altendky:
I suppose I could also say that if interest from those with authority is gotten I could take some time to catch up this PR etc.  IIRC, I stopped because it wasn't obvious that it was going somewhere that would be wanted.

jgerityneurala:
I don't have any authority, but being able to spell this with a decorator would be nice, imo.

nyanpasu64:
> I don't know if it's really better but at this point I pretty much don't use converters or validators (or `__attrs_post_init__`). If I want to do anything beyond accepting a value and putting it in the attribute I write an `@classmethod` builder. Or a regular function because I ran into issues with proper type hinting and `cls` and just trying to allow inheritance without even actually leveraging it.

I used to use `__attrs_post_init__` and/or converters, to validate input read from config files, and convert string literals or integers into domain-type enums.

Coming back to this issue over a year later, my perspective has changed somewhat. I've spent time working with Rust which influenced my view somewhat, and haven't worked with file deserialization in a while. For cases like "parsing argv" (instead of parsing complex files), usually I write `@staticmethod` builders instead of `@classmethod`, processing inputs before wrapping the values in a dataclass.

When working with file deserialization, I don't know if it's worth having separate "raw input" structs (user-supplied documents/configs or source-code literals, usually simpler types) and "parsed input" types (with validated fields and domain-enum types).

Rust only allows structs to define construction conversions through factory methods instead of initializer hooks, but factory methods are a bit crippled there since factory functions can't accept keyword arguments (only struct literals can). I think Rust literals in source code allows `.into()` with inferred return type, whereas Python doesn't.

altendky:
@nyanpasu64, why `@staticmethod` over `@classmethod`?  I presume the method is creating an instance of the class?  Anyways, for [de]serialization maybe look at [marshmallow](https://marshmallow.readthedocs.io/en/stable/api_reference.html), [cattrs](https://cattrs.readthedocs.io/en/latest/), or [desert](https://desert.readthedocs.io/en/stable/)?

nyanpasu64:
What's the benefit of typing `cls(args...)` over `SomeClass(args...)`? The ability to use a subclass, as long as it doesn't have more mandatory fields than the parent?

Rust doesn't have subclassing, C++ does but you can't really aggregate-initialize a subclass. So I haven't really explored subclassing data/struct-like objects in the past year or so, only encapsulated objects.

altendky:
The first thing for me is `@classmethod` says 'this method is related to the class rather than just arbitrarily tossed in here even though it could be a regular function'.  Inside the method `cls` says 'i am going to create an instance of this class' rather than just some class whose name you hopefully typed correctly.  And yes, subclassing.  Though as mentioned I ran into trouble getting type hints playing nicely. (and I mostly don't inherit)

On the flip side, what's the benefit of `@staticmethod` here?

Not that this is an end-all to the discussion but...  https://mail.python.org/pipermail/python-ideas/2016-July/041189.html

hynek:
So as someone with authority let me meditate on this a bit.

I feel like everybody who asks for more complex converters, eventually realizes that they are not enough for what they want to do. They usually end up writing factory methods using classmethods or using a serialization library.

I personally use converters sparingly for things like `int` or `yarl.URL`: simple transformations and IIRC, that’s why it was added originally.

***

That said, it keeps coming up, it’s an obvious hole in our API so I guess if executed well I’d merge it so we don’t have to argue about it anymore. The API impact itself shouldn’t be too bad, it’s just tedious to implement because it touches so many internal structures.

altendky:
I'm forgetting and searching didn't turn up any examples.  How do you write a `versionadded` for the 'next' version?

altendky:
It has been awhile since I worked with hypothesis but it kind of looks like there aren't really any strategies for converters so nothing to update there?

altendky:
The new PR template covers the `versionadded` question.

BrendanSimon:
> I feel like everybody who asks for more complex converters, eventually realizes that they are not enough for what they want to do. They usually end up writing factory methods using classmethods or using a serialization library.

Agreed.  I think that classmethods are probably a good solution for my use case.  The implementer has more flexibility and is not dependent on attrs.  i.e. Do any manipulations you want and ensure the created class matches the attrs generated __init__.


> I personally use converters sparingly for things like `int` or `yarl.URL`: simple transformations and IIRC, that’s why it was added originally.
> 
> That said, it keeps coming up, it’s an obvious hole in our API so I guess if executed well I’d merge it so we don’t have to argue about it anymore. The API impact itself shouldn’t be too bad, it’s just tedious to implement because it touches so many internal structures.

Maybe just better documentation and examples around this area would reduce the noise and desire to get too creative :)

altendky:
It looks like the mypy plugin might need updating if I understand correctly.  Anyone interested in working on that piece?

euresti:
Converters are the hardest thing in the mypy plugin.  My guess is this will not be easy.  (I'm working on updating the plugin with @define etc)

hynek:
JFTR: unless someone motivated turns up, I'm super OK with converter support being mostly YOLO/Any.

euresti:
So technically we kind of already do that if we can't figure out what's going on with the converter.

```
import attr
from typing import Union

def my_converter(x: Union[int, str]) -> int:
    return 5 + int(x)

@attr.dataclass
class A:
    x: int
    y: int = attr.ib(converter=my_converter)
    z: int = attr.ib(converter=lambda x: 5 + int(x))

reveal_type(A.__init__)
```

Currently mypy will say this:
```
foo.py:10: error: Unsupported converter, only named functions and types are currently supported
foo.py:13: note: Revealed type is 'def (self: foo.A, x: builtins.int, y: Any, z: Union[builtins.int, builtins.str])'
``

However, making all converters go to Any would mean that we'd lose the type annotation for z.  
Though to be fair I wonder if the error on line 10 is necessary.  I kind of would like a warning so that the user knows why it's being converted to Any.   (The user can just add `# type: ignore # Unsupported converter` to their code and move on.

**Comment by nyanpasu64** [comment]
Sorry for the late reply... I'm not sure if it was worth replying or not.

> The first thing for me is `@classmethod` says 'this method is related to the class rather than just arbitrarily tossed in here even though it could be a regular function'. Inside the method `cls` says 'i am going to create an instance of this class' rather than just some class whose name you hopefully typed correctly. And yes, subclassing. Though as mentioned I ran into trouble getting type hints playing nicely. (and I mostly don't inherit)
> 
> On the flip side, what's the benefit of `@staticmethod` here?

I think it's more idiomatic in Rust to define `Foo`'s method as returning `Foo` and constructing `Foo{...}` than `Self{...}`. So I just kinda carried it over into Python, calling `Foo(...)` instead of `cls(...)`.

> Not that this is an end-all to the discussion but... [mail.python.org/pipermail/python-ideas/2016-July/041189.html](https://mail.python.org/pipermail/python-ideas/2016-July/041189.html)

Maybe in Python I should call cls instead...

**Comment by hynek** [comment]
@altendky is this ready for review? If so, pls click the button that says so. :)

**Comment by altendky** [comment]
I had a few different places where I was maybe going to get into mypy plugins, but none of them ended up that way.  Other than the unchecked bits in the OP, this is ready afaik.  A review may end up concluding that they aren't needed so I'll go ahead and un-draft this.

**Comment by Drino** [comment]
Hello!

I've looked through the discussion and a can't understand why converters are taking `self` like factories, not `(self, attr, value)` like validators.

E.g. I think "convert attribute to its type" is a pretty common use-case:
```
def type_converter(self, attr, value):
    return attr.type(value)  # actual code may be more complex

@attr.dataclass(frozen=True)
class Class:
    a: int = attr.ib(converter=type_converter)
    b: str = attr.ib(converter=type_converter)
```
or even
```
@attr.dataclass(frozen=True, converter=type_converter)
class Class:
    a: int
    b: str
```
Or this is kind of thing that is expected to be done with `field_transformer`?

**Comment by sscherfke** [comment]
@Drino your last example is something that can be done with field transformers.

I already wrote quite a powerfull auto converter but it is not yet merged into attrs:
- https://github.com/sscherfke/attrs/blob/master/src/attr/hooks.py
- https://github.com/sscherfke/attrs/blob/master/tests/test_hooks.py#L202-L379

**Comment by hynek** [comment]
> I've looked through the discussion and a can't understand why converters are taking `self` like factories, not `(self, attr, value)` like validators.

If you mean in general, it's simply because people wanted to be able to pass `int` or `str` as an argument.

---

I think that it might be a valid question if we should or should not allow the decorator option to be more powerful?

**Comment by BrendanSimon** [comment]
My view is more powerful.  I wanted to access another attribute in the class instance to use in the converter.  e.g. trimming data bytes to a max number of bytes (which is an attribute of the class instance settable at initialisation time).  Maybe there are other ways of doing that?  I want a setter for one attribute, that uses other attributes.  Are converters the right API to use for a attribute setter is or something else more appropriate?

**Comment by Drino** [comment]
@sscherfke The field_transformers are great and extremely powerful :) Thank you for bringing them into the attrs! I just have a feeling like API here is a little bit inconsistent and can be simpler for this case (e.g. `on_setattr` even has a `@attr.s`-level hook, but for `validator`/`converter` - which look prertty similar to `on_setattr` - there is no such thing and it has to be done via `field_transformer`). 

@BrendanSimon Sounds like this case can be solved only by passing `self` into `converter` (or via `__attrs_post_init__`). This wouldn't break if extended converter signature would be `self, attr, value`.

**Comment by BrendanSimon** [comment]
But would it work as a setter?  i.e. if I did something like:

```python
    instance.data= 'this is my really long bit of data I want to trim if too long'
```

Drino:
@BrendanSimon converters are applied only on `__init__`. If you want your converter applied on setattr, you can set  `on_setattr=attr.setters.convert` or even `on_setattr=[attr.setters.convert, attr.setters.validate]`.

jgerityneurala:
> I think that it might be a valid question if we should or should not allow the decorator option to be more powerful?

I would be pretty surprised if the functionality available differed from the inline form, but I could _maybe_ see a case for `@someattr.converter(pass_self=True)`? I don't think I'd personally use it, I tend to agree with the previous discussion about passing around a partially-initialized instance.

This PR has a lot going on as it is, my inclination is to recommend exact feature parity here, with potential for this kind of change as a separate issue.

BrendanSimon:
@Drino I'm fine with using `on_setattr`, as long as I have access to the instance so I can access other attributes of the class.

From memory, I tried using the validator setter, but what I set got overidden when the validator returned.

The converter made more sense, but again from memory I didn't have access the instance (self) and I couldn't use a function decorator as I could with the validator.

sscherfke:
I’m currently trying to figure out whether this PR is ready to merge or if the recent discussions prevent merging.

If I see it correctly, this PR just allows a converter function to be wrapped with a `Converter` class.
This class wraps the converter and allows you to addtionally pass `self` to the converter (instead of just the value).
The PR also adds a `converter` decorator so that converters can be specified as decorated functions instead of passing them to `attrib()`.

So in summary, this PR just improves the way users can specifiy converters.
It does not alter the way how setattr hooks or anything else works.

I also think that the `field_transformer` hook does not interfere with this PR. The task of this hook is to automatically all attributes of a class (e.g., by adding converters), but the hook itself does not care what or how exactly the attributes are modified.  E.g., you can add both, classical converter functions and new `Converter` instances.

Is this correct or are some changes needed before this PR can be merged?

Drino:
Besides for the API design - I believe there should be an update to `attr.converters.default_if_none` (to allow `takes_self=True` in `Factory` instance, as with this PR it will be possible) and `attr.setters.convert`.

hynek:
@Drino do I understand it correctly, that all your wishes would be fulfilled, if the Converter class that gets introduced here didn't pass just self and val but also the attribute?

I _think_ I can get behind that because that would make Factory and Converter use the same APIs and their syntactic sugar versions (`factory=` and `converter=`) would be optimized for shortness.

Am I missing something? 🤔

hynek:
(@altendky you'll have to update the version tags to 20.3.0 – sorry. It's the last time, I promise.)

altendky:
oof, I actually already had _one_ at 20.3.0...  `:[`

sscherfke:
I’m very eager to get v20.3 released. Shall we add the attrib as addtional argument to make the API of converter symetrical to validator and factory? It seems to me that this would be a relatively easy change. @hynek seems to be okay with it.

If not, can we merge this PR? 👼

Drino:
Thank you for your patience, sorry to delay !

@hynek It will be awesome if `Converter` class (and maybe decorated function) will allow to take full three-arguments, not only extra-self. Thank you for considering this.

altendky:
Sounds sensible. Maybe I can get to this tonight.

altendky:
Okay, I was just taking 'your' word for it that `Factory` takes the attribute now unlike before, but that doesn't look to be the case.  Or am I missing something?  Yeah, validators do take the attribute as well.

hynek:
Factories don't right now, but they might in the future: #692 🤔 People seem to increasingly be interested in writing general-purpose factories and converters.

The only question is whether we need to add an option to `Converter` to switch it off, but I don't see a reason for it? Worst case we can add that option later.

Drino:
I'm sorry for the cofusion :( I was pointing out that converters are more similar to `validator`/`on_setattr` than to `default`, not that `Factory` object takes attrib now. And nothing else takes 2-argument function in attrs:

|                                       | default (factory)                      | converter (this PR)                 | converter | converter (on_setattr-like)         | validator | on_setattr |
|---------------------------------------|----------------------------------------|-------------------------------------|-----------|-------------------------------------|-----------|------------|
| takes `value`                       | - (doesn't make sense)                 | +                                   | +         | +                                   | +         | +          |
| piped in `attr.ib`                      | - (doesn't make sense)                 | +                                   | +         | +                                   | +         | +          |
| `attr.*` library                      | -                                      | +                                   | +         | +                                   | +         | +          |
| decorator syntax/arity                | +/1                                    | +/2                                 | -         | +/3                                 | +/3       | -          (+/3 can be easily imagined, though) |
| arity if passed to `attr.ib`            | 0 when passed to factory or value/wrapper passed to default | 1 or wrapper, passed to converter | 1         | 1 or wrapper, passed to converter? | 3         | 3          |
| works while `self` is half-instantiated | +                                      | +                                   | +         | +                                   | -         | -          |
| wrapper object/arity                  | +/0-1 (wrapper has takes_self)         | +/1-2                               | -         | +/3                                 | -         | -          |


I think that wrapper for converter makes sense (like wrapper for Factory), as it may work with half-initialized self (like factory) and therefore is a little tricky.

BrendanSimon:
Awesome.  I presume the converter decorator can be applied to a class method, just like the validator decorator.  Yes?

Drino:
@BrendanSimon `validator` decorator can't be applied to neither `@classmethod` nor `@staticmethod` (both will give you `TypeError: '{decorator_name}' object is not callable`). So is `default` decorator. But I believe this discussion is out of PR scope.

BrendanSimon:
The docs here say validators can be used as decorator, which I've used with success.  I am hoping the same can be done with converter decorator, which I can use with `on_setattr`.

https://www.attrs.org/en/stable/api.html#validators

```python
@attr.s
class C(object):
    x = attr.ib()
    y = attr.ib()

    @x.validator
    def _any_name_except_a_name_of_an_attribute(self, attribute, value):
        if value < 0:
            raise ValueError("x must be positive")

    @y.default
    def _any_name_except_a_name_of_an_attribute(self):
        return self.x + 1

C(1)
C(x=1, y=2)

C(-1)
    ...
ValueError: x must be positive
```

altendky:
Since people are waiting...  Sorry, I meant to get to this tonight but ended up with some work stuff leaking over into the evening.  Next plan, tomorrow night.  \*fingers crossed\*

altendky:
Well, there's some progress anyways, though this is a stopping point tonight more like 'tests pass with changed code' than 'my work here is done'.  The `Converter` class is now beyond trivial serving only the purpose of allowing an `isinstance()` check.  Maybe it could be just a mark or something else instead.  I feel like there's some interaction with the new `setattr` stuff too that I haven't thought about yet.  Who knows what else.

hynek:
Thank you Kyle for staying at the ball, I can o my guess how this must be annoying at this point.

Since I’m busy rn I’d like to ask the other fans of this issue to have looks at it. 

Please do **not** expand the scope anymore though, the puck stops here and we have to find an end.

sscherfke (approved):
LGTM. However, since this is my first review for an attrs PR, someone else should take a look at it, too. :)

altendky:
@hynek, annoyed?  No...  :]  But I do have to get my head back into attrs a bit.  I worked a bit on the `setattr` stuff last night.  There are a couple more points I've added to the 'Draft for:' list in the OP.  Not scope creep, I don't think, just dealing with changes in attrs since this was originally written and the consistency with validators.

And I just realized I hadn't seen this back to draft... Getting that now.

sscherfke:
> And I just realized I hadn't set this back to draft... Getting that now.

Why?  Is something still missing or are you just waiting for a reply from @hynek ?

altendky:
@sscherfke, I try to keep a todo list in the first post of my PRs.  As I identify things to do I add them, then check them off when done.  I just took a few days to work on getting a new pyqt5-tools out this morning.  Assuming it doesn' t need immediate fixup, maybe I'll get back to this.  If you have any suggestions about the open issues in my 'Draft for:' list, I'm happy for some help.  `:]`

hynek:
Taking off the milestone to take off the pressure. Take all the time you need, thank you staying with us!

alonme:
Hey @altendky @hynek  - what is status for this PR?
Any help needed?

SubeCraft:
Any merge update ?

altendky:
Anyone is welcome to branch off my work and try to get it fully ready for merge.  I don't know what more has come up in the last few years, but I did leave a list of needed work, or at least things to consider, before I thought it would be ready.

Sadly, I don't make it out to community development much these days so I don't expect to be working on this.  Nor do I get to work with attrs, which also is sad.  `:[`

hynek:
> Anyone is welcome to branch off my work and try to get it fully ready for merge. I don't know what more has come up in the last few years, but I did leave a list of needed work, or at least things to consider, before I thought it would be ready.

Yeah I'm keeping it open, because I've started a few times to implement the three-argument converters myself and using is for inspiration, but if you care about performance, it's a bit trickier than it seems on the first sight and something always came up. :|

> Sadly, I don't make it out to community development much these days so I don't expect to be working on this. Nor do I get to work with attrs, which also is sad. `:[`

💔 – thank you for all your contributions!

PR #406, opened 2018-07-05
Add narrative docs for type annotations

hynek:
Narrative docs on type annotations to quell some of the FUD that is circulated.

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -43,8 +43,8 @@ Mypy will catch that error for you:
 
 This happens *without* running your code!
 
-And it also works with one of the two supported Python 2-style annotations.
-To mypy, this code is equivalent::
+And it also works with *both* Python 2-style annotation styles.
+To mypy, this code is equivalent to the one above::
 
   @attr.s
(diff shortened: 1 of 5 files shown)


docs/types.rst
```diff
@@ -0,0 +1,58 @@
...
+Mypy will catch that error for you:
+
+.. code-block:: bash
+
+   $ mypy t.py
+   t.py:12: error: Argument 1 to "SomeClass" has incompatible type "str"; expected "int"
+
+This happens *without* running your code!
+
+And it also works with one of the two supported Python 2-style annotations.
```
anthrotype: And what about when one uses the `type` argument to `attr.ib`? Will mypy also typecheck those?

hynek: It basically says so literally in the paragraph you’ve quoted. I take it as an Signal that i should be clearer.

anthrotype: Sorry, I see now the example has a line with a type argument, only read the one above which uses py2 comment-style type annotations and misunderstood. All good, thanks!

PR #410, opened 2018-07-18
Do not copy __weakref__ from original _cls_dict

### LINKED ISSUE #407: Slots class holding a reference to the original version
I have a class hierarchy implemented with slots enabled, like this:

import attr

```python
@attr.s(slots=True)
class BaseClass(object):
    foo = attr.ib(default='foo')

@attr.s(slots=True)
class SubClass(BaseClass):
    bar = attr.ib(default='bar')
```

This works fine until I wanted to introspect the class hierarchy:

```python
BaseClass.__subclasses__()  # returns [<class '__main__.SubClass'>, <class '__main__.SubClass'>]
```

One of these is the original class, and the other is the new one created and returned by `attr.s`. This behavior manifests on both Python 3 and Python 2 (3.7.0 and 2.7.14, to be specific).

My understanding was that classes kept weak references to their subclasses to allow the `__subclasses__` method to work, and I know that attrs creates and returns a new class when `slots=True`, but it's not clear to me why the old class stays around. I don't see any obvious place where a strong reference to it is held.

I guessed that there might be a reference cycle somewhere, so I tried adding a call to `gc.collect()` and checking `gc.garbage`, but that turned out to be incorrect.

Because I didn't see that the reference-counting mentioned [in this comment](https://github.com/python-attrs/attrs/issues/102#issuecomment-252526692) was ever addressed, I also guessed that this might be a leak due to the `__class__` closure cell fixup. I decided that this isn't the cause, either, because Python 2 does not have the `__class__` closure cell.

gmacon:
This fixes #407 as discussed.

# Pull Request Check List

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,7 +99,7 @@ Testimonials
   It exerts a subtle, but positive, design influence in all the codebases I’ve see it used in.
 
 
-**Kenneth Reitz**, author of `requests <http://www.python-requests.org/>`_, `on paper no less <https://twitter.com/hynek/status/866817877650751488>`_:
+**Kenneth Reitz**, author of `Requests <http://www.python-requests.org/>`_ and Developer Advocate at DigitalOcean, (`on paper no less <https://twitter.com/hynek/status/866817877650751488>`_!):
 
   attrs—classes for humans.  I like it.
 


tests/test_make.py
```diff
@@ -1250,6 +1251,25 @@ class C(object):
...
+        @attr.s(slots=True)
+        class C(object):
+            pass
+
+        @attr.s(slots=True)
+        class C2(C):
+            pass
+
+        # The original C2 is in a reference cycle, so force a collect:
+        gc.collect()
```
njsmith: Sometimes calling `gc.collect()` once isn't enough to actually collect all garbage:
https://github.com/python-trio/trio/blob/45b4e586897f2676169b89e6b9f48d54eb14621a/trio/_core/tests/tutil.py#L26-L37
In this case I think it is probably fine but generally I feel like by the time you're calling `gc.collect` in a test you're already having a bad day and don't want to get surprised by any *new* edge cases.

gmacon: That's interesting; I wasn't aware of that. IIUC, we'll be OK here because there are no `__del__` methods involved; does that sound right to you?

njsmith: I was going to say "yes", but actually... How certain are we that there's no `__del__` involved in clearing the parent's subclasses table? (I suppose it probably uses weakrefs rather than `__del__`, but I'm not certain, and I'm not certain how weakrefs interact with `gc.collect` anyway.)

PR #413, opened 2018-07-27
Typo fixes

p-leger:
Hi,

I noticed that there was an indentation error in the `examples.rts`, which made the validations section a bit hard to read. 
I decided to make my first contribution, and went to check on your Contribution guide, while reading that I wanted to setup up my development environment and noticed that the git command read `git checkout git@github.com:python-attrs/attrs.git`, I suppose that was meant to be `git clone`. 

Furthermore I changed it from using `ssh` to `https` since using `ssh` without a configured key resulted in a `Permission denied (publickey)` error.

I also struggled with `pip install -e .[dev]`, since `zsh` was trying to do a bit of globbing, using quotes fixed that.

I hope i didn't miss anything obvious. 
i removed the check list, since I believe none of the entries are applicable to this pull request.

Cheers, 

Philipp


hynek:
Hmm the clone thing is tricky, if we change it to https, it will be a pain for everyone who uses ssh (I never remember the correct set-url syntax).

How about keeping ssh and say something like “or if you don’t use git over ssh:” and then add the second option?`

p-leger:
Sure.

I rebased the branch and amended the last commit (`f8ee7848a`) since i didn't want to create a fixup commit/extra one for such a small change.

hynek:
Thanks! FWIW, don’t worry about the number of commits, we always squash on merge anyway. Just do whatever is most comfortable/familiar for you.

p-leger:
cheers.

Issue #400, opened 2018-06-22
Most elegant way to avoid None objects in class?


madig:
(continuing from https://github.com/python-attrs/attrs/issues/170 which I don't want to derail)

I have a class like this:
```python
@attr.s(frozen=True, auto_attribs=True)
class Status:
    weight: int = attr.ib(cmp=True)
    message: str = attr.ib(
        cmp=False,
        converter=lambda value: str() if value is None else value,
    )
    code: str = attr.ib(
        cmp=False,
        converter=lambda value: str() if value is None else value,
    )

    @classmethod
    def Debug(cls, message=None, code=None):
        return cls(weight=0, message=message, code=code)

    [...]

    @classmethod
    def Error(cls, message=None, code=None):
        return cls(weight=6, message=message, code=code)
```
It can be instantiated like this: `Status.Error("This and that is wrong etc.", code="mismatched-something")`. However, instantiating it like `Status.Error()` should result in `Status(weight=6, message="", code="")`, i.e. I want to avoid None objects. Python unfortunately makes it more tedious than it should be to give out non-None default arguments.

@hynek noted that I may want to use `default=""` and use `attr.NOTHING` as the default argument in the constructors, however:
```
In [1]: import attr
   ...: @attr.s
   ...: class Status:
   ...:     a = attr.ib(default="")
   ...:     @classmethod
   ...:     def aaa(cls, a=attr.NOTHING):
   ...:         return cls(a)
   ...:

In [2]: Status.aaa()
Out[2]: Status(a=NOTHING)

In [3]: Status.aaa("")
Out[3]: Status(a='')

In [4]: import attr
   ...: @attr.s
   ...: class Status:
   ...:     a = attr.ib(default="")
   ...:     @classmethod
   ...:     def aaa(cls, a=None):
   ...:         return cls(a=a or attr.NOTHING)

In [5]: Status.aaa()
Out[5]: Status(a=NOTHING)
```

madig:
(I suppose I can use `def aaa(cls, a=""):` here because strings are immutable, but I'd like to know about a more general approach.)

hynek:
Ugh right the NOTHING way doesn’t work because attrs uses literal `""` as default parameters if possible.

hynek:
[...]

PR #414, opened 2018-07-28
Implement attr.converters.default_if_none

hynek:
So this is a stab at #400 which I find a reasonable usage of converters.

Typing tests fail tho and they also fail for `attr.validators.optional`. Am I right to suspect that this has to be fixed in mypy's plugin? Gives us a bit of a Mexican stand-off, doesn't it? It’s completely possible that I additionally got the stubs wrong so feel free to advice.

cc @euresti @chadrik


euresti:
Yeah this is something the plugin currently doesn't support:

`error: Unsupported converter, only named functions and types are currently supported`

The underlying problem is that it's hard to extract the type of the function call in the mypy pipeline.

A workaround could be to write:
```
optional_int = optional(int)

@attr.s
class ConvCOptional:
    x: Optional[int] = attr.ib(converter=optional_int)
```

But that's broken also, but I think easier to fix.  Now I remember, in our work code, when we started using mypy I removed all our uses of optional and just wrote our own converters.

hynek:
I guess until mypy opens up it's plugin API (I talked to Jukka here at EuroPython and it's legit on their roadmap) we’ll have to comment stuff out, wait for mypy to release the support? Having hard-coded support is “fine” for now…

***

Does the signature look correct to you tho? If someone rubber-stamps it, I’ll just comment it out and we open an issue on mypy.

euresti:
I can review the signature.  As for commenting it out I'm not sure since I'm not sure when we'll fix the plugin.  It might be a while.

hynek:
Just to be clear: by special casing I mean that mypy only supports our shipped converters for now. Would that be a significant amount of work too?

euresti:
Hmm.  Maybe I could detect the converters by name and handle them specially.  That might actually work.

hynek:
> Hmm. Maybe I could detect the converters by name and handle them specially. That might actually work.

Yeah, that’s what I was thinking.

euresti:
FYI: https://github.com/python/mypy/pull/5411   Here's one implementing `optional`, we'll see what the mypy folks say.

Issue #324, opened 2018-01-16
Auto-detect user-written methods


hynek:
Currently, if you want to implement a method yourself, you have to remember to do two things:

```python
@attr.s(repr=False)
class C:
    def __repr__(self):
        return "whatever"
```

I’ve seen quite a few people be confused/frustrated by that.

***

I think unless `these` is set, it would be kind of cool if we detected that the current class (and not some super class) has a user-written method and set the flag to false automatically such that

```python
@attr.s
class C:
    def __repr__(self):
        return "whatever"
```

does what you’d expect.

Opinions?

euresti:
Same with `__init__`?  (I ask because the mypy plugin will have to deal with that too.)

hynek:
The idea is to support it for all of them.  It gets a bit hairy when we get to comparison because it’s several ones.  Would we stop completely?  Or just not replace one of them?

wsanchez:
With comparison methods, perhaps [total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering) could provide guidance; if you provide enough for `total_ordering` to work, we apply it or do something similar?

It's a bit of feature creep, but it feels in line with the developer doing as little work related to attributes as necessary and we take care of the rest.

hynek:
The more I think about it the more I think we should just skip methods that have been defined and be done with it.  That allows people to customise their classes even better.

PR #416, opened 2018-07-31
 Auto-detect user-written methods

p-leger:
Hey,

I was browsing through the open issues searching for something I could take a stab at implementing.
I came across #324 and thought i would like to try that.

As a a set of 'specifications' i took last comment form @hynek on that issue:

>The more I think about it the more I think we should just skip methods that have been defined and be done with it. That allows people to customize their classes even better. 

I thought about putting that logic into the `attrs` method, however I wouldn't have fine enough control over all methods added by `add_cmp` so decided to move that logic into the `ClassBuilder`.

I can imagine people are used to `attrs` overwriting methods by default even if they are defined, thus I would suggest a feature switch (which I haven't implemented yet.).

The tests gave me a bit of trouble, especially when running against Python2.7, since it seems like `unbound` and `bound` methods are not eq (please correct me if i misunderstood this).

Furthermore if the Class should be `unhashable` any user written method is still ignored. However the argument, "A `__hash__` method has been supplied, thus the class should be considered hashable" also makes sense to me. 

This PR is still very much work in progress, before I spend more time on this, I would greatly appreciated some feedback.

Cheers, 

Philipp

# Pull Request Check List

- [ ] Documentation in `.rst` files is written using [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).

### DIFF
--- a/changelog.d/324.change.rst
+++ b/changelog.d/324.change.rst
@@ -1,0 +1,1 @@
+``attr.s`` now auto-detects user-written methods and does not overwrite them. # is missing the __hash__ == None if it shouldn't be hashable
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -556,29 +556,27 @@ class _ClassBuilder(object):
 
     def add_repr(self, ns):
         if "__repr__" in self._cls.__dict__:
-            meth = getattr(self._cls, "__repr__")
-        else:
-            meth = self._add_method_dunders(_make_repr(self._attrs, ns=ns))
+            return self
 
-        self._cls_dict["__repr__"] = meth
+        self._cls_dict["__repr__"] = self._add_method_dunders(
+            _make_repr(self._attrs, ns=ns)
+        )
         return self
 
     def add_str(self):
         if "__str__" in self._cls.__dict__:
-            meth = getattr(self._cls, "__str__")
-        else:
-            repr = self._cls_dict.get("__repr__")
-            if repr is None:
-                raise ValueError(
-                    "__str__ can only be generated if a __repr__ exists."
-                )
+            return self
 
-            def __str__(self):
-                return self.__repr__()
+        repr = self._cls_dict.get("__repr__")
+        if repr is None:
+            raise ValueError(
+                "__str__ can only be generated if a __repr__ exists."
+            )
 
-            meth = self._add_method_dunders(__str__)
+        def __str__(self):
+            return self.__repr__()
 
-        self._cls_dict["__str__"] = meth
+        self._cls_dict["__str__"] = self._add_method_dunders(__str__)
         return self
 
     def make_unhashable(self):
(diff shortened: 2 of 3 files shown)


hynek:
Heh this is not really an “easy” ticket (although it doesn’t need too much attrs tribal knowledge). The hard part (and problem with your PR) is that you can’t just look whether the method exists but you have to look whether the method exists in the *current class* vs. somewhere in the inheritance tree.

IOW:

```python
@attr.s
class A:
    x = attr.ib()
    def __init__(self):
        self.x = 42

@attr.s
class B:
    y = attr.ib()
```

Your code would detect A’s `__init__`.

Which also kinda poses an interesting question how that should be handled. My gut instinct is to set init=False on A, leave ClassBuilder alone, and make the user call `super()`.

Which takes us to the problem you’ve encountered with cmp. We could say that defining *any* comparison method disables the creation wholesale. But since we plan to make cmp more granular, that would mean more headaches down the road.

God I don’t have a good answer, but it seems that I won't be able to keep dodging the replacement of bools thru enums much longer. :|

p-leger:
I didn't think it would be an easy one, I basically just wanted an excuse to hack on the `attrs` source code ^^.

If I understood correctly the main issue is methods provided by `super classes` would be called:

I though I had solved that, by doing the following check:
```ruby
class C:
    def __init__(self):
        return 'here'


class F(C):
    def foo(self):
        return 'asdf'


print(f'C: {C.__dict__.keys()}')
print(f'F: {F.__dict__.keys()}')
``` 

Which gives be the following output:
```
C: dict_keys(['__module__', '__init__', '__dict__', '__weakref__', '__doc__'])
F: dict_keys(['__module__', 'foo', '__doc__'])
```

If I understood by code correctly, I am checking `__dict__` for just the class which is decorated with `attr.s`, what am I missing here?

---

If I understood correctly the idea of this library is making the creation of class which hold data as easy as possible without having to  I believe the biggest challenge is finding the

p-leger:
I suppose you example was meant to be:

```python
@attr.s
class A:
    x = attr.ib()
    def __init__(self):
        self.x = 42

@attr.s
class B(A):
    y = attr.ib()
```

(If not i misunderstood your point)

I didn't think it would be an easy one, basically I just wanted an topic/problem/excuse to hack on the `attrs` source code :grin:.

If I understood correctly the main issue is methods provided by `super classes` would also be detected and thus not overwritten by `attrs`:

I though I had solved that, by doing the following check:
```python
class C:
    def __init__(self):
        pass


class F(C):
    def foo(self):
        pass


print(f'C: {C.__dict__.keys()}')
print(f'F: {F.__dict__.keys()}')
``` 

Which gives be the following output:
```
C: dict_keys(['__module__', '__init__', '__dict__', '__weakref__', '__doc__'])
F: dict_keys(['__module__', 'foo', '__doc__'])
```

So I am checking for the method in question inside `__dict__` for the class which is decorated with `attr.s`, what am I missing here?

hynek:
> So I am checking for the method in question inside `__dict__` for the class which is decorated with `attr.s`, what am I missing here?

Slotted classes that don’t have a `__dict__`? 😇

p-leger:
heh.. yeah indeed. (But in terms of tackling the inheritance problem, it would have worked if it weren't for slotted classes?)

So over the last couple of weeks I have used `attr` a bit more which got me thinking about this PR...

I started to wonder if the added convenience really does outweigh the complexity/'magic', which would be introduced by implementing it.

So I think I'll just close this PR and see if I can find something which is a bit more straightforward -- any suggestions?

hynek:
Sorry for not getting back at you any sooner. Yes it might be, that it’s actually not worth the complexity. Sadly it usually isn’t entirely clear until one sees that code/complexity realized. :|

Issue #325, opened 2018-01-17
First class __weakref__ support


hynek:
Currently you get weakref support using a hack that is outlined in #174 that currently doesn’t work properly (#311).

So people like @njsmith rightfully asked for a better solution.  This ticket is to collect all input and decide on an API.

Looking forward to y’all’s input.

njsmith:
My naive suggestion would be to add a new class-level switch `weakref=True`, to control whether classes should be weakrefable. For regular classes this is trivial, so it wouldn't do anything. (Maybe it would be an error to pass `weakref=False`.) For `slots=True` classes, it would control whether attrs auto-added a `__weakref__` slot. By defaulting to `weakref=True`, we made it so that merely toggling between `slots=False` and `slots=True` stops being a breaking change in your public API.

Now I expect there will be some list of reasons involving backcompat why this pretty like suggestion has to be replaced by something more complicated...

altendky:
There wasn't much input but perhaps I can spur some again.  Didn't take too long so not much to lose, don't worry.  :]

altendky:
I suppose I should mention that the reason I ended up looking at this was that I was unable to connect a pyqt signal to a method of a slotted attrs class due to the lack of weakref support.

hynek:
Fixed by #420

PR #420, opened 2018-08-01
T325 weakref with slots

altendky:
#325


altendky:
Hmm, I should probably add a test to confirm that no extra fields are added.  Kind of a given I think, but seeing as the workaround involved creating one and I don't think we want that, it seems worth a couple lines of test code I show and verify that desire.

I should also probably change my added tests to use hypothesis...  :|

hynek:
Hmmm, IIUC, you’re only setting `__weakref__` if it's not set already, right? ISTM like we could set the option to True by default because I simply can’t come up with a realistic scenario that might break here?

It’s even questionable whether it should be an option at all, but if it is, it might even be renamed to `slots_weakref` or something since it has a very narrow use case.

Am I missing something @Tinche @njsmith?

njsmith:
I'd be fine with switching on the `weakref=True` behavior unconditionally. I only really [suggested](https://github.com/python-attrs/attrs/issues/174#issuecomment-333335228) making it a switch because @Tinche [was against](https://github.com/python-attrs/attrs/issues/174#issuecomment-290531450), wasting 8 bytes of memory per instance when a slotted type is never actually weakref'ed, so I was hoping making it an option would sneak it past his objections :-).

altendky:
@hynek, yes, the intent was to only add `"__weakref__"` if it wasn't already in `cd`...  a thing that was just created that explicitly avoids including `"__weakref__"`.  So that check is pointless.  But I am missing a check for `"__weakref__"` already being in the super class.  Adding now.  Presently `type()` raises an exception.

```python3
TypeError: __weakref__ slot disallowed: either we already got one, or __itemsize__ != 0
```

Same thing happens with an attribute named `__weakref__`.

altendky:
If the option is kept and renamed, perhaps `weakref_slot`?  That seems to fit a bit better in my head.  It is a thing (like `init` or `hash` or `slots`) which we either create or don't.

Would we want an exception in case of `slots=False` and `weakref=True`?  Though this would make more sense if it were `weakref_slot`.  If not renamed, it could be argued that `slots=False, weakref=True` is actually valid since you can weakref without slots and that the invalid case is actually `slots=False, weakref=False` since we don't disable weakrefability.  So I think yes, rename to be slot-specific and raise an exception if set to True without slots.

changelog.d/420.change.rst
```diff
@@ -0,0 +1 @@
+Add weakref parameter to attr.s() allowing for weak referenceable slots classes.
```
hynek: We call them “slotted classes”: http://www.attrs.org/en/stable/glossary.html as you do later in the diff. :)

src/attr/_make.py
```diff
@@ -719,6 +730,8 @@ def attrs(
                ``object.__setattr__(self, "attribute_name", value)``.
 
         ..  _slots: https://docs.python.org/3/reference/datamodel.html#slots
+    :param bool weakref: Make instancess weak-referenceable.  This has no
```
hynek: “instancess”

changelog.d/420.change.rst
altendky: FYI, a search actually shows several hits for `slots class`.  But yes, I'm correcting this.

altendky:
Once we switch to `weakref_slot=True` as the name and default then an exception becomes less obvious to me.  It seems the case to raise an exception then would be a regular non-slotted class without specifying `weakref_slot=False` because, well, a slot was 'requested' (by default) and we aren't going to be making it.  Since we will be defaulting to `True`, `weakref` or `weakrefable` makes more sense to me and we raise an exception in any case where the requested state is not achieved (though that leads to exceptions in PyPy which wouldn't occur in CPython).  I dunno, I guess I'm not totally sold on the exception (that I suggested).

I'm coding towards `weakref_slot=True` but won't add the exception yet.

altendky:
Changing the default broke some tests so I stashed that and went ahead and added parametrization to them for `weakref_slots` `True`/`False` to show the issue first.  But, it's bedtime so I'll have to try to fix that later.  Then back to making the default `True`.

njsmith:
It does make sense to me that `weakref=True` would mean "this has to support weakrefs", and `weakref=False` would mean "I don't care if this supports weakrefs". I don't think anyone is writing code that *relies on* getting an exception when attempting to take a weakref, and it would be very annoying if you had to go through and remove all your `weakref=False` kwargs in order to port to PyPy.

altendky:
I couldn't reason to a sensible exception with the name and default `weakref_slot=True`.  Presumably we wouldn't want the default to result in an exception in any case.  And for the value `False`, we wouldn't want the default of a not-slotted class to raise an exception, I mean it doesn't have a weakref slot as requested...  So I'm just not sure what to implement for an exception.

Anyways, I'm not waiting on anyone but myself to get back to this and figure out the failures.

altendky:
Do we have outstanding actions on this at this point?  Should any of the tests I added use Hypothesis?  Or others that haven't been added but should be?  `weakref_slot` has already been added so that other things using the Hypothesis strategies should get variations against `weakref_slot`.

changelog.d/420.change.rst
```diff
@@ -0,0 +1 @@
+Add weakref parameter to attr.s() allowing for weak referenceable slotted classes.
```
hynek: This needs an update of the name and `attr.s()` should be in double backticks.

docs/glossary.rst
```diff
@@ -43,21 +43,11 @@ Glossary
         Those methods are created for frozen slotted classes because they won't pickle otherwise.
         `Think twice <https://www.youtube.com/watch?v=7KnfGDajDQw>`_ before using :mod:`pickle` though.
 
-      - As always with slotted classes, you must specify a ``__weakref__`` slot if you wish for the class to be weak-referenceable.
-        Here's how it looks using ``attrs``:
+      - Slotted classes are weak-referenceable by default.  This can be disabled in CPython by passing ``weakref_slot=False`` to ``@attr.s`` [#pypyweakref]_.
```
hynek: We use [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) please.

src/attr/__init__.pyi
```diff
@@ -164,6 +164,7 @@ def attrs(
     init: bool = ...,
     slots: bool = ...,
     frozen: bool = ...,
+    weakref: bool = ...,
```
hynek: needs update

src/attr/__init__.pyi
```diff
@@ -180,6 +181,7 @@ def attrs(
...
```
hynek: needs update

src/attr/__init__.pyi
```diff
@@ -207,6 +209,7 @@ def make_class(
...
```
hynek: needs update

src/attr/_make.py
```diff
@@ -759,6 +785,8 @@ def attrs(
                ``object.__setattr__(self, "attribute_name", value)``.
 
         ..  _slots: https://docs.python.org/3/reference/datamodel.html#slots
+    :param bool weakref: Make instances weak-referenceable.  This has no
```
hynek: needs update

tests/strategies.py
```diff
@@ -147,24 +147,26 @@ def simple_attrs_with_metadata(draw):
...
         _b = attr.ib(default=None)
         c = attr.ib(default='text')
         _d = attr.ib(default=1.0)
         c = attr.ib(default={'t': 1})
 
-    By default, all combinations of slots and frozen classes will be generated.
-    If `slots=True` is passed in, only slots classes will be generated, and
-    if `slots=False` is passed in, no slot classes will be generated. The same
-    applies to `frozen`.
+    By default, all combinations of slots, frozen, and weakref classes will be
```
hynek: needs update

tests/strategies.py
```diff
@@ -147,24 +147,26 @@ def simple_attrs_with_metadata(draw):
...
+    generated. If `slots=True` is passed in, only slots classes will be
+    generated, and if `slots=False` is passed in, no slot classes will be
+    generated. The same applies to `frozen` and `weakref`.
```
hynek: needs update

altendky:
@hynek, sorry...  i thought i had done all that a few days ago.

hynek:
Thanks!

Issue #423, opened 2018-08-10
Caching hashcodes


rgabbard:
We currently have an application which is a heavy user of attrs.  For certain operations, a majority of the application's time is being spent in the hash method of attrs objects.  The frozen attrs objects in question are complex and themselves contain other frozen attrs objects as fields (this goes several layers deep).  Being able to cache the hash code value by user request at upper levels of this object tree ((like `prehash` in [Immutables](https://immutables.github.io/immutable.html))) would lead to a significant performance improvement for us.

Issue https://github.com/python-attrs/attrs/issues/261 discussed some performance issues with hashing and touched on this.  However, this idea seemed to get lost in the following discussion, which is now closed, so it seems worth moving to its own issue.

On that issue @DRMacIver suggested the following:
* Add a `memoize_hash` flag to `attr.s` defaulting to `None`.
* When `memoize_hash` is `True`, store the hash on a field called `__hash`.
* If hash is `False` (or inferred `False` from a default) it is an error to pass non-`None` `memoize_hash`
* If `slots` and `memoize_hash` are both `True`, add a bonus `__hash` slot to the class to store it.
* If hash is `True`, `memoize_hash` defaults to `frozen` and not `slots`

hynek:
I think I could live with that but I think `cache_hash` is a better term for people to understand?

Also it should be `__attrs_cached_hash__` or something so we stay within our own namespace.

There’s one thing that concerns me here tho: just because a class is frozen, we cannot make the assumption that it’s true for all of its attributes (although that would certainly be the technical correct behavior). Therefore I see no way around making it strictly opt-in.

DRMacIver:
>  (although that would certainly be the technical correct behavior). 

Isn't the convention in Python that non-frozen things shouldn't be hashable at all? Is there widespread usage of mutable hashable objects?

(That being said, making it explicitly opt in seems perfectly reasonable)

hynek:
It might be a convention but I’m quite positive that at least 90% of Python programmers do not know that.

rgabbard:
I agree that strictly opt-in is safer.  

Here is my updated understanding of the plan:
```
* Add a `cache_hash` flag to attr.s defaulting to `False`.
* When `cache_hash` is `True`, store the hash on a field called `__attrs_cached_hash`.
* If hash is `False` (or inferred `False` from a default) it is an error to pass non-`False` `cache_hash`
[...]

PR #426, opened 2018-08-19
Cache hash codes

rgabbard:
# Pull Request Check List

- [X] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/strategies.py).  ** n/a,I think**
- [X] Documentation in `.rst` files is written using [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).

# Description

This adds a `cache_hash` flag to `@attrs` which causes the hash code to be computed once and stored on the object.  Setting this flag to `true` is allowed only if `attrs` is set to generate both the `__hash__` and `__init__` methods for a class.

This closes issues #423.

### DIFF
--- a/changelog.d/425.change.rst
+++ b/changelog.d/425.change.rst
@@ -1,1 +1,1 @@
-Added ``cache_hash`` option which causes the hash code to be computed once and stored on the object.
+Added ``cache_hash`` option to ``@attr.s`` which causes the hash code to be computed once and stored on the object.
--- a/changelog.d/425.change.rst
+++ b/changelog.d/425.change.rst
@@ -1,0 +1,1 @@
+Added ``cache_hash`` option which causes the hash code to be computed once and stored on the object.
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -52,6 +52,7 @@ Because according to the definition_ from the official Python docs, the returned
 
 For a more thorough explanation of this topic, please refer to this blog post: `Python Hashes and Equality`_.
 
+
 Hash Code Caching
 -----------------
 
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -58,10 +58,10 @@ Hash Code Caching
 
 Some objects have hash codes which are expensive to compute.
 If such objects are to be stored in hash-based collections, it can be useful to compute the hash codes only once and then store the result on the object to make future hash code requests fast.
-To enable caching of hash codes, specify ``cache_hash=True``.
+To enable caching of hash codes, pass ``cache_hash=True`` to ``@attrs``.
 This may only be done if ``attrs`` is already generating a hash function for the object.
 If the hash code is cached, no field involved in hash code computation may be mutated after construction.
-It is strongly recommended that classes with cached hashcodes be ``frozen``
+Therefore, it is strongly recommended that classes with cached hashcodes be ``frozen``.
 
 .. [#fn1] The hash is computed by hashing a tuple that consists of an unique id for the class plus all attribute values.
 
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -53,6 +53,12 @@ Because according to the definition_ from the official Python docs, the returned
 For a more thorough explanation of this topic, please refer to this blog post: `Python Hashes and Equality`_.
 
 
+Hashing and Mutability
+----------------------
+Changing any field involved in hash code computation after the first call to `__hash__` (typically this would be after its insertion into a hash-based collection) can result in silent bugs.
+Therefore, it is strongly recommended that hashable classes be ``frozen``.
+
+
 Hash Code Caching
 -----------------
 
@@ -60,8 +66,6 @@ Some objects have hash codes which are expensive to compute.
 If such objects are to be stored in hash-based collections, it can be useful to compute the hash codes only once and then store the result on the object to make future hash code requests fast.
 To enable caching of hash codes, pass ``cache_hash=True`` to ``@attrs``.
 This may only be done if ``attrs`` is already generating a hash function for the object.
-If the hash code is cached, no field involved in hash code computation may be mutated after construction.
-Therefore, it is strongly recommended that classes with cached hashcodes be ``frozen``.
 
 .. [#fn1] The hash is computed by hashing a tuple that consists of an unique id for the class plus all attribute values.
 
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -1,6 +1,9 @@
 Hashing
 =======
 
+Hash Method Generation
+----------------------
+
 .. warning::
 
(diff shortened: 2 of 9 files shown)


rgabbard:
@hynek : I believe this is now ready for review

src/attr/_make.py
```diff
@@ -805,14 +825,31 @@ def wrap(cls):
                 "Invalid value for hash.  Must be True, False, or None."
             )
         elif hash is False or (hash is None and cmp is False):
-            pass
+            if cache_hash:
+                raise TypeError(
+                    "Invalid value for cache_hash.  To use hash caching,"
+                    " hashing must be either explicitly or implicitly "
+                    "enabled"
```
hynek: End sentence with period please. :)

src/attr/_make.py
```diff
@@ -805,14 +825,31 @@ def wrap(cls):
...
+                )
         elif hash is True or (hash is None and cmp is True and frozen is True):
             builder.add_hash()
         else:
+            if cache_hash:
+                raise TypeError(
+                    "Invalid value for cache_hash.  To use hash caching,"
+                    " hashing must be either explicitly or implicitly "
+                    "enabled"
```
hynek: End sentence with period please. :)

src/attr/_make.py
```diff
@@ -805,14 +825,31 @@ def wrap(cls):
...
+                )
             builder.make_unhashable()
 
         if init is True:
             builder.add_init()
+        else:
+            if cache_hash:
+                raise TypeError(
+                    "Invalid value for cache_hash.  To use hash caching,"
+                    " init must be True"
```
hynek: End sentence with period please. :)

tests/test_dunders.py
```diff
@@ -283,8 +293,34 @@ def test_enforces_type(self):
 
         assert exc_args == e.value.args
 
-    @given(booleans())
-    def test_hash_attribute(self, slots):
+    def test_enforce_no_cache_hash_without_hash(self):
```
hynek: A [test docstring](https://jml.io/pages/test-docstrings.html) please

changelog.d/425.change.rst
```diff
@@ -0,0 +1 @@
+Added ``cache_hash`` option which causes the hash code to be computed once and stored on the object.
```
hynek: This should mention to what. I.e. “Added `cache_hash` option to `@attr.s`…”.

docs/hashing.rst
```diff
@@ -50,6 +53,16 @@ Because according to the definition_ from the official Python docs, the returned
 For a more thorough explanation of this topic, please refer to this blog post: `Python Hashes and Equality`_.
 
 
+Hash Code Caching
+-----------------
+
+Some objects have hash codes which are expensive to compute.
+If such objects are to be stored in hash-based collections, it can be useful to compute the hash codes only once and then store the result on the object to make future hash code requests fast.
+To enable caching of hash codes, specify ``cache_hash=True``.
```
hynek: Same as above. “, pass `cache_hash=True` to `@attr.s`.” maybe?

docs/hashing.rst
```diff
@@ -50,6 +53,16 @@ Because according to the definition_ from the official Python docs, the returned
...
+This may only be done if ``attrs`` is already generating a hash function for the object.
+If the hash code is cached, no field involved in hash code computation may be mutated after construction.
+It is strongly recommended that classes with cached hashcodes be ``frozen``
```
hynek: Period at the end. Maybe start sentence with `Therefore`?

Also thinking about this…it’s kind true for all hashes except that with caching, they mustn’t change after first hashing while with normal hashes they mustn’t change after been inserted somewhere. Is there a way to communicate that?

docs/init.rst
```diff
@@ -348,6 +348,7 @@ If you need to set attributes on a frozen class, you'll have to resort to the :r
    >>> Frozen(1)
    Frozen(x=1, y=2)
 
+Note that you may not access the hash code of the object in ``__attrs_post__init__`` if ``cache_hash=True``
```
hynek: Period at the end of the sentence please. :)

Maybe use stronger wording and say `*must not* access`?

src/attr/_make.py
```diff
@@ -862,29 +899,51 @@ def _attrs_to_tuple(obj, attrs):
...
+        )
+
+        for a in attrs:
+            method_lines.append(indent + "        self.%s," % a.name)
+
+        method_lines.append(indent + "    ))")
+
+    if cache_hash:
+        method_lines.append(tab + "if self.%s is None:" % _hash_cache_field)
+        # we use __setattr__ all the time so we don't need to special-case for
```
hynek: So we? You mean in `__init__` or am I missing something?

If not and since this appears to be a hot path, I think I would prefer a special case for it? I mean in `__init__` we try to shave of ns by caching everything.

tests/test_dunders.py
```diff
@@ -283,8 +293,34 @@ def test_enforces_type(self):
...
+        assert exc_args == e.value.args
+
+        # unhashable case
+        with pytest.raises(TypeError) as e:
+            make_class(
+                "C", {}, hash=None, cmp=True, frozen=False, cache_hash=True
+            )
+        assert exc_args == e.value.args
+
+    def test_enforce_no_cached_hash_without_init(self):
```
hynek: [test docstring](https://jml.io/pages/test-docstrings.html) please

docs/hashing.rst
rgabbard: @hynek : I tried adding the following section:

```
Hashing and Mutability
----------------------
Changing any field involved in hash code computation after the first call to `__hash__` (typically this would be after its insertion into a hash-based collection) can result in silent bugs.
Therefore, it is strongly recommended that hashable classes be ``frozen``.
```

This doesn't point out that it would be okay to modify the fields after calling `__hash__` but before collection insertion is the hash code is not cached, but this seems a weird enough (and dangerous enough) case that it should probably be avoided by users anyway.

src/attr/_make.py
rgabbard: Dropped this comment and did the special-casing.

rgabbard:
@hynek : Ready for re-review.  For all the trivial changes (periods, adding phrases to docs, docstrings, etc.) I made the suggested changes. In the two more complex cases, I commented on your comments to highlight my changes.

hynek:
Great work, thank you!

rgabbard:
@hynek : no problem, thanks for reviewing and merging!

Do you happen to know when the next release is likely to be?  My team would really like to use this feature and the new kw-only args support.

hynek:
Yeah it’s overdue, isn’t it. I want to fix #429 and #427 since both look like bugs and then release. So I hope this week if I find people to review my fixes.

rgabbard:
@hynek : Great, thank you!

PR #431, opened 2018-08-21
Allow to override class vars of super classes w/ annotations

### LINKED ISSUE #427: attr.ib cleanup + auto_attribs + class vars from superclasses
Consider the following:

```python
import attr

class Base:
    x = 0

@attr.s(auto_attribs=True)
class C(Base):
    x: int
```

This raises an:

```
  File "t.py", line 9, in <module>
    class C(Base):
  File "/Users/hynek/Projects/attrs/src/attr/_make.py", line 858, in wrap
    return builder.build_class()
  File "/Users/hynek/Projects/attrs/src/attr/_make.py", line 496, in build_class
    return self._patch_original_class()
  File "/Users/hynek/Projects/attrs/src/attr/_make.py", line 514, in _patch_original_class
    delattr(cls, name)
AttributeError: x
```

(interestingly `getattr(cls, name)` works but `delattr` does not)

This is actually quite common if you implement an ABC. The current code is:

```python
                if (
                    name not in super_names
                    and getattr(cls, name, None) is not None
                ):
                    delattr(cls, name)
```

Maybe we should change the check to `isinstance(getattr(cls, name, None), _CountingAttr)`?

hynek:
This is the best I could come up with for #427. Am I missing some weird edge case here? I’m afraid to silence some legit error but can’t think of a better way.

Fixes #427


hynek:
JFTR this is the final release blocker for 18.2. I will not release it until this is resolved.

PR #433, opened 2018-08-23
Make _Nothing a singleton

gappleto97:
I'm not sure how much the checklist makes sense for this change, given that there are no external or internal API changes.

This PR changes the `_Nothing` class to be a singleton. This allows a shorter class, as well as more reliance on inherited `object` code. Because they are all the same instance `_Nothing() is _Nothing()` will be `True`, and thus `_Nothing() == _Nothing()` will be `True` without intervention. This will result in a slight performance boost.

I have confirmed that this breaks no tests on my local machine under both 2.7 and 3.6

This is my first PR for the project, so I apologize in advance if there are any mistakes.

# Pull Request Check List

- [ ] Added **tests** for changed code


gappleto97:
Note that the linter errored out. It did not fail the code. If the reviewer could restart that job, that would be great.

hynek:
Thanks!

PR #438, opened 2018-08-28
Add ability to easily create singleton classes (alternate PR)

gappleto97:
This is a pull request for the alternate branch discussed in #434. It contains two potentially controversial commits. Both of these are related to adding the singleton feature to `attr.s`.

f4382adc0461167548f89f5cbe6d189ff99843d2 adds the `singleton` argument to `attr.s`, and has `frozen` and `slots` default to the same value as `singleton` (default False). It also adds a test for this.

157aee9b3833f3a7752dad888a13bb141da9a7bb adds documentation of these changes, without versionchanged/versionadded because I do not know the format to do this properly.

(Edit) 5c837d884a8628bbbe5a675e2cb330e665c17bda probably also needs to be redone, because the conflicts it resolves are largely derived from the two commits above

These commits may be excluded to avoid this potentially controversial change, though I would prefer that they are kept.

# Pull Request Check List


gappleto97:
As a note, I don't understand why it's breaking on pypy, and the last time I tried to set it up on my system it decided to overwrite python's pip. If any of you have a better suggestion for how to debug this, I would appreciate it.

src/attr/_make.py
```diff
@@ -900,12 +911,106 @@ def wrap(cls):
...
+                return class_listings[t_key]
+
+        else:
+            raise TypeError("Must be run on frozen attrs class or namedtuple")
+
+        def __init__(self, *args, **kwargs):
+            try:
+                if self in initialized:
+                    return
+                prev_init(self, *args, **kwargs)
```
gappleto97: I don't like that these lines are repeated, but I could not easily think of a better mechanism for this. I am open to suggestions on improving this mechanism.

tests/test_make.py
```diff
@@ -632,6 +641,154 @@ def test_sugar_callable(self):
...
+            "class C(NamedTuple):\n"
+            "    x: int = 3\n\n"
+            "a = C(4)\n"
+            "b = C(x=4)\n"
+            "c = C()\n"
+            "assert a is b\n"
+            "assert a is not c\n"
+            "assert a.x is b.x\n"
+            "assert a.x != c.x"
+        )  # avoid python2 syntax error
```
gappleto97: There are probably better ways to format this test, but I wanted to avoid a python2 syntax error. Again, open to suggestions for improvement

docs/api.rst
```diff
@@ -110,6 +110,14 @@ Core
       C2(x=42, y=[])
 
 
+.. autofunction:: attr.singleton
+
+    This is a decorator to build singleton classes.
+
+    It will usually not be accessed directly, but it also has support for
+    transforming a :class:`~collections.namedtuple` into a singleton.
```
gappleto97: I think this reference to `namedtuple` needs a sphinx plugin to point it at the python documentation. Not sure if that plugin is being used.

src/attr/_make.py
gappleto97: Maybe it would be good to have:

```python
def __init__(self, *args, **kwargs):
    try:
        if self not in initialized:
            raise AttributeError()
    except AttributeError:
        prev_init(self, *args, **kwargs)
        initialized.add(self)
```

gappleto97: That is slightly less efficient, but it's a one-time cost per object, and only really affects pypy (I think)

gappleto97:
@hynek if you don't mind, do you know the rough timeline for review on 18.3? I am more than willing to help if given direction, but I do not know how to proceed with getting this merged.

hynek:
I’m on a computer-less vacation until October so please hang on. :)

gappleto97:
Fair enough. Thanks for the fast response!

hynek:
Alright, stuff took a bit longer than I hoped, but I’m trying to get on top of FOSS again. Are you still game to bring this PR to its conclusion?

gappleto97:
I am! I will not have access to the codebase I needed it for for another couple weeks, but that should not be a problem. What all do you need on my end?

hynek:
Cool! I’m gonna say:

1. Please kill off the namedtuple stuff. This is not about preferences, this is purely about maintenance load. 😓
2. Rebase and make sure everything passes and the docs are up to date (I think I saw some artifacts from the original PR?)
3. Report back. :)

hynek:
I'm closing this due to lack of activity, feel free to open a new PR with the above points addressed, thanks!

PR #443, opened 2018-09-06
Fixes generics in mypy types

### LINKED ISSUE #441: Can we enable `disallow_any_generics` in mypy checks
Here is what mypy shows:
```
src/attr/__init__.pyi:26: error: Missing type parameters for generic type
src/attr/__init__.pyi:28: error: Missing type parameters for generic type
src/attr/__init__.pyi:32: error: Missing type parameters for generic type
src/attr/__init__.pyi:51: error: Missing type parameters for generic type
src/attr/__init__.pyi:60: error: Missing type parameters for generic type
src/attr/__init__.pyi:61: error: Missing type parameters for generic type
src/attr/__init__.pyi:62: error: Missing type parameters for generic type
src/attr/__init__.pyi:63: error: Missing type parameters for generic type
src/attr/__init__.pyi:108: error: Missing type parameters for generic type
src/attr/__init__.pyi:125: error: Missing type parameters for generic type
src/attr/__init__.pyi:142: error: Missing type parameters for generic type
src/attr/__init__.pyi:190: error: Missing type parameters for generic type
src/attr/__init__.pyi:191: error: Missing type parameters for generic type
src/attr/__init__.pyi:194: error: Missing type parameters for generic type
src/attr/__init__.pyi:236: error: Missing type parameters for generic type
src/attr/validators.pyi:8: error: Missing type parameters for generic type
src/attr/validators.pyi:9: error: Missing type parameters for generic type
src/attr/validators.pyi:11: error: Missing type parameters for generic type
src/attr/validators.pyi:12: error: Missing type parameters for generic type
src/attr/validators.pyi:13: error: Missing type parameters for generic type
src/attr/validators.pyi:14: error: Missing type parameters for generic type
src/attr/validators.pyi:26: error: Missing type parameters for generic type
src/attr/filters.pyi:4: error: Missing type parameters for generic type
src/attr/filters.pyi:5: error: Missing type parameters for generic type
src/attr/filters.pyi:28: error: Missing type parameters for generic type
```
Config is:
```
[mypy]
check_untyped_defs = True
disallow_any_generics = True
disallow_untyped_defs = True
follow_imports = silent
strict_optional = True
warn_redundant_casts = True
warn_unused_ignores = True
```
Mypy is 0.620

Can these be fixed or there is some limitation in current mypy (i.e. #421 )?

tailhook:
This removes errors in attrs library when using mypy setting: `disallow_any_generics = True`

Fixes #441 

Note: I'm not entirely sure that these types are correct as I'm both not very proficient with mypy generics and not familiar with attrs library itself.


hynek:
One of @chadrik @euresti or @ethanhs please. :)

src/attr/__init__.pyi
```diff
@@ -187,11 +187,11 @@ def attrs(
...
 # TODO: add support for returning NamedTuple from the mypy plugin
-class _Fields(Tuple[Attribute, ...]):
-    def __getattr__(self, name: str) -> Attribute: ...
+class _Fields(Tuple[Attribute[_T], ...]):
+    def __getattr__(self, name: str) -> Attribute[_T]: ...
 
-def fields(cls: type) -> _Fields: ...
-def fields_dict(cls: type) -> Dict[str, Attribute]: ...
+def fields(cls: type) -> _Fields[_T]: ...
+def fields_dict(cls: type) -> Dict[str, Attribute[_T]]: ...
```
chadrik: This section is wrong:  all of the fields are not the same type.

src/attr/__init__.pyi
```diff
@@ -223,7 +223,7 @@ def make_class(
 def asdict(
     inst: Any,
     recurse: bool = ...,
-    filter: Optional[_FilterType] = ...,
+    filter: Optional[_FilterType[_T]] = ...,
```
chadrik: I don't think there is any point to adding a TypeVar here if it is only used in one place within the signature.  The idea is to create a relationship to another argument, another method on the class (not applicable here), or the result.

chadrik:
I made a few comments, but don't have time to go too deep on this.  Long story short is that a lot of changes here are creating false relationships, or simply using `TypeVar` incorrectly.  I'm not familiar with `disallow_any_generics` option, but given that changing `Sequence` to `Sequence[Any]` seems to be solving the OPs problem, I would think that using `Any` everywhere would be the safest solution.

tailhook:
You're right. Changed most of them to Any, which is right for sequences. Only left `_ValidatorType` and `_FilterType` with type vars because as far as I can see their Attribute type is tied to the value passed to a callback.

And I'm keeping typevar in comparison operators. Are comparison operators expected to compare different types of attributes?

Let me know what you think.

tailhook:
*(it seems travis is failed by some transient network error, and codecov failure is also something not related to the actual changes)*

hynek:
Add a news fragment please before merging

tailhook:
Sorry, I'm not sure how to describe these changes.

hynek:
I tried to push a fragment against this repo but I utterly failed – dunno exactly at what. Gonna merge it and add something myself. CI is failing because of a bug in pytest 3.9.1.

PR #446, opened 2018-09-17
Added Core function parameter in func definition

saginadir:
Just added a parameter that was missing in the function defintion but existed below in the parameters section


hynek:
Thanks

Issue #212, opened 2017-06-15
Allow repr (+ str?) to take callables


Julian:
It would be useful to be able to pass in a callable to do `repr`'ing of attributes.

For me this happens if I want to present a value in the repr differently than just `repr(foo)`, which is sometimes useful, especially if the "different way" is to use e.g. `reprlib.repr` to do auto-truncation of long values.

Being able to pass in callables to the `repr` argument seems at first glance like a reasonably easy way to do that.

E.g.

`attr.ib(repr=reprlib.repr)`

and

`attr.ib(repr=lambda value: repr(value) * 12)`

You could preserve backwards compatibility by having True be equivalent to `repr`, and False be a sentinel (which you need anyways).

I think this could also satisfy #41, if you do the same to `attr.s`, and then provide an `attr.simple_repr`, so you'd do `attr.s(repr=attr.simple_repr)` and get the normal bracket repr instead of the eval one.

blueyed:
See https://github.com/python-attrs/attrs/issues/224#issuecomment-319078193 for a patch to support a callable with `repr`.

Tinche:
This is an interesting idea.

I think our repr support can be improved in any case; we're doing more work in the method that is generally needed, so it can be sped up by eval-ing a tailored function. Part of this work can be changing the `Attribute.repr` type to essentially be `Union[bool, Callable[[], str]` (so, either a boolean or a callable) and applying the callable when `__repr__`-ing.

The-Compiler:
Another use case where this is useful are numbers which you'd like to show in hexadecimal (`modifiers = attr.ib(repr=lambda mod: hex(int(mod)))`, or PyQt (not Python) enum values where you want to get a human-readable value instead of an int.

nvgoldin:
Another use case I have is when you have  a list attribute and you'd like `__repr__` to print out only the list length (for example in cases where it is extremely large).

Would be happy to know if there is another way to do it.

gimbo:
I've created a small extension which provides callable reprs and allows titles to be overridden too (e.g. for the length case cited above): [attrs_flexible_reprs.py](https://gist.github.com/gimbo/19f1fede170090fdf029a4f3f5c7fece)

Input/feedback welcomed.

hynek:
@gimbo I think your flexible repr doesn’t take into account our recursion handler?

***

Re earlier comments, I wouldn’t make this issue depend on transforming our repr generator to code generation. It would be nice to have, but not required for fixing this particular issue.

[...]

PR #453, opened 2018-10-20
Allow attributes to hide themselves in the repr() when they are the default

ssbr:
The use case is something like: while it is often useful to have the repr() be "complete", it is also often useful for it to be "minimal but unambiguous". If you have 6 attributes but all but 1 are the default value, the signal might be lost in the noise, especially when those attributes themselves are attrs-based classes and the whole tree of objects takes up a few hundred bytes of repr().

The way this was implemented was to change repr from a bool to an int, which can be 0 (never show a repr), 1 (always show a repr) or 2 (only show a repr if the attribute has a non-default value). I wasn't sure if I should properly add an enum for this or a new parameter or what, so rather than fiddling with it on my own I am mailing this out as an MVP and asking for feedback.

Re pull request checklist below:

* I cannot underestand tests/strategies.py or where I would add this.
* I don't know what version I am supposed to use for versionadded/versionchanged.
* I don't know what I'm supposed to put in changelog.d. It says to add a file with the PR#, but I haven't submitted the PR yet. I guess I will do this after I create the pull request.

# Pull Request Check List


hynek:
Two things:

One lingering topic we had in attrs is the move from booleans to Enums. Integers are a rather bad way to signal anything to APIs. So far I couldn’t make up my mind about whether to use enum and add a conditional enum34 dependency on Python 2 or some other shenanigans (but the closer 2020 comes, the more I think this is the way to go).

*However*, while your use case is interesting, it still a bit too specific and your approach lacks the flexibility to accommodate other needs. Have you seen #212? I think we should pursuit the `Union[bool, Callable]` approach  in this case for now because it gains us much more.

hynek:
> Sorry for delay in responding, stuff has been busy. 

Oh god don't apologize; I'm barely keeping myself ATM. Thank you for hanging in there!

I would say that you should just open a new one, to have a fresh discussion.

We could probably add something like `attr.DROP_IF_DEFAULT` 🤔

Julian:
So now that #212 is closed... maybe worth another look at this?

Though personally.. my first intuition is that this belongs in `attr.s`, not `attr.ib` [which we explicitly decided to push off on implementing in #212]. But maybe others feel differently.

(To speak the thought out more fully: I want to just write `@attr.s(repr=attrs.minimal_repr(repr=reprlib.repr))` once, not have to pass some enum thing into 5 different `attr.ib`s that I want to not show if they're at their default vals)

hynek:
Yes, my current plan is to make all the init, repr etc arguments optinally a callable that gets a tuple of `Attribute`s and returns 1–4 methods that get attached to the class.  It's so obvious and simple in hindsight, that it hurts.

I just need to work off some preparations (one of them was the eq/order split). I'm having a run ATM, so I hope I'll be able to finish that work in the next weeks (famous last words).

hynek:
That said, I'm gonna close this, because that will be much easier once we've done the work I've outlined above!

Sorry and thanks!

astrojuanlu:
This would have been a useful addition - similar to [the `print_changed_only` setting in scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.set_config.html).

> Yes, my current plan is to make all the init, repr etc arguments optinally a callable that gets a tuple of Attributes and returns 1–4 methods that get attached to the class. It's so obvious and simple in hindsight, that it hurts.

Is there any follow-up issue we can subscribe to?

hynek:
> Is there any follow-up issue we can subscribe to?

Unfortunately there currently isn't.

PR #498, opened 2019-02-09
Add Python 2 statement

hynek:
Partly as another preparation for `import attrs`, I think it's good to spell out our Python 2 policy explicitly.

Feedback welcome!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -60,6 +60,7 @@ Day-to-Day Usage
 
    license
    backward-compatibility
+   python-2
    contributing
    changelog
 
--- a/docs/python-2.rst
+++ b/docs/python-2.rst
@@ -1,0 +1,25 @@
+Python 2 Statement
+==================
+
+While ``attrs`` has always been a Python 3-first package, we the maintainers are aware that Python 2 will not magically disappear in 2020.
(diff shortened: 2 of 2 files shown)


itamarst:
I have far fewer users, so different situation... but:

1. You can release packages with version limiters, so like "version X runs on Python 2 and 3" and "version X+1 only runs on Python 3".
    Which means Python 2 users will continue to get version X even as you do X+1, X+2, X+3, etc..
2. You can then do point releases for Python 2 if there are critical bug fixes, so like X.1, X.2.

Which means you can just drop Python 2 from the main development branch.

For Eliot I'm asking for money beyond a certain point, though that's somewhat orthogonal (https://eliot.readthedocs.io/en/stable/python2.html).

hynek:
> Which means you can just drop Python 2 from the main development branch.

Since we do CalVer, I think a separate package would make more sense.

> For Eliot I'm asking for money beyond a certain point, though that's somewhat orthogonal

I might try that but my experience so far is that people don't like to give me money.

wsanchez:
I think the policy as written is sound.  Let's worry about how to handle dropping Py2 when we get to the point that it's necessary.  Options will depend on reasons for dropping.

PR #499, opened 2019-02-09
Use isolated builds

hynek:
OK it would be nice to use properly isolated builds but it seems like mypy is [breaking](https://travis-ci.org/python-attrs/attrs/jobs/490956881) with it.

Maybe someone can help with that?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools>=40.6.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [build-system]
 requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
 
 
 [tool.black]
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ matrix:
 
 
 install:
-  - pip install tox
+  - pip install --upgrade tox pip setuptools virtualenv
 
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ matrix:
 
 
 install:
-  - pip install --upgrade tox pip setuptools virtualenv
+  - pip install --upgrade tox
 
 
 script:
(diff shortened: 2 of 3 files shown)


pganssle:
I can't reproduce this failure locally on Arch Linux - same versions of Python and pip. Try restarting the build?

hynek:
It works for me locally too and I tried multiple times. I’ve even reverted it on master after multiple attempts. :-/

pganssle:
On Twitter, @gaborbernat suggested this could be due to pypa/setuptools#1506. Does seem like the same general problem, but I'm not sure why you'd be picking up an old version of `setuptools` in the isolated build environment, and I still can reproduce the problem even if I explicitly pin `setuptools` to an old version in the `build-system.requires`, which is extra puzzling.

pyproject.toml
```diff
@@ -1,5 +1,6 @@
 [build-system]
 requires = ["setuptools", "wheel"]
```
pganssle:
```suggestion
requires = ["setuptools>=40.6.0", "wheel"]
```

Per @gaborbernat, you actually are relying on a feature of a recent `setuptools` (only for the build isolation part - the non-isolated builds don't need to be pinned specially), so in your `build-system.requires`, you'll need a more recent version.

I'm still not sure why this is happening, but this might fix it.

bskinn:
I've found Travis's Python images sometimes have moderately outdated package versions bundled in.

Maybe mypy is out of date...what happens if you add a `pip install -U --force-reinstall mypy` to your `install:` section of `.travis.yml`?

pyproject.toml
hynek: looks like a [broad update](https://github.com/python-attrs/attrs/pull/499/commits/fc95d363d54f37c26fe2fb3fdf0048d03e633bd7) fixed it. let's try smaller now…

hynek: looks like we've got it! https://travis-ci.org/python-attrs/attrs/jobs/491005258

> I'm still not sure why this is happening, but this might fix it.

Wanna bet "caching"?

bskinn: I don't know if it's technically "caching" -- IIUC Travis uses VM images with setuptools and other packages already installed. So, unless there's a specific version specification in the build/install chain, or unless a `pip install -U setuptools` or a `pip install --force-reinstall {pkg with setuptools install dependency}` is invoked, the stale version never gets bumped.

hynek: I think Paul's point was that the _isolated env_ should have installed the latest version.

gaborbernat: yeah 😎 note isolation build are easily satisfy able, so no reason not to use a new builder version 😁in your requires part, when it's used is the same work to get version 30 as it's version latest

PR #500, opened 2019-02-10
Implement first class exception support

### LINKED ISSUE #368: Add a convenience flag for upcalling on Exception subclasses
Currently, these tests both fail on 2.7 and only the latter fails on 3.6.

```python
@attr.s
class E1(Exception):
    x = attr.ib()
    y = attr.ib()


def test_exception_init_called():
    e = E1(1, 2)
    assert e.args == (e.x, e.y) == (1, 2)


def test_exception_init_called_even_with_kw():
    e = E1(y=3, x=4)
    assert e.args == (e.x, e.y) == (4, 3)
```

If attrs is overriding `__init__` then this makes sense; `args` is filled in by `__init__` in 2.x and `__new__` in 3.x, but only pulling from the `*args` and never `**kwargs`. I discussed this a bit with @markrwilliams but I'm not sure what the best way to handle an attrs-ified exception would be. Maybe it's okay to consider `args` vestigal.

Mostly I'm raising this on the issue tracker because there are some references to using attrs for exceptions, but only indirectly. It'd be nice to have some documentation on a suggested approach either way.

hynek:
Fixes #368

This adds first class exception support to `attrs`. While on the first sight it might make more sense to have an argument like `exc=True` or even have a separate decorator, I think in the ligtht of `import attrs` this approach is better, because – of course – `auto_exc` would be True by default in `@attrs.define`.

Let me know what you think or whether I've missed some subtleties.

~Open Questions:~

~1. Is it a good idea to put attributes with defaults into args too?~
~2. Should we add `init=False, default=NOTHING` too?~

~The more I look at the code I feel like it should be all (including default=NOTHING) or nothing (only non-default).~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,8 @@ matrix:
       stage: docs
       env: TOXENV=docs
     - python: "3.7"
+      env: TOXENV=doctest
+    - python: "3.7"
       env: TOXENV=pypi-description
     - python: "3.7"
       env: TOXENV=changelog
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,6 @@ matrix:
       stage: docs
       env: TOXENV=docs
     - python: "3.7"
-      env: TOXENV=doctest
-    - python: "3.7"
       env: TOXENV=pypi-description
     - python: "3.7"
       env: TOXENV=changelog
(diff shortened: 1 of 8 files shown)


hynek:
Ouch, the second one is embarrassing. The first one…is there something we can do? 🤷‍♂️

rouge8:
For the first one, I add an `__attrs_post_init__` that calls `Exception.__init__(obj, *attr.astuple(obj))`: https://github.com/python-attrs/attrs/issues/368#issuecomment-385874857

hynek:
Well, this looks much nicer. What do you think?

That also resolves my original questions. Only init=True goes into it – period.

rouge8:
Nice! 🎉

PR #505, opened 2019-02-14
Improve documentation of safety constraints on hashable objects

### LINKED ISSUE #503: Requirement for safety of cache_hash is stronger than documented
The documentation states:

> `cache_hash` (`bool`) – Ensure that the object’s hash code is computed only once and stored on the object. If this is set to True, hashing must be either explicitly or implicitly enabled for this class. If the hash code is cached, then no attributes of this class which participate in hash code computation may be mutated after object creation.

A reader could read this as meaning that only the local fields of this object should not be assigned to and that using this on any `frozen` class is okay.  We should clarify that the object needs to be "deeply immutable" for this to be safe.

rgabbard:
Although #503 referred to `cache_hash`, on reflection it occurred to me that the concern is valid for all hashable objects.  At the same time, I've clarified the documentation to be clear that if you mutate a `cache_hash` object or an object referred to in its hash computation, the resulting object hash code is undefined.  This seemed better than making any more particular comment since it leaves the API uncommitted as to whether the hash code is computed at object-construction time or lazily on the first call to `hash`

This closes #503 .  

# Pull Request Check List

- [X] Added **tests** for changed code. *(n/a, docs only)*
- [X ] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/strategies.py). *(n/a, docs only)*
- [X] Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``). *(n/a, docs only)*
    - [X] ...and used in the stub test file `tests/typing_example.py`. *(n/a, docs only)*
    - [X ] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).
- [X] Documentation in `.rst` files is written using [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).


rgabbard:
@hynek : Ready for review.

hynek:
thanks!

PR #506, opened 2019-02-18
added first doc stub

chobeat:


hynek:
So the first steps would be:

1. update your branch to latest master (you can press the button above and pull)
2. make sure the CI passes. :) you can usually make that sure locally by running tox.

You may want to give another look to our [contributing help file](http://www.attrs.org/en/stable/contributing.html) since it has a bunch of valuable information that will save you a lot of time!

chobeat:
@hynek now the test passes but it's trivial stuff so far. From the other thread on the issue we left open the discussion on how to proceed from here.

tests/test_make.py
```diff
@@ -1587,3 +1587,13 @@ class B(A):
             "Comparision of subclasses using %s is deprecated and will be "
             "removed in 2019." % (op,)
         ) == w.message.args[0]
+
+
+class TestDocs:
+    def test_init_docs(self):
```
hynek: Please add a docstring that describes what you're testing. [This article](https://jml.io/pages/test-docstrings.html) will show you our best practices in a nice example.

src/attr/_make.py
```diff
@@ -720,6 +720,13 @@ def _add_method_dunders(self, method):
         except AttributeError:
             pass
 
+        try:
+            method.__doc__ = "Method generated by attrs for {}".format(
```
hynek: Please add a period at the end of the sentence.

tests/test_make.py
```diff
@@ -1587,3 +1587,13 @@ class B(A):
             "Comparision of subclasses using %s is deprecated and will be "
             "removed in 2019." % (op,)
         ) == w.message.args[0]
+
+
+class TestDocs:
```
hynek: Why do you think we need a new test class for this?

tests/test_make.py
```diff
@@ -1587,3 +1587,13 @@ class B(A):
...
+    def test_init_docs(self):
+        @attr.s
+        class A(object):
+            pass
+
+        if hasattr(A, "__qualname__"):
```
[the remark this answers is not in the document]
chobeat: I check because otherwise it breaks the test in python 2. The behavior for python 2 is not specified yet and I would develop it on top.

tests/test_make.py
```diff
@@ -1587,3 +1587,13 @@ class B(A):
...
+            assert A.__qualname__ in A.__init__.__doc__
```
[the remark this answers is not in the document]
chobeat: Alright

tests/test_make.py
chobeat: I wasn't sure on the policy you use to separate test classes but usually I create test classes around a concern (in this case the docs). If at the end of the PR I see that there won't be enough tests to justify a dedicated class, I will add them to an existing class.

chobeat:
I've added the newsfragment and I've addressed most of the points. When you answer in the open conversations I can proceed.

hynek:
Sorry I dropped the ball on this, could you resolve the conflicts please?

hynek:
Thanks and sorry for the delays!

PR #511, opened 2019-03-03
Support on azure-pipelines.yml

tonybaloney:
Starting a WIP PR just so I can see the build status

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,7 +18,7 @@ jobs:
       - script: python -m pip install --upgrade tox codecov
         displayName: install tox
 
-      - script: tox -e py27,py34,py35,py36,py37 --junitxml=report.xml
+      - script: tox -e py27,py34,py35,py36,py37
         displayName: run tox for Python runtimes
       
       - script: tox -e typing
@@ -34,12 +34,6 @@ jobs:
           coverage combine
           codecov -t $(CODECOV_TOKEN)
         displayName: Report Coverage
-
-      - task: PublishTestResults@2
-        inputs:
-          testResultsFiles: 'report.xml'
-          testRunTitle: '$(Agent.OS) - $(Build.DefinitionName) - Python $(python.version)'
-        condition: succeededOrFailed()
 
   - job: Build_Packages
     dependsOn: 'Test'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,8 +7,21 @@ jobs:
       vmImage: 'Ubuntu-16.04'
     strategy:
       matrix:
+        Python27:
+          python.version: '2.7'
+          tox.target: 'py27'
+        Python34:
+          python.version: '3.4'
+          tox.target: 'py34'
+        Python35:
+          python.version: '3.5'
+          tox.target: 'py35'
+        Python36:
+          python.version: '3.6'
+          tox.target: 'py36'
         Python37:
           python.version: '3.7'
+          tox.target: 'py37'
       maxParallel: 2
     steps:
       - task: UsePythonVersion@0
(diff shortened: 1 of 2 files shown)


tonybaloney:
@hynek once you change https://github.com/python-attrs/attrs/pull/511/files#diff-fec826feae04e51c0d94076385408bdcR42 with the actual token this build _should_ work :-)

tonybaloney:
Also, you might want to make `pytest-azurepipelines` optional in tox.ini, since the regular user doesn't need it

hynek:
> @hynek once you change https://github.com/python-attrs/attrs/pull/511/files#diff-fec826feae04e51c0d94076385408bdcR42 with the actual token this build _should_ work :-)

I presume you mean the CODECOV_TOKEN? that's a secret variable in Azure Pipelines; is it not available if someone else runs it? I guess that makes sense, since you could open a PR that just sends a token to yourself? What is the correct way here? I don't want to put it there in cleartext?

The build fails because of [trailing whitespace](https://attrs.visualstudio.com/attrs/_build/results?buildId=33&view=logs&jobId=4902a98b-b56a-5166-3cb6-3d0b3604bae7&taskId=3f5976e1-c73b-5e3a-f2a1-188eab9960a2&lineStart=47&lineEnd=47&colStart=1&colEnd=27) btw.

tonybaloney:
Ah,  that makes more sense if the variable is scoped to only contributors. In that case, you can keep my changes to the codecov job or scrap them. Fixed the trailing whitespace

azure-pipelines.yml
```diff
@@ -7,59 +7,61 @@ jobs:
...
-        condition: succeededOrFailed()
+          versionSpec: '$(python.version)'
+          architecture: 'x64'
+      - script: python -m pip install --upgrade tox codecov
+        displayName: install tox
+
+      - script: tox -e typing
+        displayName: run tox for typing
+
+      - script: tox -e lint,docs
```
hynek: any reason why you grouped these together?

hynek:
> Ah, that makes more sense if the variable is scoped to only contributors. In that case, you can keep my changes to the codecov job or scrap them. Fixed the trailing whitespace

Yeah but that's kinda bad, because contributors won't get feedback on coverage. :|

I believe I saw something that AP might have their own solution to coverage? Do you know anything about it? Is it ready?

tonybaloney:
I have a solution for coverage gap. Bear with me 🐻

hynek:
giphy

tonybaloney:
@hynek checkout https://attrs.visualstudio.com/attrs/_build/results?buildId=37&view=results and see the summary tab you'll see ~70% coverage, you can download the coverage report ZIP file in the summary page as well, and click on lines to see details.

hynek:
We’ve got 100% coverage though. Is this coverage from only one particular run (maybe just py2 or just py3)?

hynek:
Yeah something is going very wrong:

image

Note the py34, but even that isn't correct. But this might be just us using pytest-cov wrong? I _think_ there was a way to achieve https://hynek.me/articles/testing-packaging/ w/ pytest-cov? 🤔

hynek:
Ah so one think I just noticed: we have special envs for 2.7 and 3.7 because that determines our coverage. The rest of our tox envs used to run w/o coverage for performance reasons (especially PyPy).

With your changes you run the tests that used to _not_ run with coverage with pytest-cov and you've left the coverage-envs using the old way that probably gets ignored?

Is it even possible to combine coverages over multiple runs into AP coverage?

P.S. thanks for all your efforts!

hynek:
thanks for everything, we're on AP now! I think/hope I got all missing things from your PR over time. :)

PR #536, opened 2019-05-30
Improve the error message informing the value that it got it

williamjamir:
I would like to propose a change on the error message returned by the validator `is_callable()`.

Currently, only a message "TypeError: 'x' must be callable" it's informed to the user, I would like to add the value that was passed to it within the message.

This approach besides given more information to the user resembles the errors messages returned by others validators that also include the value informed =)

# Pull Request Check List

- [X] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/strategies.py).
    - [X] ...and used in the stub test file `tests/typing_example.py`.


hynek:
Hm I wish it would work like that but I'm afraid that this is a breaking change. :(  Maybe make it an option?

williamjamir:
Hi hynek, thanks for your time review it :)

> Hm I wish it would work like that but I'm afraid that this is a breaking change. :(
 
Oh sorry, I didn't consider that users would be relying/depending on the errors message raised by the validators.

> Maybe make it an option?

Sure, do you have anything in mind? I mean, where would I introduce this option?

wsanchez:
If we want people to distinguish this `TypeError` from others without looking at string values, and the value of the thing-that-isn't-callable is useful, then that should all be made available without having to parse a string.

So perhaps a new exception type `NotCallableError`, subclassing `TypeError`, which has a `message` attribute, a `value` attribute, and make sure that it's `__str__` produces the same not-informative-enough message as before?

hynek:
What @wsanchez said SGTM!

williamjamir:
Hi @wsanchez, I not sure if I got all right.

> If we want people to distinguish this `TypeError` from others without looking at string values, and the value of the thing-that-isn't-callable is useful, then that should all be made available without having to parse a string.

Yep, I totally agree =) 

> So perhaps a new exception type `NotCallableError`, subclassing `TypeError`, which has a `message` attribute, a `value` attribute, and make sure that it's `__str__` produces the same not-informative-enough message as before?

Right, I agree with adding this new exception (and I can add in the next commit without any problem), but the "original issue" that I brought was not solved, the message still not-informative and doesn't resemble the errors messages returned by others validators.

So just to make sure that I got right, we are going to keep with the not-informative-enough message in order to avoid breaking changes right?

wsanchez:
Yeah, I think @hynek is saying that the current string can't change for compatibility, so adding a new type that at least can get you the data you need is an improvement.

I agree that the default string could be better, and I'm less sympathetic to compatibility with code that's inspecting the string, but `attrs` has a high compatibility bar.  (I'm also not sure where an option to enable an improved message would go.)

hynek:
Firstly, having a dedicated class with a `got` field (do have have precedents on how to call fields like that?) allows you to write tests that don't break on changes. I guess I can live with changing the string in the end however if we do it, we should also add a good way to introspect the error.

src/attr/exceptions.py
```diff
@@ -55,3 +55,20 @@ class PythonTooOldError(RuntimeError):
...
     .. versionadded:: 18.2.0
     """
+
+
+class NotCallableError(TypeError):
+    """
+    A ``attr.ib()`` requiring a callable has been set with a value
+    that is not callable.
+
+    .. versionadded:: 19.2.0
```
williamjamir: Questions =)

1) Is this 'versionadded' supposed to be the next version?
2) Does it need a changelog entry?

hynek: yes and yes!

tests/test_exceptions.py
```diff
@@ -0,0 +1,9 @@
+class TestNotCallableError:
```
hynek: I don't think it makes sense to open a new file (or class for that matter) here. Just put it along with the callable tests.

src/attr/exceptions.pyi
```diff
@@ -1,7 +1,14 @@
...
 class FrozenInstanceError(AttributeError):
     msg: str = ...
 
 class AttrsAttributeNotFoundError(ValueError): ...
 class NotAnAttrsClassError(ValueError): ...
 class DefaultAlreadySetError(RuntimeError): ...
 class UnannotatedAttributeError(RuntimeError): ...
+
+class NotCallableError(TypeError):
+    message: str = ...
```
hynek: Since `FrozenInstanceError` already is using `msg`, we should stay consistent here.

tests/test_validators.py
```diff
@@ -354,8 +354,14 @@ def test_noncallable_validators(
         """
         with pytest.raises(TypeError) as e:
             deep_iterable(member_validator, iterable_validator)
-
-        e.match(r"\w* must be callable")
+        value = 42
+        message = "must be callable (got {value} that is a {type_}).".format(
+            value=value, type_=value.__class__
+        )
```
hynek: An empty line before `assert` blocks please.

tests/test_validators.py
```diff
@@ -466,7 +472,14 @@ def test_noncallable_validators(
         with pytest.raises(TypeError) as e:
             deep_mapping(key_validator, value_validator, mapping_validator)
 
-        e.match(r"\w* must be callable")
+        value = 42
+        message = "must be callable (got {value} that is a {type_}).".format(
+            value=value, type_=value.__class__
+        )
```
hynek: An empty line before `assert` blocks please.

tests/test_validators.py
```diff
@@ -550,7 +563,13 @@ def test_fail(self):
         with pytest.raises(TypeError) as e:
             v(None, a, None)
 
-        e.match("'test' must be callable")
+        value = None
+        message = "'test' must be callable (got {value} that is a {type_})."
+        expected_message = message.format(value=value, type_=value.__class__)
```
hynek: An empty line before `assert` blocks please.

changelog.d/536.change.rst
```diff
@@ -0,0 +1,2 @@
+``is_callable()`` validator now raises a ``NotCallableError`` exception informing the value and type received.
```
hynek: 1. We use fully qualified names in the changelog (both is_callable and NotCallableError please)
2. We should stress that `NotCallableError` is a subclass of `TypeError`.

williamjamir:
Is there any further change in order to merge this pull request?

hynek:
All good, thank you for your patience and your contribution!

williamjamir:
Thank you, I'm glad that I could I contribute with this project somehow.=)

hynek:
I hope my lagginess doesn't discourage you from doing it more often. This year was awesome, but super stressful. 🙈

PR #540, opened 2019-06-04
[WIP] fix coverage reporting on 3.8

hynek:
Since the python is installed differently, there seem to be problems finding it.

Codecov itself runs coverage from $PATH.


hynek:
OK, 3.8 is at this point wayyy too broken still.

PR #544, opened 2019-06-16
Rename remaining validator/default decorator examples with unclear names

obestwalter:
The documentation already explains well why validator or default decorated methods
must not have the name of existing attributes (because they would be overridden).

This replaces the last remaining examples not using the explicit explanatory name.


hynek:
Thanks!

PR #552, opened 2019-07-12
adding matches_re validator

kurtbrose:
still needs more work just getting the ball rolling

# Pull Request Check List

- [*] Added **tests** for changed code.


src/attr/validators.py
```diff
@@ -63,6 +66,63 @@ def instance_of(type):
...
+class _MatchesReValidator(object):
+    regex = attrib()
+    flags = attrib()
+    func = attrib()
+
+    def __call__(self, inst, attr, value):
+        """
+        We use a callable class to be able to change the ``__repr__``.
+        """
+        if self.func is re.match:
```
hynek: We should keep the call overhead as low as possible, so I would suggest to pass bound methods into this calls instead of a regex and a func and the dispatching.

src/attr/validators.py
```diff
@@ -63,6 +66,63 @@ def instance_of(type):
...
+
+def matches_re(regex, flags=0, func=_native_fullmatch):
+    """
+    A validator that raises :exc:`ValueError` if the initializer is called
+    with a string that doesn't match the given regex, and :exc:`TypeError`
+    if the initializer is called with a non-string.
+
+    :param regex: a regex string or compiled regex to match against
+    :param flags: flags that will be passed to the underlying re function (default 0)
+    :param func: which underlying re function to call (options are `re.fullmatch()`, `re.search()`, `re.match()`)
```
hynek: Apart from what flake8 is yelling at you, please also add the default and a `.. versionadded::` tag.

src/attr/validators.py
```diff
@@ -63,6 +66,63 @@ def instance_of(type):
...
+    """
+    regex_attr, flags_attr, func_attr = _MatchesReValidator.__attrs_attrs__
+    instance_of(str)(None, regex_attr, regex)
+    in_((_native_fullmatch, re.search, re.match))(None, func_attr, func)
+    if func is None:
+        # python 2 fullmatch emulation
+        regex = r"(?:{})\Z".format(regex)
+    regex = re.compile(regex, flags)
+    if func is None:
```
hynek: I would suggest we make the default `None` and then do a

```python
if func is None:
    if PY2:
        ...
    else:
        func = re.fullmatch
```

That would simplify the code a lot.

tests/test_validators.py
```diff
@@ -78,6 +81,38 @@ def test_repr(self):
...
+    Tests for `matches_re`.
+    """
+
+    def test_in_all(self):
+        """
+        Verify that this validator is in ``__all__``.
+        """
+        assert matches_re.__name__ in validator_module.__all__
+
+    def test(self):
```
hynek: Please add a [test docstring](https://jml.io/pages/test-docstrings.html) and it'd also make sense to break the test up a little bit.

src/attr/validators.py
```diff
@@ -63,6 +66,63 @@ def instance_of(type):
...
+    with a string that doesn't match the given regex, and :exc:`TypeError`
+    if the initializer is called with a non-string.
+
+    :param regex: a regex string or compiled regex to match against
+    :param flags: flags that will be passed to the underlying re function (default 0)
+    :param func: which underlying re function to call (options are `re.fullmatch()`, `re.search()`, `re.match()`)
+    """
+    regex_attr, flags_attr, func_attr = _MatchesReValidator.__attrs_attrs__
+    instance_of(str)(None, regex_attr, regex)
+    in_((_native_fullmatch, re.search, re.match))(None, func_attr, func)
```
hynek: I don't think you have to validate this extensively since we usually don't do that ourselves. Most of the things will explode while constructing anyways.

kurtbrose:
thanks for the comments, will address soon

sorry for the slowness, I've been a bit hung up on getting the local dev environment set up, I keep running into this:

```
> py.exe -m pip install -e .
Obtaining file:///C:/Users/Kurt/workspace/attrs
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  Complete output from command C:\Users\Kurt\AppData\Local\Programs\Python\Python37-32\python.exe C:\Users\Kurt\AppData\Local\Programs\Python\Python37-32\lib\site-packages\pip\_vendor\pep517\_in_process.py get_requires_for_build_wheel C:\Users\Kurt\AppData\Local\Temp\tmp2j45vhxz:
    File "C:\Users\Kurt\AppData\Local\Programs\Python\Python37-32\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 207, in <module>
      main()
    File "C:\Users\Kurt\AppData\Local\Programs\Python\Python37-32\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 197, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "C:\Users\Kurt\AppData\Local\Programs\Python\Python37-32\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 54, in get_requires_for_build_wheel
      return hook(config_settings)
    File "C:\Users\Kurt\AppData\Local\Temp\pip-build-env-m4ck65jz\overlay\Lib\site-packages\setuptools\build_meta.py", line 145, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "C:\Users\Kurt\AppData\Local\Temp\pip-build-env-m4ck65jz\overlay\Lib\site-packages\setuptools\build_meta.py", line 126, in _get_build_requires
      self.run_setup()
    File "C:\Users\Kurt\AppData\Local\Temp\pip-build-env-m4ck65jz\overlay\Lib\site-packages\setuptools\build_meta.py", line 141, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 99, in <module>
      + read("AUTHORS.rst")
  AttributeError: 'NoneType' object has no attribute 'group'
```

hynek:
I wasn’t aware of that problem until recently. Could you help my understand how it happens in #561?

hynek:
I believe pip install -dev on windows should work now?

kurtbrose:
yes!  worked like a charm after update, thanks!

src/attr/validators.py
kurtbrose: oh good idea -- this logic could be done in the constructor instead of at every call

src/attr/validators.py
kurtbrose: I think I see what you mean, cleaned it up a bit

kurtbrose:
just an FYI, pre-commit still fails with an SSL failure trying to fetch black from github

Errors:
    fatal: unable to access 'https://github.com/ambv/black/': error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version

I've checked the code is flake8 clean, hopefully I'm not missing anything else format wise

src/attr/validators.py
kurtbrose: 👍

tests/test_validators.py
kurtbrose: broke it into two pieces with more semantic names and docstrings

kurtbrose: I really like the guidelines in that link you provided -- especially dropping the do-nothing words like "check this" or "test that"

along those lines though, the word "Tests" and "Verify" each occur 9x in docstrings in test_validators.py

should I follow the pattern, or not add any more instances of "tests" or "verify" in docstrings?

src/attr/validators.py
kurtbrose: yea, even as I wrote this it seemed like it was being kind of cute -- the main purpose is to get the nicely formatted `"'func' must be one of (...), got (...)"`

probably the correct thing to do is raise it's own exception

(I hear you about the constructor exploding downstream anyway  -- but I think where possible putting the fix right in the error message "received X, but only A, B or C are valid" leads to a better user experience; you never know when the tiny bit of frustration will be the last straw that causes someone to get stuck or give up)

hynek:
> just an FYI, pre-commit still fails with an SSL failure trying to fetch black from github
> 
> Errors:
> fatal: unable to access 'https://github.com/ambv/black/': error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version

Hm, that must be some local problem. Under what Python version is your pre-commit running? It must be something ancient, if the ssl is too old to connect to GitHub?! Anthony is AFAIK using Windows too, so that shouldn't be the issue. The tox env uses explicitly 3.7 tho?

(I've took this opportunity to fix black's URL tho)

> I've checked the code is flake8 clean, hopefully I'm not missing anything else format wise

You did: https://attrs.visualstudio.com/attrs/_build/results?buildId=231&view=logs&jobId=810e27dd-b084-5a0b-5c49-7fe688ece5ee&taskId=2dbb6c1a-2c14-506e-d517-3d9e1024c6b7&lineStart=39&lineEnd=40&colStart=1&colEnd=1 😇

hynek:
Hey an release is imminent, do you think you can bang it into shape on short notice or do you want to push this to the next release?

kurtbrose:
let me give it one more push this weekend and see if I can resolve everything -- sorry for being so flakey!

kurtbrose:
alright I think this is ready to go now ... only thing is I'm not sure about hypothesis or the type testing area

not sure if those apply since this is more of a "leaf" than structural

PR #557, opened 2019-07-23
Replace black link

serhii73:
I changed black link in readme, because black moved it under the PSF umbrella and link changed.
https://twitter.com/llanga/status/1153417050024550401


hynek:
Thanks!

PR #560, opened 2019-07-30
Make unique_filename not depend on repr.

### LINKED ISSUE #558: init unique_filename is problematic for exception aggregation
We had a case where a custom converter was often failing due to an app issue.  The traceback will include a line with a unique ID based on the initial attr values in the class:

```
File "<attrs generated init 81db48c8801f50d1c320ad1b8d1e0028320432b2>", line 3, in __init__
```

Having a unique ID in the traceback in turn causes problems for exception fingerprinting (e.g. by Sentry).  I'm not thrilled about doing custom traceback filtering for such cases, and would like to explore other options.

I wonder what the full rationale is for ensuring the unique filename, and whether it could be made optional.

Including a hash of the attribute values is essentially like dumping the values of all attributes of a class instance as part of the traceback whenever there is a constructor error regarding a single attribute-- which is less than precise error reporting.

https://github.com/python-attrs/attrs/blob/dc1b5a01e98a9ba49d45ad07ba6ae0ee2d9c1b8e/src/attr/_make.py#L1268

euresti:
For debugging purposes, when a method is created it needs to be put
into a "filename".  This filename was generated using `repr(attrs)`
however this meant that the filename could change every time the
program was loaded depending on the whims of repr.  (In particular
free functions and lambdas would include ids)

This solves the problem by changing the name to be <attrs generated
{method} {module}.{qualname}> If it should happen that the name is
taken then -2 will be appended and incremented until a free filename
is found.

Fixes #558

I'll wait on a reply as to whether this is more or less what we were looking for before I go through the Pull Request checklist.

# Pull Request Check List

### DIFF
--- a/changelog.d/560.change.rst
+++ b/changelog.d/560.change.rst
@@ -1,0 +1,2 @@
+The fake filename for generated methods is now more stable.
+It won't change when you restart the process.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1,10 +1,10 @@
 from __future__ import absolute_import, division, print_function
 
 import copy
-import hashlib
 import linecache
 import sys
 import threading
+import uuid
 import warnings
 
 from operator import itemgetter
@@ -665,7 +665,10 @@ class _ClassBuilder(object):
     def add_hash(self):
         self._cls_dict["__hash__"] = self._add_method_dunders(
             _make_hash(
-                self._attrs, frozen=self._frozen, cache_hash=self._cache_hash
+                self._cls,
+                self._attrs,
+                frozen=self._frozen,
+                cache_hash=self._cache_hash,
             )
         )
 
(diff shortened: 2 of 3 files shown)


euresti:
I just tested on some code and this is what the traceback looks like.  This is a file with 2 classes called `Foo` just to show the numbering scheme.

```
  File "foo.py", line 24, in <module>
    Foo()
  File "<attrs generated init __main__.Foo-2>", line 2, in __init__
    self.__attrs_post_init__()
  File "foo.py", line 20, in __attrs_post_init__
    raise Exception("Foo")
Exception: Foo

  File "foo.py", line 30, in <module>
    X(6)
  File "<attrs generated init __main__.Foo>", line 3, in __init__
    self.__attrs_post_init__()
  File "foo.py", line 12, in __attrs_post_init__
    raise Exception("Foo")
Exception: Foo
```

hynek:
Would you mind adding a test for that (substring check is fine)?

And a newsfragment please. 😇

euresti:
All right.  Let me know if something else is needed.

hynek:
Thanks you so much David, this looks great!

PR #563, opened 2019-08-18
Make auto_exc=True classes hashable by ID

### LINKED ISSUE #543: auto_exc's hashing behavior and its documentation do not match
The [documentation](https://github.com/python-attrs/attrs/blob/25a02bbc7b5309acebda20101125372d42c978db/src/attr/_make.py#L846) for `auto_exc` says:
```
  - the values for *cmp* and *hash* are ignored and the instances compare
          and hash by the instance's ids (N.B. ``attrs`` will *not* remove
          existing implementations of ``__hash__`` or the equality methods. It
          just won't add own ones.),
```

However, the test for `auto_exc` [says](https://github.com/python-attrs/attrs/blob/25a02bbc7b5309acebda20101125372d42c978db/tests/test_dark_magic.py#L519):
```
    Classes with auto_exc=True have a Exception-style __str__, are neither
        comparable nor hashable, and store the fields additionally in
        self.args.
```

and [tests](https://github.com/python-attrs/attrs/blob/25a02bbc7b5309acebda20101125372d42c978db/tests/test_dark_magic.py#L558):

```
            with pytest.raises(TypeError):
                hash(e)
```

The documentation and the code should be made consistent.  I noticed this because `unittest` complained that the exception type was unhashable when thrown from a test and I see some [references to `logging` expecting hashable exceptions](https://github.com/schematics/schematics/issues/452), so the documented behavior is probably better than the implemented behavior.

hynek:
They now behave as the docs claim.

Fixes #543

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -913,7 +913,8 @@ def attrs(
                 "Invalid value for hash.  Must be True, False, or None."
             )
         elif hash is False or (hash is None and cmp is False) or is_exc:
-            # Hash by ID aka NOP.
+            # Don't do anything. Should fall back to __object__'s __hash__
+            # which is by id.
             if cache_hash:
                 raise TypeError(
                     "Invalid value for cache_hash.  To use hash caching,"
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -912,20 +912,19 @@ def attrs(
             raise TypeError(
                 "Invalid value for hash.  Must be True, False, or None."
             )
-        elif hash is False or (hash is None and cmp is False):
+        elif hash is False or (hash is None and cmp is False) or is_exc:
+            # Hash by ID aka NOP.
             if cache_hash:
                 raise TypeError(
                     "Invalid value for cache_hash.  To use hash caching,"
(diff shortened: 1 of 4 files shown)


src/attr/_make.py
```diff
@@ -912,20 +912,19 @@ def wrap(cls):
             raise TypeError(
                 "Invalid value for hash.  Must be True, False, or None."
             )
-        elif hash is False or (hash is None and cmp is False):
+        elif hash is False or (hash is None and cmp is False) or is_exc:
+            # Hash by ID aka NOP.
```
rgabbard: A minor quibble about the comment: since this leaves `__hash__` untouched, wouldn't the behavior here be "Use any `__hash__` provided by the user; otherwise fall back to using the `__hash__` of the base class.  If the base class is `object`, this is hashing by ID."?  The case of a user-supplied hash is slightly unclear in [the documentation](https://github.com/python-attrs/attrs/blob/8174b03b962165278dd65508bf0f947c98a1190a/src/attr/_make.py#L773) as well.

hynek:
Thanks, I've clarified it a bit.

PR #568, opened 2019-09-05
support custom repr() callable for attributes

### LINKED ISSUE #212: Allow repr (+ str?) to take callables

### LINKED ISSUE #567: Feature Request: Custom __repr__ for particular attribute?
Reading the [docs](http://www.attrs.org/en/stable/api.html#attr.ib), it seems that there's no ability to do a custom repr for a particular attribute.  E.g., It would be nice to be able to specify a function for  `repr` like:

```
@attr.s 
class Datapoint:
    image = attr.ib(type=np.ndarray, repr = lambda x: '<array of shape {} and dtype {}>'.format(x.shape, x.dtype))
    label = attr.ib(type=int)
```
so that when you `print(Datapoint(img, lab))`, it does not display numpy's awkward default formatting.  

As far as I see you can currently only set a boolean flag `repr=False` to disable to whole attribute from being shown in the object's repr, which is not what I want.  We could still maintain this behaviour if we just say `repr: Optional[Union[bool, Callable[[Any], str]]]=True`

wbolster:
This adds support for using a custom callable to be used instead of the default `repr()` when formatting individual attributes in the generated `__repr()__`. The `attr.ib(repr=…)` argument now takes either a `bool` (as before), or a callable that takes a value and returns a string.

Fixes #567. Fixes #212.

### DIFF
--- a/changelog.d/568.change.rst
+++ b/changelog.d/568.change.rst
@@ -1,1 +1,1 @@
-The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a custom callable, which makes it possible to influence how the attribute value is formatted by the generated ``__repr__()_`` method.
+The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a custom callable, which makes it possible to influence how the attribute value is formatted by the generated ``__repr__()`` method.
--- a/changelog.d/568.change.rst
+++ b/changelog.d/568.change.rst
@@ -1,1 +1,2 @@
-The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a custom callable, which makes it possible to influence how the attribute value is formatted by the generated ``__repr__()`` method.
+The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a callable.
+That callable must return a string and is then used for formatting the attribute by the generated ``__repr__()`` method.
--- a/changelog.d/568.change.rst
+++ b/changelog.d/568.change.rst
@@ -1,0 +1,1 @@
+The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a custom callable, which makes it possible to influence how the attribute value is formatted by the generated ``__repr__()_`` method.
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -656,7 +656,7 @@ It will get called at the end of the generated ``__init__`` method.
    >>> obj
    C(x=1, y=2, z=3)
 
-Finally, you can exclude single attributes from certain methods:
+You can exclude single attributes from certain methods:
 
 .. doctest::
 
@@ -665,4 +665,15 @@ Finally, you can exclude single attributes from certain methods:
    ...     user = attr.ib()
    ...     password = attr.ib(repr=False)
    >>> C("me", "s3kr3t")
-   C(user='me')
+   C(user='me', password=***)
+
+Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify custom callable to be used instead of the `repr()` built-in function:
+
+.. doctest::
+
+   >>> @attr.s
+   ... class C(object):
+   ...     user = attr.ib()
+   ...     password = attr.ib(repr=lambda value: '***')
+   >>> C("me", "s3kr3t")
+   C(user='me', password=***)
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -667,7 +667,7 @@ You can exclude single attributes from certain methods:
    >>> C("me", "s3kr3t")
    C(user='me')
 
-Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify a custom callable to be used instead of the `repr()` built-in function:
+Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify a custom callable to be used instead of the ``repr()`` built-in function:
 
 .. doctest::
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -665,7 +665,7 @@ You can exclude single attributes from certain methods:
    ...     user = attr.ib()
    ...     password = attr.ib(repr=False)
    >>> C("me", "s3kr3t")
-   C(user='me', password=***)
+   C(user='me')
 
 Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify custom callable to be used instead of the `repr()` built-in function:
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -667,7 +667,7 @@ You can exclude single attributes from certain methods:
    >>> C("me", "s3kr3t")
    C(user='me')
 
-Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify custom callable to be used instead of the `repr()` built-in function:
+Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify a custom callable to be used instead of the `repr()` built-in function:
 
 .. doctest::
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1221,9 +1221,8 @@ def _make_repr(attrs, ns):
     name.
     """
 
-    # Figure out which attributes should be included in the class's repr(),
-    # and which function should be used to format those attributes.
-    # The a.repr attribute can be either a bool or a custom callable.
+    # Figure out which attributes to include, and which function to use to
+    # format them. The a.repr value can be either bool or a custom callable.
     attr_names_with_reprs = tuple(
         (a.name, repr if a.repr is True else a.repr)
         for a in attrs
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1220,6 +1220,15 @@ def _make_repr(attrs, ns):
     name.
     """
 
+    # Figure out which attributes should be included in the class's repr(),
+    # and which function should be used to format those attributes.
+    # The a.repr attribute can be either a bool or a custom callable.
+    attr_names_with_reprs = tuple(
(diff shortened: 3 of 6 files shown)


wbolster:
not sure  why the doctest fails. in an interactive shell it works:

```python
>>> import attr
>>> @attr.s
... class C(object):
...     user = attr.ib()
...     password = attr.ib(repr=lambda value: '***')
... 
>>> C("me", "s3kr3t")
C(user='me', password=***)
```

hynek:
The doctests says `   ...     password = attr.ib(repr=False)`.

hynek:
@Julian is this good enough for your – I'm afraid ancient – #212?

Julian:
Cool. Other than https://github.com/python-attrs/attrs/issues/567#issuecomment-528773554, seems likely... Can review the patch more carefully in a bit.

changelog.d/568.change.rst
```diff
@@ -0,0 +1 @@
+The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a custom callable, which makes it possible to influence how the attribute value is formatted by the generated ``__repr__()_`` method.
```
Julian: You have an extra `_` after `__repr__()`

docs/examples.rst
```diff
@@ -665,4 +665,15 @@ Finally, you can exclude single attributes from certain methods:
    ...     user = attr.ib()
    ...     password = attr.ib(repr=False)
    >>> C("me", "s3kr3t")
-   C(user='me')
+   C(user='me', password=***)
```
Julian: (This presumably is @hynek's comment too, this one is not showing exclusion anymore now.)

docs/examples.rst
```diff
@@ -665,4 +665,15 @@ Finally, you can exclude single attributes from certain methods:
...
+
+Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify custom callable to be used instead of the `repr()` built-in function:
```
Julian: specify "a"

src/attr/_make.py
```diff
@@ -128,8 +128,12 @@ def attrib(
 
     :type validator: ``callable`` or a ``list`` of ``callable``\\ s.
 
-    :param bool repr: Include this attribute in the generated ``__repr__``
-        method.
+    :param repr: Include this attribute in the generated ``__repr__``
```
Julian: "is formatted in the" ... ? "resulting repr" presumably?

src/attr/_make.py
```diff
@@ -128,8 +128,12 @@ def attrib(
...
+        method. If ``True``, include the attribute; if ``False``, omit it. By
+        default, the built-in ``repr()`` function is used. To override how the
+        attribute value is formatted in the, pass a custom ``callable``, which
+        should take a single value, and return a string.
```
Julian: Not sure it's worth explicitly calling out (would possibly personally lean towards "yes" slightly), and mention that the resulting string is `str`ed not `repr`ed into the final repr? I.e. if you return a string, you get no quotes, which will possibly trip some new user up.

src/attr/_make.py
```diff
@@ -1244,12 +1248,18 @@ def __repr__(self):
         try:
             result = [class_name, "("]
             first = True
-            for name in attr_names:
+            for attr in attrs:
+                if not attr.repr:
```
Julian: You are going to filter out falsy-callables here "accidentally".

src/attr/_make.py
```diff
@@ -1210,9 +1214,9 @@ def _add_cmp(cls, attrs=None):
 
 def _make_repr(attrs, ns):
     """
-    Make a repr method for *attr_names* adding *ns* to the full name.
+    Make a repr method that includes relevant *attrs*, adding *ns* to the full
```
Julian: Not sure if y'all do this style-wise, but this breaks the docstring summary line now that it's 2 lines.

Julian:
Cool! Left random comments just to help myself understand what this was doing mostly, but yeah looks like it'd indeed hit #212.

hynek:
Additionally to what Julian wrote,  please add an example to `typing_example.py` to verify typing works. Thanks!

docs/examples.rst
wbolster: fixed in f056bce46bb7a6fcb68287b751bf15f88dd05e8d

changelog.d/568.change.rst
wbolster: fixed in b18eb9e6ef3cdc264b5461e168e3329a13dcce16

src/attr/_make.py
wbolster: fixed in 6d37c8a4eaa5ffa39077ac5a64e70687b5ccde53

src/attr/_make.py
wbolster: fixed in 6d37c8a4eaa5ffa39077ac5a64e70687b5ccde53

src/attr/_make.py
wbolster: fixed in fixed in 6d37c8a4eaa5ffa39077ac5a64e70687b5ccde53

src/attr/_make.py
wbolster: there are may more with multi-line ‘summary’ docstring lines. and this is internal api only so not really an issue at all (not used in docs or processed by external tools).

changelog.d/568.change.rst
```diff
@@ -0,0 +1 @@
+The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a custom callable, which makes it possible to influence how the attribute value is formatted by the generated ``__repr__()`` method.
```
hynek: I find this a big vague if you don't know what this is about.

How about this:

```rst
The value passed to ``@attr.ib(repr=…)`` can now be either a boolean (as before) or a callable.
That callable must return a string and is then used for formatting the attribute by the generated ``__repr__()`` method.
```

docs/examples.rst
```diff
@@ -666,3 +666,14 @@ Finally, you can exclude single attributes from certain methods:
    ...     password = attr.ib(repr=False)
    >>> C("me", "s3kr3t")
    C(user='me')
+
+Alternatively, to influence how the generated ``__repr__()`` method formats a specific attribute, specify a custom callable to be used instead of the `repr()` built-in function:
```
hynek: `repr()` is using markdown

src/attr/_make.py
hynek: Yes, please be explicit here.

wbolster: refactored it a bit in 23515814a717ad83869dba252b401582a5f7274b

strict `bool` checking, and only do the filtering and ‘which repr to use?’ detection once, instead of per `repr()` call.

docs/examples.rst
wbolster: fixed in f6b766f1ada5f8791738b51d851076695d64645c

docs/examples.rst
wbolster: fixed in e6c98fc52102989a84006ea41f6d87181cfd920d

Julian: (FWIW this syntax *does* work in rst if you set `default_role`, which is totally great).

wbolster:
> Almost there, but please also add an `versionchanged` tag to `attrib`, thanks!

sure, but what's the right value?

changelog.d/568.change.rst
wbolster: no strong opinion; applied your changes as-is: 72046cb9c79c0bdbad9ed29a8f5a9ec80b1d277b

hynek:
> > Almost there, but please also add an `versionchanged` tag to `attrib`, thanks!
> 
> sure, but what's the right value?

Next release is 19.2.0.

src/attr/_make.py
```diff
@@ -128,8 +128,14 @@ def attrib(
...
-    :param bool repr: Include this attribute in the generated ``__repr__``
-        method.
+    :param repr: Include this attribute in the generated ``__repr__``
+        method. If ``True``, include the attribute; if ``False``, omit it. By
+        default, the built-in ``repr()`` function is used. To override how the
+        attribute value is formatted, pass a ``callable`` that takes a single
+        value and returns a string. Note that the resulting string is used
+        as-is, i.e. it will be used directly *instead* of calling ``repr()``
+        (the default).
+    :type repr: a ``bool`` or a ``callable`` to use a custom function.
```
wbolster: added

```diff
+    .. versionchanged:: 19.2.0 *repr* also accepts a custom callable.
```

in 0c7ae9fb33660da5b75cd36de49f3eb488858ba1

wbolster:
> Additionally to what Julian wrote, please add an example to `typing_example.py` to verify typing works. Thanks!

added in b1bd940d9c2dbc395b5640df92cda179eda74bc4

wbolster:
ok, please review again. i think i fully addressed all comments so far.

wbolster:
pr description updated as well

wbolster:
:rocket:  thanks all the quick turn-around

PR #570, opened 2019-09-07
Stop comparing/ordering subclasses

hynek:
This has been deprecated for a year and was raising a warning.

Done in preparation for splitting cmp in eq/order.


Julian:
More hooray!

PR #571, opened 2019-09-08
Add default_role for Sphinx.

Julian:
(Mentioned in https://github.com/python-attrs/attrs/pull/568#discussion_r321687907)

Fix a bunch of broken refs along the way, which become
errors now via -W if you have default_role set.

In theory you can catch those via sphinx-build -n (i.e.
nitpick mode), which IMHO is a decent idea anyhow, but it's
a longer diff to enable that because it'd involve fixing a
bunch of the places that try to reference types that don't
exist (e.g. :type foo: Any value). But obviously can be done.

Also didn't actually use this anywhere yet (the any role),
but will do so in a follow-up if this is acceptable.

But throwing the PR up as-is just to show what I meant / see if it's something we want.

# Pull Request Check List


Julian:
OK and now I just used this for a bunch of builtin refs just to show you how it looks. I can keep going, but actually there are some more broken refs hiding that don't become errors without `nitpick`. E.g.: https://github.com/python-attrs/attrs/blob/f84f451557bd816b1ba525d98a431709449fdf1e/src/attr/validators.py#L207 this is actually a broken ref (to `NotCallableError`), which doesn't seem like it's included in any document, but that doesn't raise an error without having nitpick mode turned on unless you try to `any`-role-ref it.

hynek:
That looks cool overall. If we go that way it should definitely pass in nitpick mode if it's possible at all.

Julian:
OK done. And looks like there weren't very many more broken references after the first few, so that's good, was fairly easy.

So nitpick mode is enabled -- I use the most "conservative" way to handle non-types being mentioned as types myself -- which is basically from now on you literally need to whitelist those in Sphinx's config. IMO this is the right idea considering just how easy it is to get Sphinx to silently not link a reference (as you saw without -n :/) but there are some other ways to do this if you don't like this one.

(Sphinx *will* though error if the default_role finds multiple possible targets, so if you accidentally have a doc and a type named the same thing or whatever, you're still fine in that you'll get a non-silent failure.)

hynek:
Cool, TIL, thanks!

PR #574, opened 2019-09-16
Split cmp into eq and order

### LINKED ISSUE #170: Suggestion: differentiate in cmp between (== , !=) and (>,>=, <, <=).
Thanks for a fantastic library!

One thing that surprised me at first was that I could not write an effective `__gt__()`-function for my attr-class. I found the `cmp`-init-parameter later (in the code), but I still wonder if it wouldn't be a good idea if I could have the option of keeping the tuple-based `__eq__` and `__ne__`-implementations, but override size-comparisons (maybe based on a single `__lt__`-function?).

hynek:
Fixes #170

Whoa, this is a doozy. People have been asking for this for a long time and I really wanted it too. So here it is, with a lot more changes that you'd expect.

Since this one is rather radical, I think we'll give it at least 18 months of deprecation period. I also think that I'll add a `sys.version_info`-style version number so people can easily write warnings-free code that won't break.

This is a very important stepping stone towards `detect_own` (which implies `@attr.auto`/`import attrs`) and pluggable method factories. In auto/import attrs, `order` will become False by default.


tests/test_dark_magic.py
```diff
@@ -581,3 +590,48 @@ class FooError(Exception):
...
+            )
+
+            with pytest.raises(TypeError) as ei:
+                C(5) < C(6)
+
+            assert ei.value.args[0] in possible_errors
+        else:
+            i = C(42)
+            for m in ("lt", "le", "gt", "ge"):
+                assert None is getattr(i, m, None)
```
Julian: This looks broken (passing always accidentally).

Probably should use the same kind of test as above where you actually run comparison instead of testing negatives.

changelog.d/574.deprecation.rst
```diff
@@ -0,0 +1,11 @@
+The ``cmp`` argument to ``attr.s()`` and ``attr.ib()`` is now deprecated.
+
+Please use ``eq`` to add equality methods (``__eq__`` and ``__ne__``) and ``order`` to add ordering methods (``__lt__``, ``__le__``, ``__gt__``, and ``__ge__``) instead – just like with `dataclasses <https://docs.python.org/3/library/dataclasses.html>`_.
+
+Both are effectively ``True`` by default but it's enough to set ``eq=False`` to disable both at once.
+The passing explicitly ``eq=False, order=True`` will raise a ``ValueError`` though.
```
Julian: s/The passing explicitly .../Passing \1 explicitly/

changelog.d/574.deprecation.rst
```diff
@@ -0,0 +1,11 @@
...
+
+Since this is arguably a deeper backward-compatibilty break, it will have an extended deprecation period until 2021-06-01.
```
Julian: compatibility

changelog.d/574.deprecation.rst
```diff
@@ -0,0 +1,11 @@
+The ``cmp`` argument to ``attr.s()`` and ``attr.ib()`` is now deprecated.
+
+Please use ``eq`` to add equality methods (``__eq__`` and ``__ne__``) and ``order`` to add ordering methods (``__lt__``, ``__le__``, ``__gt__``, and ``__ge__``) instead – just like with `dataclasses <https://docs.python.org/3/library/dataclasses.html>`_.
+
+Both are effectively ``True`` by default but it's enough to set ``eq=False`` to disable both at once.
```
Julian: It seems to not just be "effectively" True, from the below it looks literally True, no?

docs/hashing.rst
```diff
@@ -10,7 +10,7 @@ Hash Method Generation
    Leave it at ``None`` which means that ``attrs`` will do the right thing for you, depending on the other parameters:
 
    - If you want to make objects hashable by value: use ``@attr.s(frozen=True)``.
-   - If you want hashing and comparison by object identity: use ``@attr.s(cmp=False)``
+   - If you want hashing and comparison by object identity: use ``@attr.s(eq=False)``
```
Julian: I didn't execute this to check but this doesn't give comparison anymore like cmp did, right? I.e. s/comparison/equality?

hynek: Ah yeah totally.

changelog.d/574.deprecation.rst
hynek: Literally it's `None` and the `True` is derived if `cmp` is kept `None` too. Or am I missing something?

changelog.d/574.deprecation.rst
hynek: I wonder if I'll _ever_ manage to type this word correctly.

changelog.d/574.deprecation.rst
hynek: You didn't catch a group but I think I know what you mean.  🥴

tests/test_dark_magic.py
hynek: How? Python 2 happily compares incomparable types.

changelog.d/574.deprecation.rst
Julian: Ahhh ok, got it (the "below" I was seeing was the repr output, but yeah I see the actual argument default is as you say).

changelog.d/574.deprecation.rst
Julian: :D

tests/test_dark_magic.py
Julian: Possibly by just checking you get the default behavior. Either by creating two classes and comparing them and seeing it doesn't blow up even if you compare them to each other, or by creating two instances and asserting the one with higher ID won.

But if all that sounds like too much work, glad at least that it looks like you added the double underscores so that this does something now :)

src/attr/_make.py
```diff
@@ -754,17 +811,28 @@ def attrs(
...
+        and ``__ge__`` methods that behave like *eq* above and allow instances
+        to be ordered. If ``None`` (default) mirror value of *eq*.
+    :type order: `bool` or `None`
+    :param cmp: Setting to ``True`` is equivalent to setting ``eq=True,
+        order=True``. Deprecated in favor of *eq* and *order*, has precedence
+        over them for backward-compatibility though. Must not be mixed with
+        *eq* or *order*.
+    :type cmp: `bool` or `None`
     :param hash: If ``None`` (default), the ``__hash__`` method is generated
         according how *cmp* and *frozen* are set.
```
Julian: Missed a cmp

hynek: Goddamit I keep grepping for it and keep missing it.

src/attr/_make.py
```diff
@@ -1757,9 +1835,19 @@ def from_counting_attr(cls, name, ca, type=None):
...
+            cmp=None,
             **inst_dict
         )
 
+    @property
+    def cmp(self):
+        """
+        Simulate the presence of a cmp attribute and warn.
+        """
+        warnings.warn(_CMP_DEPRECATION, DeprecationWarning, stacklevel=2)
```
Julian: Is this (accessing Attribute.cmp directly) tested somewhere?

(The deprecation warning, and whether it gives the right answer both if you used eq and / or cmp?)

hynek: The deprecation is tested in https://github.com/python-attrs/attrs/pull/574/files#diff-11d337c4d38019bd30952e7e1ef21ec0R624-R637

I guess I [cs]ould add a test for the values too.

tests/test_dark_magic.py
hynek: It does, let's keep it that way. :)

hynek:
OK, I think I've got everything?

src/attr/_make.py
```diff
@@ -135,10 +138,15 @@ def attrib(
...
+    :param bool eq: If ``True`` (default), include this attribute in the
+        generated ``__eq__`` and ``__ne__`` methods that check two instances
+        for equality.
+    :param bool order: If ``True`` (default), include this attributes in the
+        generated ``__lt__``, ``__le__``, ``__gt__`` and ``__ge__`` methods.
+    :param bool cmp: Setting to ``True`` is equivalent to setting ``eq=True,
+        order=True``. Deprecated in favor of *eq* and *order*.
     :param hash: Include this attribute in the generated ``__hash__``
-        method.  If ``None`` (default), mirror *cmp*'s value.  This is the
+        method.  If ``None`` (default), mirror *eq*'s value.  This is the
```
Julian: This seems especially "suspicious" that eq is special (rather than order) if they're settable independently -- i.e., why is this any more related to eq than order from a default perspective (so just a continuation of the skepticism above)?

src/attr/_make.py
```diff
@@ -709,12 +731,45 @@ def _add_method_dunders(self, method):
...
+    if cmp is not None and any((eq is not None, order is not None)):
+        raise ValueError("Don't mix `cmp` with `eq' and `order`.")
+
+    # cmp takes precedence due to bw-compatibility.
+    if cmp is not None:
+        warnings.warn(_CMP_DEPRECATION, DeprecationWarning, stacklevel=3)
+
+        return cmp, cmp
+
+    # Both on by default.
```
Julian: Is this comment right? It seems like if eq is False and order is None you still get order = False?

I guess "both on by default" means "both on unless you touch either one", which yeah ok nevermind.

tests/test_dark_magic.py
```diff
@@ -581,3 +592,76 @@ class FooError(Exception):
...
+    def test_cmp_deprecated_attribute(self, cmp, eq, order):
+        """
+        Accessing Attribute.cmp raises a deprecation warning but returns True
+        if cmp is True, or eq and order are *both* effectively True.
+        """
+        # These cases are invalid and raise a ValueError.
+        assume(cmp is None or (eq is None and order is None))
+        assume(not (eq is False and order is True))
+
+        if cmp is not None:
```
Julian: This is super twisty for a test case, but I don't have a concrete recommendation personally (other than "maybe ideally this should be multiple different test cases, and where you explicitly set the right values in the hypothesis strategies that each one should set").

hynek: I did have a version that was still one test but ended up having a whole table of values which was big and even harder to understand. Hence I decided to switch to property testing…both isn't ideal. :|

src/attr/_make.py
hynek: I have clarified the comment.

src/attr/_make.py
```diff
@@ -135,10 +138,15 @@ def attrib(
         as-is, i.e. it will be used directly *instead* of calling ``repr()``
         (the default).
     :type repr: a ``bool`` or a ``callable`` to use a custom function.
-    :param bool cmp: Include this attribute in the generated comparison methods
-        (``__eq__`` et al).
+    :param bool eq: If ``True`` (default), include this attribute in the
```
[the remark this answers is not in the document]
hynek: My question would be: is the fact that we can't come up with a reason right now worth it to:

1. break the symmetry of `@attr.s()` and `attr.ib()`
2. confusingly  keep an argument around, that doesn't exist in `@attr.s` anymore?

Since `order` mirrors `eq` by default, it's not more arguments to pass and maybe there is at some point _a_ reason to order over a subset but compare over everything? `eq=False, order=True` is invalid ATM and I'd keep that invariant around until someone give me a good use case.

src/attr/_make.py
hynek: Hashing [relies on equality](https://hynek.me/articles/hashes-and-equality/) and doesn't care about ordering at all. Or am I misunderstanding?

src/attr/_make.py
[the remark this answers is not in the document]
Julian: I have to think about this, like I said my initial intuition is the other way, but I'll try to convince myself otherwise.

But, for clarity, the other side of the argument you mentioned (and in fact the side mine above is entirely rooted in) is "the symmetry here is false -- these arguments have the same name, but they affect a very different part of the equation" and more broadly, that you can't put cats back in bags.

We can always add the capability to change these individually later if someone compels with a use case, and to your point, the cmp argument is staying no matter what whilst it's supported for backwards compat, so it's a net *non* addition to not add this functionality.

Whereas adding it potentially, again at first glance, adds possibility of mistake for end users to create "non-sensical objects".

And yeah I don't think this has to do with that eq=False order=True combo, the nonsensical objects come about when both *are* enabled.

Just my 0.02 before having morning coffee though...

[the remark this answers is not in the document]
hynek: `cmp`is supposed to go ion 2021 though. And I'd be rather unhappy to keep around just for this one case.

I really don't think those objects are really non-sensical?

I mean `eq=False,order=True` Can't happen. Both True and both False make sense. As for `True,False` how about this:

```
@attr.s
class Person:
    id = attr.ib(order=False)
    name = attr.ib()
```

Since we're both not entirely sure, let's try to ask someone else: @glyph?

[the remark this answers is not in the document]
Julian: OK *so* -- now that I'm home and actually have time to open a REPL and try using this :) -- I totally misunderstood that the `eq=False, order=True` error applied here too, I thought you were telling me about `attr.s(eq=False, order=True)`, not `attr.ib(eq=False, order=True)` -- so it was definitely the latter that I think is non-sensical.

The person example I think is... "ugly", but someone is going to want to do it anyhow, so yeah not bad enough to me that I'd think you'd not allow it.

So anyways the way this seems to work I *think* makes sense to me now :) so apologies for the slow understanding.

Probably still makes sense to get a third opinion on this point but yeah I think this is probably fine to me too!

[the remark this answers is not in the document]
hynek: Cool, thanks for taking the time on your trip to talk it through with me – your instagram had nicer vistas than my crummy code. :)

hynek:
Thank you _so_ much Julian, I realize this PR looks daunting and I appreciate your trooped through!

src/attr/_make.py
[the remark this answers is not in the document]
Julian: HAH

glyph:
Glad to see my opinion was not in fact required 😄 .

Also, nice to see progress on this!

PR #580, opened 2019-09-25
Add attr.__version_info__

hynek:
This allows users to check for features and avoid deprecation warnings without
breaking backward compatibility.

This is especially important with the cmp → order/eq split.

---

@euresti do I need to add it to `__init__.pyi` if it's an attrs class? I tried to just import it into it, but that started throwing errors about stuff in `_make.py`…

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -508,7 +508,7 @@ Deprecated APIs
 
 .. _version-info:
 
-To help you writing backward compatible code that doesn't throw warnings on modern releases, the ``attr`` module has an ``__version_info__`` attribute as of version 19.2.0.
+To help you write backward compatible code that doesn't throw warnings on modern releases, the ``attr`` module has an ``__version_info__`` attribute as of version 19.2.0.
 It behaves similarly to `sys.version_info` and is an instance of `VersionInfo`:
 
 .. autoclass:: VersionInfo
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -518,7 +518,7 @@ It behaves similarly to `sys.version_info` and is an instance of `VersionInfo`:
    >>> if getattr(attr, "__version_info__", (0,)) >= (19, 2):
    ...     cmp_off = {"eq": False}
    ... else:
-   ...     cmp_off= {"cmp": False}
+   ...     cmp_off = {"cmp": False}
    >>> cmp_off == {"eq":  False}
    True
    >>> @attr.s(**cmp_off)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -506,7 +506,31 @@ Converters
 Deprecated APIs
 ---------------
 
+.. _version-info:
+
+To help you writing backward compatible code that doesn't throw warnings on modern releases, the ``attr`` module has an ``__version_info__`` attribute as of version 19.2.0.
+It behaves similarly to `sys.version_info` and is an instance of `VersionInfo`:
+
+.. autoclass:: VersionInfo
+
+   With its help you can write code like this:
+
+   >>> if getattr(attr, "__version_info__", (0,)) >= (19, 2):
+   ...     cmp_off = {"eq": False}
+   ... else:
+   ...     cmp_off= {"cmp": False}
+   >>> cmp_off == {"eq":  False}
+   True
+   >>> @attr.s(**cmp_off)
+   ... class C(object):
+   ...     pass
+
+
+----
+
 The serious business aliases used to be called ``attr.attributes`` and ``attr.attr``.
 There are no plans to remove them but they shouldn't be used in new code.
+
+The ``cmp`` argument to both `attr.s` and `attr.ib` has been deprecated in 19.2 and shouldn't be used.
 
 .. autofunction:: assoc
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -22,6 +22,17 @@ from . import validators as validators
 
 from ._version import VersionInfo
 
+__version__: str
+__version_info__: VersionInfo
+__title__: str
+__description__: str
+__url__: str
+__uri__: str
+__author__: str
+__email__: str
+__license__: str
+__copyright__: str
+
 _T = TypeVar("_T")
 _C = TypeVar("_C", bound=type)
 
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -20,6 +20,8 @@ from . import filters as filters
 from . import converters as converters
 from . import validators as validators
 
+from ._version import VersionInfo
+
 _T = TypeVar("_T")
 _C = TypeVar("_C", bound=type)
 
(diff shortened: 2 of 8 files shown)


docs/api.rst
```diff
@@ -506,7 +506,31 @@ Converters
...
+It behaves similarly to `sys.version_info` and is an instance of `VersionInfo`:
+
+.. autoclass:: VersionInfo
+
+   With its help you can write code like this:
+
+   >>> if getattr(attr, "__version_info__", (0,)) >= (19, 2):
+   ...     cmp_off = {"eq": False}
+   ... else:
+   ...     cmp_off= {"cmp": False}
```
wsanchez: space missing before '='

euresti:
I think you should add it to the __init__.pyi and the other missing ones like this:

```
__version__: str
__title__: str
__description__: str
__url__: str
__uri__: str
__author__: str
__email__: str
__license__: str
__copyright__: str
```

hynek:
Technically, mypy knows the types of VersionInfo…but if I import it, I get the errors…is there a way around here? Or do I have to either fix mypy for attrs or duplicate the information?

```
src/attr/_compat.py:18: error: Cannot assign multiple types to name "ordered_dict" without an explicit "Type[...]" annotation
src/attr/_make.py:1538: error: All conditional function variants must have identical signatures
src/attr/_make.py:1907: error: Cannot assign to a type
src/attr/_make.py:1909: error: "Attribute" has no attribute "hash"
src/attr/_make.py:2039: error: Cannot assign to a type
```

are the new ones.

euresti:
Oh.  You shouldn't run mypy on attrs source code.  At least we haven't made it pass type checking.  The only file that passes type checking is `typing_example.py`

euresti:
Although:
> src/attr/_make.py:1538: error: All conditional function variants must have identical signatures
is complaining that 
```
            def fmt_setter(attr_name, value_var):
            def fmt_setter(attr_name, value_var):
        # Not frozen.
        def fmt_setter(attr_name, value):
```

hynek:
> Oh. You shouldn't run mypy on attrs source code. At least we haven't made it pass type checking. The only file that passes type checking is typing_example.py

I’m not! 🙈 I’m just adding a `from ._version import VersionInfo` And it seem to follow the imports into _make?

euresti:
Oh I see.  You added `_version.py`  you'll probably have to add `_version.pyi`
But I can't repro your error messages.  What command are you running to trigger them?

hynek:
Now that I've rebased, if I add  a line `from ._version import VersionInfo
` to `__init__.pyi` _without_ adding a `_version.pyi`, I get:

```
❯ tox -e typing
typing inst-nodeps: /Users/hynek/Projects/attrs/.tox/.tmp/package/1/attrs-19.2.0.dev0.tar.gz
typing installed: atomicwrites==1.3.0,attrs==19.2.0.dev0,coverage==4.5.4,hypothesis==4.36.0,importlib-metadata==0.22,more-itertools==7.2.0,mypy==0.720,mypy-extensions==0.4.1,packaging==19.1,pluggy==0.13.0,py==1.8.0,Pympler==0.7,pyparsing==2.4.2,pytest==5.1.2,six==1.12.0,typed-ast==1.4.0,typing-extensions==3.7.4,wcwidth==0.1.7,zipp==0.6.0,zope.interface==4.6.0
typing run-test-pre: PYTHONHASHSEED='3295264721'
typing run-test: commands[0] | mypy src/attr/__init__.pyi src/attr/converters.pyi src/attr/exceptions.pyi src/attr/filters.pyi src/attr/validators.pyi
src/attr/_compat.py:18: error: Cannot assign multiple types to name "ordered_dict" without an explicit "Type[...]" annotation
src/attr/_make.py:1538: error: All conditional function variants must have identical signatures
src/attr/_make.py:1907: error: Cannot assign to a type
src/attr/_make.py:1909: error: "Attribute" has no attribute "hash"
src/attr/_make.py:2039: error: Cannot assign to a type
ERROR: InvocationError for command /Users/hynek/Projects/attrs/.tox/typing/bin/mypy src/attr/__init__.pyi src/attr/converters.pyi src/attr/exceptions.pyi src/attr/filters.pyi src/attr/validators.pyi (exited with code 1)
```

---

> You added _version.py you'll probably have to add _version.pyi

1. Let's appreciate the irony of me saying that we don't add files a lot _yesterday_.
2. So I'll have to replicate the type information I guess? :(

hynek:
OK I'm not happy but I think this it r4r now. And I thought this is gonna a bunch of lines – I'll never learn! 🤪

src/attr/__init__.pyi
```diff
@@ -20,6 +20,8 @@ from . import filters as filters
 from . import converters as converters
 from . import validators as validators
 
+from ._version import VersionInfo
+
 _T = TypeVar("_T")
 _C = TypeVar("_C", bound=type)
```
euresti: You should also add the constants here.

src/attr/_version.pyi
```diff
@@ -0,0 +1,5 @@
+class VersionInfo:
+    year: int
```
euresti: I I was being pedantic I would mention that since the class is frozen=True this should be:
```
class VersionInfo:
    @property
    def year(self) -> int: ...
    @property
    def minor(self) -> int: ...
    @property
    def micro(self) -> int: ...
    @property
    def releaselevel(self) -> str: ...
```

euresti:
All the typing stuff looks good to me!

hynek:
Thanks David!

Anyone else for the rest? :)

docs/api.rst
```diff
@@ -506,7 +506,31 @@ Converters
 Deprecated APIs
 ---------------
 
+.. _version-info:
+
+To help you writing backward compatible code that doesn't throw warnings on modern releases, the ``attr`` module has an ``__version_info__`` attribute as of version 19.2.0.
```
euresti: "To help you write backward compatible code" I think sounds better.

tests/test_version.py
```diff
@@ -0,0 +1,49 @@
...
+        assert vi == VersionInfo._from_version_string("19.2.0")
+
+    @pytest.mark.parametrize("other", [(), (19, 2, 0, "final", "garbage")])
+    def test_wrong_len(self, vi, other):
+        """
+        Comparing with a tuple that has the wrong length raises an error.
+        """
+        assert vi != other
+
+        if not PY2:
```
euresti: Why is this only py3?

docs/api.rst
hynek: Ugh you're right. Sometimes I wonder what I was thinking.

tests/test_version.py
hynek: Because Python 2 is too YOLO to raise an error if comparison returns `NotImplemented`. :|

However this should be made using a decorator that explains your question.

hynek:
Thanks, you followed the protocol perfectly! 💛

PR #605, opened 2020-01-03
Mention attr.fields() after __attrs_attrs__ in 'How does it work'

altendky:
The mentioning of __attrs_attrs__ in a 'how does it work' document makes sense but it seems important to mention that there are good non-dunder ways to access the data.

WIP for:
- [x] Checking doc rendering for the attrs.fields() link

### DIFF
--- a/docs/how-does-it-work.rst
+++ b/docs/how-does-it-work.rst
@@ -40,7 +40,7 @@ No magic, no meta programming, no expensive introspection at runtime.
 Everything until this point happens exactly *once* when the class is defined.
 As soon as a class is done, it's done.
 And it's just a regular Python class like any other, except for a single ``__attrs_attrs__`` attribute that ``attrs`` uses internally.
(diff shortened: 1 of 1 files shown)


docs/how-does-it-work.rst
```diff
@@ -39,7 +39,8 @@ No magic, no meta programming, no expensive introspection at runtime.
 
 Everything until this point happens exactly *once* when the class is defined.
 As soon as a class is done, it's done.
-And it's just a regular Python class like any other, except for a single ``__attrs_attrs__`` attribute that can be used for introspection or for writing your own tools and decorators on top of ``attrs`` (like `attr.asdict`).
+And it's just a regular Python class like any other, except for a single ``__attrs_attrs__`` attribute that ``attrs`` uses internally.
+Much of the information is accessible via :meth:`attr.fields()` and other functions which can be used for introspection or for writing your own tools and decorators on top of ``attrs`` (like `attr.asdict`).
```
hynek: I think it should be just
```
`attr.fields`
```

now that we set a default domain (see also attr.asdict further down).

hynek:
Thanks! I'll ignore CI here. *shakes fist at AP*

PR #608, opened 2020-01-06
Officially drop Python 3.4

hynek:
AP dropped it and there's really no reason to support it anymore. Xenial is on 3.5.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -130,7 +130,7 @@ Project Information
 its documentation lives at `Read the Docs <https://www.attrs.org/>`_,
 the code on `GitHub <https://github.com/python-attrs/attrs>`_,
 and the latest release on `PyPI <https://pypi.org/project/attrs/>`_.
-It’s rigorously tested on Python 2.7, 3.4+, and PyPy.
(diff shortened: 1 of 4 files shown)


henryiii:
> But we don’t test it anymore and will block it once someone reports breakage.

If you are using `python_requires`, this is the wrong way to use it. If you wait till it's broken _then_ update `python_requires`, `pip install attrs` will remain broken forever after in that version of Python. It's better to do what NumPy, etc. do, and make a clean break while it is still working, so that `pip install` will keep working, just installing a known working version of Python. (Now, some of the old versions of Python don't have a new enough pip to do this, and if you are referring to setup.py triggering a block, that's fine).

hynek:
Yes, that's true.

I'll try to find a way to run the test suite against 3.4 before releasing and the do or don't add the metadat. I guess I'll be able to concoct something with Docker (3.4 doesn't work on macOS anymore).

henryiii:
GHA recently added 3.0-3.4 for setup-python@v2, and of course there’s docker. I’m on Big Sur now so need Docker now for anything before about 3.8 anyway. :/

PR #612, opened 2020-01-07
Fixed frozen and cache_hash incompatibility

### LINKED ISSUE #611: frozen=True incompatible with cache_hash=True as of 19.1.0
Prior to #489, this used to work:

```python
import attr
import copy

class FrozenWithCache:
    pass


if __name__ == "__main__":
    copy.deepcopy(FrozenWithCache())
```

The issue is that `frozen` is implemented by throwing an error in `__setattr__`, and clearing the hash cache calls `setattr`. I think this can be solved by using [the same mechanism that `_make_hash` uses for frozen classes](https://github.com/python-attrs/attrs/blob/b6bd8c8d2bb0c53bbad0f11377df6083313d26c2/src/attr/_make.py#L1119), but I have not tried implementing this before.

Another option is to not clear the hash cache on serialization for frozen classes.

pganssle:
This uses the same mechanism to clear the cache hash as `__hash__` uses to set the cache hash in frozen classes - bypassing the object's `__setattr__` in favor of `object.__setattr__`. Neat trick.

Fixes #611.

# Pull Request Check List

- (N/A) New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/strategies.py).
- (N/A) Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``).
- (N/A) Updated **documentation** for changed code.

### DIFF
--- a/changelog.d/611.change.rst
+++ b/changelog.d/611.change.rst
@@ -1,0 +1,1 @@
+Fixed serialization of classes with ``frozen=True`` and ``cache_hash=True``, which has been broken since 19.1.0.
--- a/tests/test_make.py
+++ b/tests/test_make.py
@@ -1466,6 +1466,18 @@ class TestClassBuilder(object):
 
         assert [C2] == C.__subclasses__()
 
+    @pytest.mark.xfail(raises=attr.exceptions.FrozenInstanceError)
+    def test_cache_hash_with_frozen_serializes(self):
+        """
+        Frozen classes with cache_hash should be serializable.
+        """
+
+        @attr.s(cache_hash=True, frozen=True)
+        class C(object):
+            pass
+
+        copy.deepcopy(C())
+
 
 class TestMakeOrder:
     """
--- a/tests/test_make.py
+++ b/tests/test_make.py
@@ -1466,7 +1466,6 @@ class TestClassBuilder(object):
 
         assert [C2] == C.__subclasses__()
 
-    @pytest.mark.xfail(raises=attr.exceptions.FrozenInstanceError)
     def test_cache_hash_with_frozen_serializes(self):
         """
         Frozen classes with cache_hash should be serializable.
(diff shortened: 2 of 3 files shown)


tests/test_make.py
```diff
@@ -1467,6 +1467,15 @@ class C2(C):
         assert [C2] == C.__subclasses__()
 
 
+    def test_cache_hash_with_frozen_serializes(self):
+        """Frozen classes with cache_hash should be serializable."""
+        @attr.s(cache_hash=True, frozen=True)
+        class C:
```
hynek: Unfortunately, we still support Python 2.7. :( This is the reason the 2.7 builders broke.

changelog.d/611.change.rst
```diff
@@ -0,0 +1 @@
+Fixed serialization of classes with `frozen=True` and `cache_hash=True`, which has been broken since 19.1.0.
```
hynek: This ain't Markdown, Paul. 😇

tests/test_make.py
```diff
@@ -1467,6 +1467,15 @@ class C2(C):
         assert [C2] == C.__subclasses__()
 
 
+    def test_cache_hash_with_frozen_serializes(self):
+        """Frozen classes with cache_hash should be serializable."""
```
hynek: Nit: please give the `"""` lines their own lines, thanks.

changelog.d/611.change.rst
pganssle: This is what I get for making a PR when I'm in a rush to get out the door. Good catch, thanks!

pganssle:
Oops, thanks to the coverage result I realized that the test was copying the *class*, not an *instance* of the class. This is what I get for skipping my usual semi-TDD style of putting the test under an `xfail` in a separate commit, and removing the `xfail` in the commit that fixes it, so I've refactored the commit history a bit.

Feel free to squash the commits if you don't like that style - I mainly do it to give an easy(-enough)-to-verify way to show that the test is probably testing what we expect.

hynek:
Thanks, Paul!

PR #615, opened 2020-01-13
WIP: Implement __reduce__ with cache_hash

### LINKED ISSUE #494: Allow cache_hash=True to be used when a custom __setstate__ is present

### LINKED ISSUE #613: cache_hash breaks copy and pickle for non-slots classes

```python
        class C(object):
            x = attr.ib()

        a = C(1)
        b = copy.deepcopy(C(1))

        assert a == b
```

```python
>>> import attr
>>> import copy
... class SomeClass:
...     x = attr.ib()
...
>>> SomeClass(1)
SomeClass(x=1)
SomeClass(x=NOTHING)
```

pganssle:
It turns out that the hash cache-clearing implementation for non-slots classes was flawed and never quite worked properly. This switches away from using `__setstate__` and instead adds a custom `__reduce__` that removes the cached hash value from the default serialized output.

This commit also refactors some of the tests a bit, to try and more cleanly organize the tests related to this issue.

Fixes #613.
Fixes #494.

# Pull Request Check List

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -449,7 +449,9 @@ def _cache_hash_reduce(self):
         state = obj_reduce[2]
 
         if isinstance(state, dict) and _hash_cache_field in state:
+            state = copy.copy(state)
             state[_hash_cache_field] = None
+            obj_reduce = obj_reduce[0:2] + (state,) + obj_reduce[3:]
 
     return obj_reduce
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -426,6 +426,17 @@ def _frozen_delattrs(self, name):
     raise FrozenInstanceError()
 
 
+def _cache_hash_reduce(self):
+    obj_reduce = object.__reduce__(self)
+    if len(obj_reduce) > 2:
+        state = obj_reduce[2]
+
+        if isinstance(state, dict) and _hash_cache_field in state:
+            state[_hash_cache_field] = None
+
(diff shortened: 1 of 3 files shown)


pganssle:
I am not sure I understand what's going on with the pypy target - I cannot reproduce the failure locally.

src/attr/_make.py
```diff
@@ -46,6 +46,19 @@
 _sentinel = object()
 
 
+if PY2 and object.__reduce__ is not int.__reduce__:
```
hynek: pypy doesn't run under coverage because it's dead slow. You'll have to mark this for no coverage.

src/attr/_make.py
```diff
@@ -483,6 +513,13 @@ def __init__(
             self._cls_dict["__setattr__"] = _frozen_setattrs
             self._cls_dict["__delattr__"] = _frozen_delattrs
 
+        if (
+            cache_hash
+            and _method_eq(cls.__reduce__, object.__reduce__)
```
pganssle: Note to reviewers: This is probably the most controversial part - the `__reduce__` is generated only if you haven't defined your own `__reduce__` or `__reduce_ex__`.

I have done this because it seems mildly inappropriate to squash someone's existing `__reduce__` and I am not confident that we can count on the third return value from a *custom* `__reduce__` being the object's `__dict__`.

hynek:
Closing in favor of #620.

PR #616, opened 2020-01-14
WIP: Switch over to a global hash cache

### LINKED ISSUE #494: Allow cache_hash=True to be used when a custom __setstate__ is present

### LINKED ISSUE #613: cache_hash breaks copy and pickle for non-slots classes

pganssle:
This uses an alternate approach to hash caching than the current master or #615. Rather than trying to store the hash cache on the object itself, we create a global hash cache object mapping the id of the object to its hash.

This pretty dramatically simplifies the code. The two biggest downsides:

1. This introduces weak references and mutable global state into the code. I think I've got a thread-safe implementation, but it's very hard to test for that. Even if we weren't worried about memory leaks, we definitely need the cache lifecycle tied to the object's lifecycle, because otherwise the interpreter might assign another `cache_hash=True` object the same `id()` value.

2. At least the way I've implemented it (and I don't know if there's a much faster way to do it), this method adds what I would consider a lot of overhead, compare the speed for this version of the code (both using Python 3.7.4):

```python
>>> import attr
>>> @attr.s(cache_hash=True, hash=True)
... class C:
...     x = attr.ib()
...
>>> %timeit hash(C(1))
6.2 µs ± 176 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>> x = C(1)
>>> %timeit hash(x)
1.3 µs ± 7.58 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

To the speed of the equivalent code using #615 

```python
>>> import attr
>>> @attr.s(cache_hash=True, hash=True)
... class C:
...     x = attr.ib()
...
>>> %timeit hash(C(1))
684 ns ± 9.06 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
>>> x = C(1)
>>> %timeit hash(x)
207 ns ± 1.42 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

For a more complicated class (where `cache_hash=True` might be more valuable), the "create-and-hash" numbers narrow *a bit*, but there's still a pretty significant difference for the "lookup in hash" case:

```python
>>> import attr
>>> @attr.s(cache_hash=True, hash=True)
... class C:
...     x = attr.ib(default="qbert")
...     y = attr.ib(default="hi")
...     z = attr.ib(default=2**120)
...     a = attr.ib(default=tuple(["a"] * 500))
...     b = attr.ib(default=b"A" * 100000)
...     c = attr.ib(default=200)
...
>>> %timeit hash(C())
8.85 µs ± 72.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>> x = C()
>>> %timeit hash(x)
1.24 µs ± 19.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

#615 version:

```python
>>> @attr.s(cache_hash=True, hash=True)
... class C:
...     x = attr.ib(default="qbert")
...     y = attr.ib(default="hi")
...     z = attr.ib(default=2**120)
...     a = attr.ib(default=tuple(["a"] * 500))
...     b = attr.ib(default=b"A" * 100000)
...     c = attr.ib(default=200)
...
>>> %timeit hash(C())
3.13 µs ± 156 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>> x = C()
>>> %timeit hash(x)
212 ns ± 3.54 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I have spent less time working on this version, so it could probably be optimized somewhat, but I doubt we'll get the first-hash time down much, and I think this way will always be slower.

Fixes #613.
Fixes #494.

# Pull Request Check List

### DIFF
--- a/src/attr/_hash_cache.py
+++ b/src/attr/_hash_cache.py
@@ -1,0 +1,39 @@
+import functools
+import weakref
+
+
+# Cache mapping the results of id() to its hash
+HASH_CACHE = {}
+
+if getattr(weakref, "finalize", None) is not None:
+
+    def cache_hash(obj, obj_hash):
+        global HASH_CACHE
+        id_ = id(obj)
+        cleanup_cache = functools.partial(HASH_CACHE.pop, id_, None)
+        finalizer = weakref.finalize(obj, cleanup_cache)
+
+        # At exit this will clean itself up - these finalizers are just to
+        # avoid memory leaks
+        finalizer.atexit = False
+
+        # Use `setdefault` for thread-safety
+        HASH_CACHE.setdefault(id_, obj_hash)
+
+
+else:
+    _WEAK_REFS = {}
+
+    def _cleanup_weakref(obj_id):
+        global HASH_CACHE
+        global _WEAK_REFS
+        HASH_CACHE.pop(obj_id, None)
+        _WEAK_REFS.pop(obj_id, None)
+
+    def cache_hash(obj, obj_hash):
+        id_ = id(obj)
+        cleanup_cache = functools.partial(_cleanup_weakref, id_)
+
+        r = weakref.ref(obj, cleanup_cache)
+        _WEAK_REFS.setdefault(id_, r)
+        HASH_CACHE.setdefault(id_, obj_hash)
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -38,7 +38,6 @@ _classvar_prefixes = ("typing.ClassVar", "t.ClassVar", "ClassVar")
 # we don't use a double-underscore prefix because that triggers
 # name mangling when trying to create a slot for the field
 # (when slots=True)
-_hash_cache_field = "_attrs_cached_hash"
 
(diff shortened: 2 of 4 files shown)


gabbard:
I would be inclined to say the performance difference would out-weigh the simplification to the attrs code, especially since if someone is bothering to turn on `cache_hash=True`, it's probably because they know or suspect the object may be used in a context where performance matters (this is certainly true for many of our uses of it at my workplace).  But that's not my call. :-)

pganssle:
This seems like the worst of the approaches, so I'm going to go ahead and close it.

PR #618, opened 2020-01-23
Add and_ converter

NI1993:
# Pull Request Check List

- [V] Added **tests** for changed code.
- [V] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/strategies.py).
- [V] Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``).
    - [V] ...and used in the stub test file `tests/typing_example.py`.
- [V] Updated **documentation** for changed code.
    - [V] New functions/classes have to be added to `docs/api.rst` by hand.
    - [V] Changes to the signature of `@attr.s()` have to be added by hand too.
    - [V] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).
- [V] Documentation in `.rst` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).
- [V] Changes (and possible deprecations) have news fragments in [`changelog.d`](https://github.com/python-attrs/attrs/blob/master/changelog.d).


src/attr/converters.py
```diff
@@ -7,6 +7,26 @@
...
+def and_(*converters):
+    """
+    A converter that composes multiple converters into one.
+
+    When called on a value, it runs all wrapped converters.
+
+    :param converters: Arbitrary number of converters.
+    :type converters: callables
+
+    .. versionadded:: ???
```
NI1993: How can one tell the versionadded?

wsanchez:
Lint error:
```console
tests/test_converters.py:13:1: F401 'attr.s' imported but unused
```

wsanchez:
No tests for the `and_converter` nested function.

hynek:
Looking at the code, I find the name `and` unfortunate. It's confusing by being the same as for validators but doing something different. I guess `chain` would work?

NI1993:
> Lint error:
> 
> ```
> tests/test_converters.py:13:1: F401 'attr.s' imported but unused
> ```

Thanks, removed it

NI1993:
> Looking at the code, I find the name `and` unfortunate. It's confusing by being the same as for validators but doing something different. I guess `chain` would work?

Certainly!
Thanks, renamed it

NI1993:
> No tests for the `and_converter` nested function.

I'm not sure [how](https://stackoverflow.com/questions/13606178/python-how-do-you-unit-test-a-nested-function) to test it :\
Can u please explain or refer me so I can learn how?

src/attr/converters.py
hynek: It's in `attr.__init__` without the `.dev0` suffix.

tests/test_converters.py
```diff
@@ -95,3 +97,39 @@ def test_none_factory(self):
...
+        # Last wrapped converter fails:
+        with pytest.raises(ValueError):
+            c("33")
+
+    def test_sugar(self):
+        """
+        `chain(c1, c2, c3)` and `[c1, c2, c3]` are equivalent.
+        """
+
+        @s
```
hynek: Please use the long names here. 😅

src/attr/converters.py
```diff
@@ -7,6 +7,26 @@
 from ._make import NOTHING, Factory
 
 
+def chain(*converters):
```
hynek: Please don't define it twice. Just import it from ._make just like we do with validators._and.

NI1993: Done

tests/test_converters.py
NI1993: Done

NI1993:
> Looks good overall, please address what I've mentioned.
> 
> The answer to how to test a nested function: you find a way to exercise it. :) But I believe you do that now?

Well I did, but it was mentioned that I missed testing the nested function + the codecov check failed so I just wondering how can I pass it

src/attr/converters.py
NI1993: Done

tests/test_converters.py
```diff
@@ -95,3 +99,39 @@ def test_none_factory(self):
...
+
+    def test_sugar(self):
+        """
+        `chain(c1, c2, c3)` and `[c1, c2, c3]` are equivalent.
+        """
+
+        @attr.s
+        class C(object):
+            a1 = attrib("a1", validator=chain(str, strtobool))
+            a2 = attrib("a2", validator=[str, strtobool])
```
hynek: Yeah I was wondering why you're missing coverage and here we have it. :)
```suggestion
            a1 = attrib("a1", converter=chain(str, strtobool))
            a2 = attrib("a2", converter=[str, strtobool])
```

tests/test_converters.py
```diff
@@ -95,3 +99,39 @@ def test_none_factory(self):
...
+
+        assert C.__attrs_attrs__[0].converter == C.__attrs_attrs__[1].converter
```
hynek: This won't work anymore because we're returning a closure.  You'll have to exercise it to demonstrate they do the same thing I guess.

tests/test_converters.py
NI1993: My bad, Thanks!

tests/test_converters.py
NI1993: Did so

NI1993:
@hynek it seems I still miss some..
What I'm I missing here :thinking:

hynek:
Looks like either out-of-date branch or another codecov glitch.

I think we're done here, thank you for your patience!

NI1993:
@hynek Many thanks!

PR #620, opened 2020-01-27
Use a wrapper class for cache hash to prevent cache serialization

### LINKED ISSUE #494: Allow cache_hash=True to be used when a custom __setstate__ is present

### LINKED ISSUE #613: cache_hash breaks copy and pickle for non-slots classes

pganssle:
When reviewing #615, @godlygeek came up with this brilliant alternative suggestion: use an `int` subclass that always pickles as `None`, so that unless someone goes out of their way to capture the cache value and include it in a `__reduce__` or `__setstate__`, the cache will be cleared on any copy operation.

My main fears (both of which turned out to be unsubstantiated) were:

1. This might slow down `hash()` in favor of correctness in serialization/deserialization for people implementing their own `__reduce__` functions, which seems like a very niche concern.
2. Returning a `_CacheHashValue` from `__hash__` might cause other problems, like if someone were taking the hash value of something and pickling or copying it for some deliberate reason (i.e. not copying the object it's stored on, but copying the return value of `hash(x)`).

In terms of performance, it seems like the only performance impact this has is on the *first* call to `hash()`, with:

```python
import attr
class A:
    x = attr.ib()
```

For `%timeit hash(A(1))`, I get around 800ns with the current version of attrs and ~1-1.2μs with this change, so about a 300-400ns cost on cache misses. It's a decently high percentage of the hash for this class, but it is kinda stupid to use `cache_hash` for this class if you're mostly getting cache misses anyway...

For `a = A(1)`, `%timeit hash(a)`, I get around 220-250ns per call with the old and new versions.

Regarding point 2, returning an integer subclass here instead of an integer doesn't seem to be a problem, because `hash()` casts the value of `__hash__` to `int` anyway, at least on CPython.

One alternate implementation I tried out is to use a 1-slot slots class instead of an integer subclass for `_CacheHashWrapper`, and unwrap the value on every call to `__hash__`. This was a bit slower than the integer subclass version in both the cache miss and cache hit cases. It was around 275ns in the cache hit case instead of 220-250ns.

One last thing to note: There is one use case that this is likely to break, which is people who are creating an `attrs` class with one version of `attrs`, pickling it, and then unpickling it in another version. Pickles made on Python 3 even with the same attrs version cannot be unpickled in Python 2 (though Python 2 pickles will probably work in Python 3). Since using pickle other than to send data between identical Python environments is unsupported and a bad idea, I think we can discount this concern.

I think this is probably the best approach. If you give me the green light I'll tweak whatever documentation needs to be tweaked and add a changelog. I actually don't expect many if any documentation changes with this version since there should be nothing to caution people about - we would just need to remove any language about #494.

This PR includes the additional tests added in #615.

Fixes #613.
Fixes #494.

# Pull Request Check List

### DIFF
--- a/changelog.d/620.change.rst
+++ b/changelog.d/620.change.rst
@@ -1,0 +1,8 @@
+Fixed serialization behavior of non-slots classes with ``cache_hash=True``.
+The hash cache will be cleared on operations which make "deep copies" of instances of classes with hash caching,
+though the cache will not be cleared with shallow copies like those made by ``copy.copy()``.
+
+Previously, ``copy.deepcopy()`` or serialization and deserialization with ``pickle`` would result in an un-initialized object.
+
+This change also allows the creation of ``cache_hash=True`` classes with a custom ``__setstate__``,
+which was previously forbidden (`#494 <https://github.com/python-attrs/attrs/issues/494>`_).
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -77,7 +77,7 @@ class _CacheHashWrapper(int):
     This is used for non-slots classes with ``cache_hash=True``, to avoid
     serializing a potentially (even likely) invalid hash value. Since ``None``
     is the default value for uncalculated hashes, whenever this is copied,
-    the copy's value for hte hash should automatically reset.
+    the copy's value for the hash should automatically reset.
 
     See GH #613 for more details.
     """
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -70,6 +70,31 @@ Sentinel to indicate the lack of a value when ``None`` is ambiguous.
 """
 
 
+class _CacheHashWrapper(int):
+    """
+    An integer subclass that pickles / copies as None
+
+    This is used for non-slots classes with ``cache_hash=True``, to avoid
+    serializing a potentially (even likely) invalid hash value. Since ``None``
+    is the default value for uncalculated hashes, whenever this is copied,
+    the copy's value for hte hash should automatically reset.
+
+    See GH #613 for more details.
+    """
+
+    if PY2:
+        # For some reason `type(None)` isn't callable in Python 2, but we don't
+        # actually need a constructor for None objects, we just need any
+        # available function that returns None.
+        def __reduce__(self, _none_constructor=getattr, _args=(0, "", None)):
+            return _none_constructor, _args
+
+    else:
+
+        def __reduce__(self, _none_constructor=type(None), _args=()):
+            return _none_constructor, _args
+
+
 def attrib(
     default=NOTHING,
     validator=None,
@@ -522,34 +547,6 @@ class _ClassBuilder(object):
         # Attach our dunder methods.
         for name, value in self._cls_dict.items():
             setattr(cls, name, value)
-
-        # Attach __setstate__. This is necessary to clear the hash code
-        # cache on deserialization. See issue
-        # https://github.com/python-attrs/attrs/issues/482 .
-        # Note that this code only handles setstate for dict classes.
-        # For slotted classes, see similar code in _create_slots_class .
-        if self._cache_hash:
-            existing_set_state_method = getattr(cls, "__setstate__", None)
-            if existing_set_state_method:
-                raise NotImplementedError(
-                    "Currently you cannot use hash caching if "
-                    "you specify your own __setstate__ method."
-                    "See https://github.com/python-attrs/attrs/issues/494 ."
-                )
-
-            # Clears the cached hash state on serialization; for frozen
-            # classes we need to bypass the class's setattr method.
-            if self._frozen:
-
-                def cache_hash_set_state(chss_self, _):
-                    object.__setattr__(chss_self, _hash_cache_field, None)
-
-            else:
-
-                def cache_hash_set_state(chss_self, _):
-                    setattr(chss_self, _hash_cache_field, None)
-
-            cls.__setstate__ = cache_hash_set_state
 
         return cls
 
@@ -1103,7 +1100,23 @@ def _make_hash(cls, attrs, frozen, cache_hash):
     unique_filename = _generate_unique_filename(cls, "hash")
     type_hash = hash(unique_filename)
 
-    method_lines = ["def __hash__(self):"]
+    hash_def = "def __hash__(self"
+    hash_func = "hash(("
+    closing_braces = "))"
+    if not cache_hash:
+        hash_def += "):"
+    else:
+        if not PY2:
+            hash_def += ", *"
+
+        hash_def += (
+            ", _cache_wrapper="
+            + "__import__('attr._make')._make._CacheHashWrapper):"
+        )
+        hash_func = "_cache_wrapper(" + hash_func
+        closing_braces += ")"
+
+    method_lines = [hash_def]
 
     def append_hash_computation_lines(prefix, indent):
         """
@@ -1111,14 +1124,18 @@ def _make_hash(cls, attrs, frozen, cache_hash):
         Below this will either be returned directly or used to compute
         a value which is then cached, depending on the value of cache_hash
         """
+
         method_lines.extend(
-            [indent + prefix + "hash((", indent + "        %d," % (type_hash,)]
+            [
+                indent + prefix + hash_func,
+                indent + "        %d," % (type_hash,),
+            ]
         )
 
         for a in attrs:
             method_lines.append(indent + "        self.%s," % a.name)
 
-        method_lines.append(indent + "    ))")
+        method_lines.append(indent + "    " + closing_braces)
 
     if cache_hash:
         method_lines.append(tab + "if self.%s is None:" % _hash_cache_field)
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -609,11 +609,10 @@ class _ClassBuilder(object):
             __bound_setattr = _obj_setattr.__get__(self, Attribute)
             for name, value in zip(state_attr_names, state):
                 __bound_setattr(name, value)
-            # Clearing the hash code cache on deserialization is needed
-            # because hash codes can change from run to run. See issue
-            # https://github.com/python-attrs/attrs/issues/482 .
-            # Note that this code only handles setstate for slotted classes.
-            # For dict classes, see similar code in _patch_original_class .
+
+            # The hash code cache is not included when the object is
+            # serialized, but it still needs to be initialized to None to
+            # indicate that the first call to __hash__ should be a cache miss.
             if hash_caching_enabled:
                 __bound_setattr(_hash_cache_field, None)
 
--- a/tests/test_dunders.py
+++ b/tests/test_dunders.py
@@ -314,17 +314,27 @@ class TestAddRepr(object):
 # they need to be out here so they can be un-pickled
 @attr.attrs(hash=True, cache_hash=False)
 class HashCacheSerializationTestUncached(object):
-    foo_value = attr.ib(default=20)
+    foo_value = attr.ib()
 
 
 @attr.attrs(hash=True, cache_hash=True)
 class HashCacheSerializationTestCached(object):
-    foo_value = attr.ib(default=20)
+    foo_value = attr.ib()
 
 
 @attr.attrs(slots=True, hash=True, cache_hash=True)
 class HashCacheSerializationTestCachedSlots(object):
-    foo_value = attr.ib(default=20)
+    foo_value = attr.ib()
+
+
+class IncrementingHasher(object):
+    def __init__(self):
+        self.hash_value = 100
+
+    def __hash__(self):
+        rv = self.hash_value
+        self.hash_value += 1
+        return rv
 
 
 class TestAddHash(object):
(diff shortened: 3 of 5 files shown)


gabbard:
This is a clever solution!

src/attr/_make.py
```diff
@@ -70,6 +70,30 @@ def __repr__(self):
 """
 
 
+class _CacheHashWrapper(int):
+    """An integer subclass that pickles / copies as None
+
+    This is used for non-slots classes with ``cache_hash=True``, to avoid
+    serializing a potentially (even likely) invalid hash value. Since ``None``
+    is the default value for uncalculated hashes, whenever this is copied,
+    the copy's value for hte hash should automatically reset.
```
gabbard:
```suggestion
    the copy's value for the hash should automatically reset.
```

tests/test_dunders.py
```diff
@@ -310,6 +310,33 @@ def test_str_no_repr(self):
...
+class HashCacheSerializationTestCached(object):
+    foo_value = attr.ib()
+
+
+@attr.attrs(slots=True, hash=True, cache_hash=True)
+class HashCacheSerializationTestCachedSlots(object):
+    foo_value = attr.ib()
+
+
+class IncrementingHasher(object):
```
gabbard: This is a nice solution to the testing need below.

src/attr/_make.py
```diff
@@ -70,6 +70,30 @@ def __repr__(self):
...
+    is the default value for uncalculated hashes, whenever this is copied,
+    the copy's value for the hash should automatically reset.
+
+    See GH #613 for more details.
+    """
+
+    if PY2:
+        # For some reason `type(None)` isn't callable in Python 2, but we don't
+        # actually need a constructor for None objects, we just need any
+        # available function that returns None.
```
wsanchez: FWIW, Python 2.7 has `types.NoneType`:

```console
Python 2.7.16 (default, Nov  9 2019, 05:55:08) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.32.4) (-macos10.15-objc-s on darwin
>>> type(None)
<type 'NoneType'>
>>> import types
>>> types.NoneType
<type 'NoneType'>
```

Interestingly, while `NoneType` in Python3 is a thing, I don't know how to reach it:

```console
Python 3.8.0 (v3.8.0:fa919fdf25, Oct 14 2019, 10:23:27) 
[Clang 6.0 (clang-600.0.57)] on darwin
>>> type(None)
<class 'NoneType'>
>>> import types
>>> types.NoneType
AttributeError: module 'types' has no attribute 'NoneType'
>>> 
```

src/attr/_make.py
```diff
@@ -70,6 +70,30 @@ def __repr__(self):
...
+        def __reduce__(self):
+            return (getattr, (0, "", None))
```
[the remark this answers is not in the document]
godlygeek: I'd be curious to know how that compares from a performance point of view.

src/attr/_make.py
godlygeek: `types.NoneType` isn't helpful to us here in Python 2, since it isn't callable:

```
>>> types.NoneType()
TypeError: cannot create 'NoneType' instances
```

Why that doesn't just return the singleton instance is beyond me, but oh well...

src/attr/_make.py
```diff
@@ -1103,7 +1099,10 @@ def _make_hash(cls, attrs, frozen, cache_hash):
     unique_filename = _generate_unique_filename(cls, "hash")
     type_hash = hash(unique_filename)
 
-    method_lines = ["def __hash__(self):"]
+    method_lines = [
+        "from attr._make import _CacheHashWrapper",
```
godlygeek: I don't think I've ever seen an import inside a class definition before - this is a clever way to work around the fact that we need to access something from inside `attr` but this code will be `eval`'d in a namespace we don't own and can't control.  But, to avoid subclasses seeing this, it should probably be `del`'d after we're done defining `__hash__`.  An alternative would be:

```py
def __hash__(self, _cache_wrapper=__import__('attr._make')._make._CacheHashWrapper):
```

That's one line instead of 3, and `__import__` is maybe a bit less weird than an import inside a class body - at least, I've seen it more.

src/attr/_make.py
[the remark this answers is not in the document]
llllllllll: Here is a not super scientific check:
b
```Python
In [64]: class C: 
    ...:     def __reduce__(self): 
    ...:         return getattr, (0, '', None) 
    ...:                                                                                   

In [65]: class D: 
    ...:     def __reduce__(self): 
    ...:         return type(None), () 
    ...:                                                                                   

In [66]: def return_none(): 
    ...:     return None 
    ...: class E: 
    ...:     def __reduce__(self): 
    ...:         return return_none, () 
    ...:                                                                                   

In [67]: len(pickle.dumps(C()))                                                            
Out[67]: 39

In [68]: len(pickle.dumps(D()))                                                            
Out[68]: 31

In [69]: len(pickle.dumps(E()))                                                            
Out[69]: 31

In [70]: %timeit pickle.loads(pickle.dumps(C()))                                           
3.33 µs ± 14.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [71]: %timeit pickle.loads(pickle.dumps(D()))                                           
3.12 µs ± 14.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [72]: %timeit pickle.loads(pickle.dumps(E()))                                           
2.62 µs ± 39.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

src/attr/_make.py
wsanchez: Ah, yes.

src/attr/_make.py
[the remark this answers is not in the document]
hynek: Looks like return_none is almost 20% faster? Which makes sense, since it doesn't do any function calls itself. 🤔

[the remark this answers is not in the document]
hynek: Could you try to additionally define the method in the class scope please? Global lookups are relatively slow.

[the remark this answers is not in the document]
godlygeek: I don't think defining it as a class method will work - at least in Python 2, you can't use a class method as a constructor and will get something like:

```
> pickle.PicklingError: Can't pickle <function _return_none at 0x7f7b13d250c8>: it's not found as __main__._return_none
```

But, this works to avoid the global lookup:

```py
def _return_none(): 
    return None 
class E: 
    def __reduce__(self, _return_none=_return_none): 
        return _return_none, () 
```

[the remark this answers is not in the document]
pganssle: I think we probably shouldn't be optimizing based on the speed of `pickle.loads(pickle.dumps())` - almost always if you're using pickle then you are using it in an RPC or writing it to a file. Maybe on-device RPC calls through a unix socket or something are fast enough that the speed of `__reduce__` calls matters, but I think if we're optimizing for speed then `copy.copy()` is a better metric.  When I time `copy.copy`, all of these classes give me approximately the same speed.

If we're really micro-optimizing this function, I'd say that `copy.copy` should be the benchmark for speed and `len(pickle.dumps(x))` should be the benchmark for  size.

That said, it would be unusual to use `cache_hash` for a particularly small object in the first place. For a big enough object that calculating the hash is expensive, you'd also expect it to take a while to pickle it, so an extra 200ns is going to be noise on the process.

> It may be more clear to create a module scope function like:

I don't know what @hynek's plans are for Python 2 support, but I assume that it won't be too long before Python 2 is gone. I feel like it would preferable to use the built-in `NoneType` in Python 3 since that's already a builtin that does the exact same thing. I like the `getattr` hack because it doesn't leave any compatibility functions lying around as critical components of the system, at the cost of needing a comment and having the pickle be a little bigger.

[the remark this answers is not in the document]
godlygeek: The biggest downside I see to the `def _return_none()` approach is that the pickled object will contain a reference to `attr._make._return_none`, meaning it'll be larger than necessary and unnecessarily fragile in the case of pickling with one version of `attr` and unpickling with a different one (should this ever get refactored in the future).  I think the best balance of maintainability, performance, and future-proofness might be something like:

```py
class UnpicklableInt(int):
    if six.PY2:
        def __reduce__(self, _none_constructor=getattr, _args=(0, "", None)):
            return _none_constructor, _args
    else:
        def __reduce__(self, _none_constructor=type(None), _args=()):
            return _none_constructor, _args
```

That uses a global function for Python 2

pganssle:
So just to throw one *potential* monkey wrench into the mix here. It turns out that `copy.copy()` only calls `__reduce__` on the instance - it does *not* recursively call `__reduce__` on all the attributes, you need to call `copy.deepcopy()` in order for that to happen.

I tend to think that this is fine, since we're mainly worried about the hash cache persisting between different interpreter runs, and if a shallow copy preserves the hash cache that shouldn't be so bad, particularly because `cache_hash=True` doesn't work on classes with a custom hash or where the auto-generated hash is based on the `id` of the object. Even if the auto-generated hash relies on the `id` of the attributes, a `copy.copy` call won't change that anyway, so I think the behavior is still sane.

src/attr/_make.py
pganssle: Apparently this would normally give import errors because `_make_hash(` is actually called from `attr._make`, leading to a dependency cycle that apparently is resolved by using the `import` syntactic sugar.

However, that led me to the realization that the current code unconditionally imports `_CacheHashWrapper` and unconditionally wraps the hash value in `_cache_wrapper`, even for `cache_hash=False` classes.

Since the calls to `_make_hash` in `attr` don't use `cache_hash=True`, switching over to conditionally including those features (which should be done anyway) fixes that problem.

src/attr/_make.py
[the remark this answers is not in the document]
pganssle: I like @godlygeek's version, so I updated the code to use that. I was hoping to make `_none_constructor` and `_args` keyword-only arguments in Python 3, but that's not possible in 2/3 compatible code because it's a syntax error even in the "dead code" branches. Alas.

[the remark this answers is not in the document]
hynek: > I don't know what @hynek's plans are for Python 2 support, but I assume that it won't be too long before Python 2 is gone.

I wish: https://www.attrs.org/en/stable/python-2.html

However the rule is: optimize for Python 3.  Make Python 2 work sOmEhOw.

hynek:
As far as my limited understand of the whole topic goes, this looks mostly good to me – modulo missing newsfragment. Is it still “WIP”? Has anyone any quibbles left?

pganssle:
@hynek I'm going to give the documentation a once-over to make sure that this doesn't invalidate any existing documentation and to see if it makes sense to add some documentation for this behavior, but I'll remove the WIP since the "to-do" part of it is captured adequately by the checklist.

pganssle:
@hynek I've added a changelog entry. Looking through the docs, I didn't see anywhere that the old `__setstate__` + `cache_hash=True` = exception behavior was documented, or any other places that would obviously need updating with this change. Let me know if it's too long or if you want any other adjustments.

I wasn't sure how much of this new behavior you want me to officially document outside the changelog. Should I add a `..versionchanged:: 20.1.0` to `cache_hash`? Should I maybe add some information about this to the [hash code caching](https://www.attrs.org/en/stable/hashing.html#hash-code-caching) page? Up to you how you want to do it. We can also kick the can down the road by merging this and then spawn a new issue for improving the documentation around cache hashing.

Barring any documentation updates, I think this is ready to go.

changelog.d/620.change.rst
```diff
@@ -0,0 +1,5 @@
+Fixed serialization behavior of non-slots classes with ``cache_hash=True``. Operations which make "deep copies" of instances of classes with hash caching, the hash cache will be cleared on copy (though the cache will not be cleared with shallow copies like those made by ``copy.copy``).
```
hynek: Sorry I hate it but I have to be a bit nitpicky here. :(

- Firstly: we probably should delete `changelog.d/611.change.rst`?
- 1 sentence per line please.
- functions like `copy.copy` please write with () i.e. `copy.copy()`

changelog.d/620.change.rst
```diff
@@ -0,0 +1,5 @@
...
+
+Previously, ``copy.deepcopy`` or serialization and deserialization with ``pickle`` would result in an un-initialized object.
```
hynek: Same with `copy.deepcopy`.

changelog.d/620.change.rst
```diff
@@ -0,0 +1,5 @@
...
+
+This change also allows the creation of ``cache_hash=True`` classes with a custom ``__setstate__``, which previously forbidden (#494).
```
hynek: - Probably "was" forbidden?
- Please make #494 a link to the PR

src/attr/_make.py
```diff
@@ -70,6 +70,30 @@ def __repr__(self):
 """
 
 
+class _CacheHashWrapper(int):
+    """An integer subclass that pickles / copies as None
```
hynek: Newline after `"""` please.

pganssle:
@hynek Thanks for the nitpicking, I admit I'm often a bit sloppy about stylistic issues (especially with `black` allowing me to throw caution to the wind in like 95% of situations), so it's helpful to get feedback about the conventions of the project.

Your comments prompted me to actually finish reading through the "semantic newlines" document and I realized I was doing it wrong (plus I decided to reword the changelog to read a bit naturally) - let me know if I've got it wrong.

hynek:
I think we're good! Except that codecov is acting up once again but hey that's rock'n'roll. Thanks everybody involved!

Issue #435, opened 2018-08-26
`cmp`ing fails when attrib is a numpy array


meereeum:
The generated `cmp` methods fail if the attrib is a numpy array (of size >1) with
```py
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
this is because:
```py
# equivalent to how __eq__compares fields:
(np.array([47,47],) == (np.array([47,47],) # ValueError
# vs:
(np.array([47,47]) == np.array([47,47])).all() # True
```

I realize that this can be switched off with `cmp=False`, but often comparing these attribs is useful!

reproducible example:
```py
@attr.s(auto_attribs=True)
class A:
    arr: np.ndarray

foo = A(np.array([47,47]))
bar = A(np.array([47,47]))

foo == bar # ValueError
```

the simplest fix might be to trust user annotations about what is/isn't a numpy array and check those separately. or, maybe better, would be awesome to supply a custom `cmp` function, e.g.
```py
@attr.s
class A:
    arr: np.ndarray = attr.ib(cmpf=lambda f,x,y: f(x,y).all())
```
where `f` is the standard cmp function (`eq`, `lt`, etc) and `x,y` are the items being compared.

jriddy:
I've found this wildly frustrating myself, but I think this is a problem with `numpy` breaking normal `__eq__` semantics, and `attrs` actually shouldn't try and solve this. Normally the signature of `__eq__` should match [`object.__eq__`'s signature](https://github.com/python/typeshed/blob/a95519d8e6d475fab970686700a1a2aa0b96498f/stdlib/3/builtins.pyi#L44) but numpy goes its own way:

```python
# object
def __eq__(self, other: object) -> bool: ...

# numpy
def __eq__(self, other: np.ndarray) -> np.ndarray: ...
```

This is so you can get the equality product of two arrays easily. And every data scientist I talk to seems to think this is the most natural thing in the world. I deeply disagree with this design decision, but I think the reasoning behind this is clear: in numpy, everything operates on arrays, normal python language semantics be damned!

[...]

PR #625, opened 2020-02-15
Allow user to customize Attribute comparison by supplying a *cmpspec*

botant:
I've implemented the initial part of issue #435. I wanted to stop at this point and get some feedback in case you guys think I'm going in the wrong direction.

The main idea of how this feature would be used can be found in ``test_cmp.py``.

I did the basics of the check list below, and I tried my best to update the ``.pyi`` but I'm not sure whether what I did is correct. If this goes through, I'll update the documentation once we have all features down.

- [**?**] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/master/tests/strategies.py).
    - [**not yet**] ...and used in the stub test file `tests/typing_example.py`.
- [**not yet**] Updated **documentation** for changed code.
    - [**not yet**] New functions/classes have to be added to `docs/api.rst` by hand.
- [**not yet**] Documentation in `.rst` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -109,6 +109,7 @@ def attrib(
     kw_only=False,
     eq=None,
     order=None,
+    cmpspec=None,
 ):
     """
     Create a new attribute on a class.
@@ -200,6 +201,10 @@ def attrib(
         in the generated ``__init__`` (if ``init`` is ``False``, this
         parameter is ignored).
 
+    :param cmpspec: Class used to compare attribute. It must be initializable
+        with the value of the attribute, and implement equality
+        (e.g. ``__eq__``) and ordering (e.g. ``__lt__'') functions.
+
     .. versionadded:: 15.2.0 *convert*
     .. versionadded:: 16.3.0 *metadata*
     .. versionchanged:: 17.1.0 *validator* can be a ``list`` now.
@@ -216,6 +221,7 @@ def attrib(
     .. versionchanged:: 19.2.0 *repr* also accepts a custom callable.
     .. deprecated:: 19.2.0 *cmp* Removal on or after 2021-06-01.
     .. versionadded:: 19.2.0 *eq* and *order*
+    .. versionadded:: 20.1.0.dev0_botant *cmpspec*
     """
     eq, order = _determine_eq_order(cmp, eq, order)
 
@@ -250,6 +256,7 @@ def attrib(
         kw_only=kw_only,
         eq=eq,
         order=order,
+        cmpspec=cmpspec,
     )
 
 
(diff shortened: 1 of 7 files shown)


jriddy:
Does doing this at the class level make sense? It's my intuition that the problems are individual attributes, which can be of types that have ~ill-conceived~ non-standard behavior of the comparison methods.

This is an honest question I have; I don't know the right answer. It may just be too unergonomic to specify comparison types on each attribute, and then too difficult to efficiently generate a comparison method from that.

botant:
Hi Josh,

Apologies, maybe I got the naming convention mixed up and the text of the
PR is not what I meant.

It is at the level of individual attributes, not the class decorated with
attr.s( ).

Something like this:

@attr.s()
class MyClass:
   my_att = attr.ib(cmpspec=MySpec)


Cheers!
Antonio

jriddy:
Okay awesome! This will form a good base for a more ergonomic API then!

I think the target API for this should be something very similar to how validators and default providers work: they give you the choice of a `keyword=provider` method or a decorator based `@attribute.keyword` method. And I don't think this would be hard with what you've done, something like...

```python
@attr.s
class ColorspacedImage:
    space: str = attr.ib()
    data: np.ndarray = attr.ib()

    @data.comparator
    class _NdCmp:
        def eq(self, inst, other):
            return (inst == other).all()
        def ne(self, inst, other):
            return not self.eq(inst, other)
        ...
```

I don't think we need to support all that now, but I could see that as a goal and direction to take this in.

I'll start working on a PR to this to show what I mean.

botant:
Yes, that is what I had in mind as well, something consistent with
attr.validators and attr.converters.

The ``@data.comparator`` decorator is a great idea, especially together
with ``@functools.total_ordering``.

src/attr/_make.py
```diff
@@ -1190,52 +1210,43 @@ def __ne__(self, other):
...
+        """
+        return _attrs_to_cmp_tuple(obj, attrs)
+
     # We can't just do a big self.x = other.x and... clause due to
     # irregularities like nan == nan is false but (nan,) == (nan,) is true.
-    if attrs:
-        lines.append("    return  (")
-        others = ["    ) == ("]
-        for a in attrs:
-            lines.append("        self.%s," % (a.name,))
```
jriddy: I think we should strive to keep this generated code behavior here if possible. It looks like a bit of work and tuning was put into getting a bunch of details right (for speed, debugging, serialization), and I don't want to throw that out to solve and edgier edge case. I have an idea, i'll try it in my coming PR

jriddy:
I will also say that the more I look at this, the less I like "cmpspec" as a word for this concept. It's just clumsy and it doesn't make a pleasant sound in my head. How do you feel about "comparator"?

botant:
Yes, cmpspec doesn't exactly roll of the tongue, does it? I don't like comparator either, mainly because it isn't a word (outside electronics) but at least it is in line with validator and converter...

jriddy:
Okay I'll toss it around in my head, see if I can't come up with something
we can all like better. I would have suggested just using `cmp` and
triggering the advanced behavior on whether that attribute is a class, but
they've deprecated that attribute and I think that might be confusing now.

Which reminds me... I'm thinking we should also define the way that this
feature interacts with `eq` and `order`, with my gut telling me to just
raise an error if you try to use this rich comparison class and those are
not `None`.

botant:
I had a different take on the interaction between `comparator` and `eq` + `order`. We have already been through one round of deprecation from `cmp` to `eq` + `order`, so maybe we should avoid a second one.

How about keeping `eq` and `order` exactly as they are now, i.e. their values determine whether a given attribute is included in the dunder comparison methods, whilst `comparator` provides the *how*?

jriddy:
Hmm that actually makes sense, especially if you think of providing default
comparators for some common use cases (numpy, pandas, etc)

Issue #626, opened 2020-02-21
``__ne__`` dunders changing?


botant:
I found that ``__ne__`` dunders are changing during the lifetime of a class. 

I'm using Python 3.7 on Ubuntu and the latest version of ``attrs``.

To reproduce the issue, try this:
```
import attr
A = attr.make_class('A', [], eq=True)
print((id(A.__eq__), A.__eq__.__module__, A.__eq__.__qualname__, id(A.__ne__), A.__ne__.__module__, A.__ne__.__qualname__))
B = attr.make_class('B', [], eq=True)
print((id(A.__eq__), A.__eq__.__module__, A.__eq__.__qualname__, id(A.__ne__), A.__ne__.__module__, A.__ne__.__qualname__))
```

I believe this is happening because ``_make_eq`` returns the module-defined ``__ne__``:

```python
def __ne__(self, other):
    """
    Check equality and either forward a NotImplemented or return the result
    negated.
    """
    result = self.__eq__(other)
    if result is NotImplemented:
        return NotImplemented

    return not result


def _make_eq(cls, attrs):
   ...
    return local_vars["__eq__"], __ne__
```

This should be easy to solve. If someone can confirm whether this is indeed unintended, I'll create a PR this weekend.

Thanks!

hynek:
Ah yes, changing qualnames are unfortunate!

botant:
Can I submit a PR for this?

hynek:
sure!

hynek:
This has been fixed.

PR #632, opened 2020-03-07
Wrapped __ne__ in a _make_ne function #626

botant:
Wrapped `__ne__` in a `_make_ne` function to guarantee that each attrs class gets a new instance of the `__ne__` function.

# Pull Request Check List

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing left to do.


botant:
A note on this change: `_make_ne` is called by `_make_eq`:
```python
def _make_eq(cls, attrs):
    """
    Create *equal* method.
    """
    ...
    return locs["__eq__"], _make_ne()
```

Another way of writing this would be:
```python
def _make_eq(cls, attrs):
    """
    Create *equal* method.
    """
    ...
    return locs["__eq__"]

def _add_eq(cls, attrs=None):
    """
    Add equality methods to *cls* with *attrs*.
    """
    if attrs is None:
        attrs = cls.__attrs_attrs__

    cls.__eq__ = _make_eq(cls, attrs)
    cls.__ne__ = _make_ne()

    return cls
```

Arguably the second way looks nicer and makes more sense.  Let me know if you think the solution makes sense, and which solution looks nicer to you.

Thanks

hynek:
Yeah sure, second version looks nicer.  Just don't call it `*equal*` please, asterisks are used for arguments by convention. Just say `__eq__` and `__ne__` and be done with it. :) For order, just call them `ordering methods`.

PR #635, opened 2020-03-30
Fix attribute collection

### LINKED ISSUE #428: MRO vs. inherited attributes
The way how attributes are collected from super classes is different comparing to how attributes and methods are searched by `getattr` (based on MRO for new-style classes).

Example:
```python
@attr.s
class A(object):

    x = attr.ib(10)

    def xx(self):
        return 10


@attr.s
class B(A):
    y = attr.ib(20)


@attr.s
class C(A):
    x = attr.ib(50)

    def xx(self):
        return 50


class D(B, C):
    pass


d = D()
print(d.x)  # prints 10
print(d.xx())  # prints 50
```

I think it would be great to the use the same approach for collecting attributes as used for searching methods and attributes. 

The difference is caused by the fact that the function `_transform_attrs` in `attrs/_make.py` considers all attributes (own+inherited stored in `__attrs_attrs__`) and not only own attributes of super classes: https://github.com/python-attrs/attrs/blob/6a07b035b77ea8756408d65a36160f8670c66933/src/attr/_make.py#L367-L368

hynek:
This fixes #428 and has pestered me for a long while.

I'm not entirely sure how to go forward with the option, I guess I'll just set it to `True` in `@attrs.define`?

Anyhow, I this is the penultimate ticket stopping me from tackling the new APIs…

### DIFF
--- a/changelog.d/428.change.rst
+++ b/changelog.d/428.change.rst
@@ -1,4 +1,4 @@
-Added ``@attr.s(collect_by_mro=False`` argument that if set to ``True`` fixes the collection of attributes from base classes.
+Added ``@attr.s(collect_by_mro=False)`` argument that if set to ``True`` fixes the collection of attributes from base classes.
 
 It's only necessary for certain cases of multiple-inheritance but is kept off for now for backward-compatibility reasons.
 It will be turned on by default in the future.
--- a/changelog.d/428.change.rst
+++ b/changelog.d/428.change.rst
@@ -1,0 +1,6 @@
+Added ``@attr.s(collect_by_mro=False`` argument that if set to ``True`` fixes the collection of attributes from base classes.
+
+It's only necessary for certain cases of multiple-inheritance but is kept off for now for backward-compatibility reasons.
+It will be turned on by default in the future.
+
+As a side-effect, ``attr.Attribute`` now *always* has an ``inherited`` attribute indicating whether an attribute on a class was directly defined or inherited.
--- a/changelog.d/635.change.rst
+++ b/changelog.d/635.change.rst
@@ -1,4 +1,4 @@
-Added ``@attr.s(collect_by_mro=False`` argument that if set to ``True`` fixes the collection of attributes from base classes.
+Added ``@attr.s(collect_by_mro=False)`` argument that if set to ``True`` fixes the collection of attributes from base classes.
 
 It's only necessary for certain cases of multiple-inheritance but is kept off for now for backward-compatibility reasons.
 It will be turned on by default in the future.
--- a/changelog.d/635.change.rst
+++ b/changelog.d/635.change.rst
@@ -1,0 +1,6 @@
+Added ``@attr.s(collect_by_mro=False`` argument that if set to ``True`` fixes the collection of attributes from base classes.
+
+It's only necessary for certain cases of multiple-inheritance but is kept off for now for backward-compatibility reasons.
+It will be turned on by default in the future.
+
+As a side-effect, ``attr.Attribute`` now *always* has an ``inherited`` attribute indicating whether an attribute on a class was directly defined or inherited.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -405,7 +405,7 @@ def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
     If *these* is passed, use that and don't look for them on the class.
 
     *collect_by_mro* is True, collect them in the correct MRO order, otherwise
-     use the old -- incorrect -- order.  See #428.
+    use the old -- incorrect -- order.  See #428.
 
     Return an `_Attributes`.
     """
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1946,14 +1946,14 @@ class Attribute(object):
     """
     *Read-only* representation of an attribute.
 
-    :attribute ``name``: The name of the attribute.
-    :attribute ``inherited``: Whether or not that attribute has been inherited
-       from a base class.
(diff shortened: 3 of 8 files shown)


changelog.d/428.change.rst
```diff
@@ -0,0 +1,6 @@
+Added ``@attr.s(collect_by_mro=False`` argument that if set to ``True`` fixes the collection of attributes from base classes.
```
webknjaz: typo?
```suggestion
Added ``@attr.s(collect_by_mro=False)`` argument that if set to ``True`` fixes the collection of attributes from base classes.
```

changelog.d/635.change.rst
```diff
@@ -0,0 +1,6 @@
+Added ``@attr.s(collect_by_mro=False`` argument that if set to ``True`` fixes the collection of attributes from base classes.
```
webknjaz: dupe fragment?
```suggestion
Added ``@attr.s(collect_by_mro=False)`` argument that if set to ``True`` fixes the collection of attributes from base classes.
```

changelog.d/428.change.rst
```diff
@@ -0,0 +1,6 @@
+Added ``@attr.s(collect_by_mro=False)`` argument that if set to ``True`` fixes the collection of attributes from base classes.
```
webknjaz: Is it necessary to have two identical change fragments?

src/attr/_make.py
```diff
@@ -343,12 +343,70 @@ def _counter_getter(e):
...
+    base_attrs = []
+    base_attr_map = {}  # A dictionary of base attrs to their classes.
+
+    # Traverse the MRO and collect attributes.
+    for base_cls in reversed(cls.__mro__[1:-1]):
+        for a in getattr(base_cls, "__attrs_attrs__", []):
+            if not a.inherited and a.name not in taken_attr_names:
+                a = a._assoc(inherited=True)
+                base_attrs.append(a)
+                base_attr_map[a.name] = base_cls
```
webknjaz: It's probably a matter of taste but would you consider reducing the indentation?
```suggestion
            if a.inherited or a.name in taken_attr_names:
                continue
            a = a._assoc(inherited=True)
            base_attrs.append(a)
            base_attr_map[a.name] = base_cls
```

src/attr/_make.py
```diff
@@ -998,6 +1048,13 @@ def attrs(
           default value are additionally available as a tuple in the ``args``
           attribute,
         - the value of *str* is ignored leaving ``__str__`` to base classes.
+    :param bool collect_by_mro: Setting this to `True` fixes the way ``attrs``
```
webknjaz:
```suggestion
    :param bool collect_by_mro: Setting this to ``True`` fixes the way ``attrs``
```

src/attr/_make.py
```diff
@@ -1883,12 +1942,16 @@ class Attribute(object):
...
+    :attribute ``inherited``: Whether or not that attribute has been inherited
+       from a base class.
 
     Plus *all* arguments of `attr.ib` (except for ``factory``
     which is only syntactic sugar for ``default=Factory(...)``.
 
-    For the version history of the fields, see `attr.ib`.
+    ..versionadded:: 20.1.0 *inherited*
+
+    For the full version history of the fields, see `attr.ib`.
```
webknjaz:
```suggestion
    For the full version history of the fields, see ``attr.ib``.
```

changelog.d/635.change.rst
hynek: Heh thanks for both, i fixed it myself when i saw your suggestion. :)

changelog.d/428.change.rst
hynek: Yeah that's how towncrier works nowadays I'm afraid. It will detect that it's the same and coalesce them into one entry.

src/attr/_make.py
hynek: It's very much my taste; as you may have guessed there used to be more code and it just stayed that way.

src/attr/_make.py
hynek: This one is actually correct thanks to #571!

src/attr/_make.py
hynek: Same here. All glory to the default role.

tests/test_make.py
```diff
@@ -337,6 +337,92 @@ class E(C, D):
...
+        assert (
+            "E(a1='a1', a2='a2', d1='d1', d2='d2', b1='b1', b2='b2', c1='c1', "
+            "c2='c2', e1='e1', e2='e2')"
+        ) == repr(E())
+
+    def test_mro(self):
+        """
+        Attributes and methods are looked up the same way.
+
+        See #428
```
underyx:
```suggestion
        See https://github.com/python-attrs/attrs/issues/428
```
teensy tiny convenient lil' change

hynek: Generally I would agree, however I refer to it multiple times and that would get really noisy. :|

changelog.d/428.change.rst
webknjaz: Oh wow. I'm curious if it'd work with symlinks...

src/attr/_make.py
webknjaz: TIL

hynek: Yes, it's very cool!  I also only learned about it when Julian submitted the PR…

jacobg:
This commit doesn't seem to work with slots. See related issue: https://github.com/python-attrs/attrs/issues/667

hynek:
This commit is unrelated to #667.

PR #642, opened 2020-04-28
Add control to generate __[sg]etstate__

### LINKED ISSUE #512: Support custom __getstate__, __setstate__ for slotted classes (or improve docs)?

hynek:
Pass *getstate_setstate* to attr.s to control the generation.  Mostly interesting for disabling it in slotted classes that received autogenerated `__getstate__` and `__setstate__` by default until now.

Here we go @andhus – how does this look to you?

Fixes #512,#513


andhus:
LGTM 👍

PR #650, opened 2020-06-08
Document how to write aliases that are recognized by mypy

### LINKED ISSUE #630: Way to change defaults?
I read this page on the mypy docs, https://mypy.readthedocs.io/en/stable/additional_features.html which says that you can't simply write a function like this:

```
def my_dataclass(cls):
    return attr.s(cls, auto_attribs=True)
```

And expect things to type check correctly. 

Is there any convenient, mypy compatible way, to change defaults in attrs? I saw the RFC about defaults being inconvenient; better defaults (above all auto_attribs=True) will help but of course people will not necessarily agree on the best defaults, hence why being able to wrap it would be such a win.

It looks like pydantic has done it (https://github.com/samuelcolvin/pydantic/blob/master/pydantic/dataclasses.py), but it's hard for me to separate out the bits that are related to their extra functionality, from the bare minimum.

hynek:
This fIxEs  #630 as far as _we_ can fix it. I would love to add a more concrete example (a minimal example that people can just drop in and that explains _how_ to drop in) but I guess I'll leave that as a contribution opportunity to someone else since I don't have the bandwidth for that ATM.

### DIFF
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -74,6 +74,19 @@ At the moment, the best workaround is to hold your nose, write a fake mypy plugi
       return MyPlugin
 
 
+Then tell mypy about your plugin usin your project's ``mypy.ini``:
+
+.. code:: ini
+
+   [mypy]
+   plugins=<path to file>
+
+
+.. warning::
+   Please note that it is currently *impossible* to let mypy know that you've changed defaults like *eq* or *order*.
+   You can only use this trick to tell mypy that a class is actually an ``attrs`` class.
+
+
 Types
 -----
 
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -44,6 +44,35 @@ A more elegant way can be to wrap ``attrs`` altogether and build a class `DSL <h
 
 An example for that is the package `environ-config <https://github.com/hynek/environ-config>`_ that uses ``attrs`` under the hood to define environment-based configurations declaratively without exposing ``attrs`` APIs at all.
 
+Another common use case is to overwrite ``attrs``'s defaults.
+
+Unfortunately, this currently `confuses <https://github.com/python/mypy/issues/5406>`_ mypy's ``attrs`` plugin.
+At the moment, the best workaround is to hold your nose, write a fake mypy plugin, and mutate a bunch of global variables::
+
+   from mypy.plugin import Plugin
+   from mypy.plugins.attrs import (
+      attr_attrib_makers,
+      attr_class_makers,
+      attr_dataclass_makers,
+   )
+
+   # These work just like `attr.dataclass`.
+   attr_dataclass_makers.add("my_module.method_looks_like_attr_dataclass")
+
+   # This works just like `attr.s`.
+   attr_class_makers.add("my_module.method_looks_like_attr_s")
+
+   # These are our `attr.ib` makers.
+   attr_attrib_makers.add("my_module.method_looks_like_attrib")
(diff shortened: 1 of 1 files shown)


docs/extending.rst
```diff
@@ -44,6 +44,35 @@ A more elegant way can be to wrap ``attrs`` altogether and build a class `DSL <h
...
+   # This works just like `attr.s`.
+   attr_class_makers.add("my_module.method_looks_like_attr_s")
+
+   # These are our `attr.ib` makers.
+   attr_attrib_makers.add("my_module.method_looks_like_attrib")
+
+   class MyPlugin(Plugin):
+      # Our plugin does nothing but it has to exist so this file gets loaded.
+      pass
+
```
euresti, line 71: extra blank line?

docs/extending.rst
```diff
@@ -44,6 +44,35 @@ A more elegant way can be to wrap ``attrs`` altogether and build a class `DSL <h
...
+
+   def plugin(version):
+      return MyPlugin
+
```
euresti, line 75: If you want to add the change to mypy.ini
```
[mypy]
plugins=<path to file>
```

docs/extending.rst
```diff
@@ -44,6 +44,35 @@ A more elegant way can be to wrap ``attrs`` altogether and build a class `DSL <h
 
 An example for that is the package `environ-config <https://github.com/hynek/environ-config>`_ that uses ``attrs`` under the hood to define environment-based configurations declaratively without exposing ``attrs`` APIs at all.
 
+Another common use case is to overwrite ``attrs``'s defaults.
```
euresti, line 47: I worry about putting this line right next to the mypy stuff because there's no way change a default (e.g. `eq` or `order`.) The only thing the plugin lets you do is add a method that has the same defaults as three methods you're overriding.  But maybe it won't be a problem.

Tinche:
I actually had issues getting mypy to recognize my wrappers. I think the package implementing the wrappers needs to contain a `py.typed` marker file.

euresti:
Also, what would make us not hold our nose?  I have a couple of ideas:

1. Add a new decorator that does nothing but it lets mypy know that this method overrides an attrs method.  Something like:
```
@class_maker(auto_attribs_default=True)
def my_class_maker(...):
   ...
```
This looks great but is probably pretty hard in mypy.  Also how many defaults do we allow to set.  We'd want to iterate on the API.

2. Add a method that lets us not feel like it's not modifying a global.  e.g.
```
from mypy.plugins.attrs import register_extra_makers

class MyPlugin(Plugin):
    def __init__(self, options: Options) -> None:
        register_extra_makers(dataclass=..., attrs=..., attrib=...)
        super().__init__(options)        

def plugin(version):
    return MyPlugin
```

3. Maybe there's a way to do this in mypy.ini?  But this is not something I've investigated.  The mypy team might have special rules for what's allowed in the mypy.ini file.

docs/extending.rst
hynek: Agreed, that should be clarified. :|

docs/extending.rst
hynek: It's a function after a class so let's stick to black's standard. ;)

hynek:
> Also, what would make us not hold our nose? I have a couple of ideas:

This seems worth a news ticket I think!

I like both but someone intimate with mypy has to investigate what you wrote. :|

Issue #649, opened 2020-05-26
[RFC] Add optional automatic validation and conversion


sscherfke:
Validation and conversion of attribute values is currently only done if the user passes the corresponding arguments to `attrib()`.

In contrast, libs like [Pydantic](https://pydantic-docs.helpmanual.io/) automatically validate and convert all values (in `__init__()` as well as `__setattr__()`).  It also allows the user to easily pass the most common validators like `ge=0` or `maxlength=50`.

This lets you do fancy stuff like this, which is very useful, e.g., for REST API server models and clients, or generally for loading nested JSON data. Settings/config management is another usecase:

```python
>>> from dateatime import datetime                                                    
>>> from pydantic import BaseModel                                                    
>>>                                                                                   
>>> class Child(BaseModel):                                                           
...     x: int                                                                        
...     y: int                                                                        
...     d: datetime                                                                   
...                                                                                   
>>> class Parent(BaseModel):                                                          
...     name: str                                                                     
...     child: Child                                                                  
...                                                                                   
>>> data = {                                                                          
...     'name': 'spam',                                                               
...     'child': {                                                                    
...         'x': 23,                                                                  
...         'y': '42',  # sic!                                                        
...         'd': '2020-05-04T13:37:00',                                               
...     },                                                                            
... }                                                                                 
>>> Parent(**data)                                                                    
Parent(name='spam', child=Child(x=23, y=42, d=datetime.datetime(2020, 5, 4, 13, 37))) 
```

While pydantic works quite nicely, it is way slower than attrs and also uses inheritance which is not always what users may want.

I tried implementing a similar functionality but failed. IMHO, there are some changes needed directly within attr’s class-buildng facilities, as the following example illustrates:
```python
from datetime import datetime
from functools import partial

import attr


# Two examples for pre-defined validators for later use
def check_ge(inst, attr, value, bound):
    if value < bound:
        raise ValueError(f'"{attr.name}" is not >= {bound}: {value}')

def check_le(inst, attr, value, bound):
    if value > bound:
        raise ValueError(f'"{attr.name}" is not <= {bound}: {value}')


def validating_class(auto_converters=True, **kwargs):
    # Extension wrapper for "attrs()"
    def wrap(cls):
        # if auto_converters:
        #   Here, I'd have to manually iterate/evaluate all attribs.
        #   This duplicates work in "attr.s()".
        cls = attr.s(auto_attribs=True, frozen=True, slots=True, **kwargs)(cls)
[...]

Issue #643, opened 2020-05-01
Order of attributes in a attr.s class is dependent upon the order the attributes were created


hansonap:
This is related to #300, #339. amd #343, but this issue doesn't just occur with attr.make_class but also normal classes defined across modules. I am submitting this issue not necessarily to prompt change to how attributes are ordered, but to document the situation, discuss documentation change, and help others you have had this issue.

Take for example, module_a defines a collection of available attributes that can be queried across an interface. module_a also defines ClassA that contains the minimum required attributes for working across the interface. Three types of attributes exist: required, optional, and deferred, but deferred attributes are not defined in module_a

model_b (and dozens of other modules) define classes, such as ClassB that inherit from ClassA, and also include a varying number of attributes (depending on the use case). Some of these attributes are deferred attributes (default=None and metadata{'attr_type': AttributeType.Deferred}). Deferred attributes are ignored when retrieving data across the interface. When retrieving data across the interface, a function takes ClassB and builds a list of required attributes, and a dictionary of optional attributes (maps attribute init position to attribute name), and deferred attributes are ignored. The list and dictionary are sent across the interface, massive amount of data is fetched, and a list of ClassBs are initialized. Deferred attributes are either derived from required / optional attributes, or must be gathered through another interface in a second step and serve as a placeholder.

```python
@attr.s
class ClassA:
    index = attr.ib(type=int, metadata={'required': True, 'attr_type': AttributeType.Required})
    tree_position = attr.ib(type=str, metadata={'required': True, 'attr_type': AttributeType.Required})
    ...

SourceAttr = attr.ib(type=ProductSource,
                     converter=convert_product_source,
                     metadata={'attr_type': AttributeType.Optional})

RevisionAttr = attr.ib(type=str, metadata={'attr_type': AttributeType.Optional})
DefinitionAttr = attr.ib(type=str, metadata={'attr_type': AttributeType.Optional})
# hundreds more

@attr.s
class ClassB(ClassA):
    revision = RevisionAttr
    source = SourceAttr
    ...
    # deferred attributes
    other_info = attr.ib(type=List[view_limits.OtherInfo], default=attr.Factory(list),
                             metadata={'attr_type': AttributeType.Deferred})
```
[...]

PR #653, opened 2020-06-17
Improve (de)serialization and validation capabilities

sscherfke:
Improve (de)serialization and validation capabilities by adding hooks that can automatically update a class’ attributes and that can change the way how objects are serialized by `asdict()`.

See: #649

# Pull Request Check List

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing left to do.

### DIFF
--- a/changelog.d/653.change.rst
+++ b/changelog.d/653.change.rst
@@ -1,0 +1,3 @@
+``attr.s()`` now has a *field_transformer* hook that is called for all ``Attribute``\ s and returns a (modified or updated) list of ``Attribute`` instances.
+``attr.asdict()`` has a *value_serializer* hook that can change the way values are converted.
+Both hooks are meant to help with data (de-)serialization workflows.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -24,7 +24,7 @@ Core
 
 .. autodata:: attr.NOTHING
 
-.. autofunction:: attr.s(these=None, repr_ns=None, repr=None, cmp=None, hash=None, init=None, slots=False, frozen=False, weakref_slot=True, str=False, auto_attribs=False, kw_only=False, cache_hash=False, auto_exc=False, eq=None, order=None, auto_detect=False, collect_by_mro=False, getstate_setstate=None, on_setattr=None)
+.. autofunction:: attr.s(these=None, repr_ns=None, repr=None, cmp=None, hash=None, init=None, slots=False, frozen=False, weakref_slot=True, str=False, auto_attribs=False, kw_only=False, cache_hash=False, auto_exc=False, eq=None, order=None, auto_detect=False, collect_by_mro=False, getstate_setstate=None, on_setattr=None, field_transformer=None)
 
    .. note::
 
@@ -93,11 +93,7 @@ Core
       ValueError: x must be positive
 
 .. autoclass:: attr.Attribute
-
-   Instances of this class are frequently used for introspection purposes like:
-
-   - `fields` returns a tuple of them.
-   - Validators get them passed as the first argument.
+   :members: assoc
 
    .. warning::
 
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -159,3 +159,116 @@ Here are some tips for effective use of metadata:
     ...     x = typed(int, default=1, init=False)
     >>> attr.fields(C).x.metadata[MY_TYPE_METADATA]
     <class 'int'>
+
+
+.. _transform-fields:
+
+Automatic Field Transformation and Modification
+-----------------------------------------------
+
+Attrs allows you to automatically modify or transform the class' fields while the class is being created.
+You do this by passing a *field_transformer* hook to `attr.define` (and its friends).
+Its main purpose is to automatically add converters to attributes based on their type to aid the development of API clients and other typed data loaders.
+
+This hook must have the following signature:
+
+.. function:: your_hook(cls: type, fields: List[attr.Attribute]) -> List[attr.Attribute]
+   :noindex:
+
+- *cls* is your class right *before* it is being converted into an attrs class.
+  This means it does not yet have the ``__attrs_attrs__`` attribute.
+
+- *fields* is a list of all :class:`attr.Attribute` instances that will later be set to ``__attrs_attrs__``.
+  You can modify these attributes any way you want:
+  You can add converters, change types, and even remove attributes completely or create new ones!
+
+For example, let's assume that you really don't like floats:
+
+.. doctest::
+
+   >>> def drop_floats(cls, fields):
+   ...     return [f for f in fields if f.type not in {float, 'float'}]
+   ...
+   >>> @attr.frozen(field_transformer=drop_floats)
+   ... class Data:
+   ...     a: int
+   ...     b: float
+   ...     c: str
+   ...
+   >>> Data(42, "spam")
+   Data(a=42, c='spam')
+
+A more realistic example would be to automatically convert data that you, e.g., load from JSON:
+
+.. doctest::
+
+   >>> from datetime import datetime
+   >>>
+   >>> def auto_convert(cls, fields):
+   ...     results = []
+   ...     for field in fields:
+   ...         if field.converter is not None:
+   ...             results.append(field)
+   ...             continue
+   ...         if field.type in {datetime, 'datetime'}:
+   ...             converter = (lambda d: datetime.fromisoformat(d) if isinstance(d, str) else d)
+   ...         else:
+   ...             converter = None
+   ...         results.append(field.assoc(converter=converter))
+   ...     return results
+   ...
+   >>> @attr.frozen(field_transformer=auto_convert)
+   ... class Data:
+   ...     a: int
+   ...     b: str
+   ...     c: datetime
+   ...
+   >>> from_json = {"a": 3, "b": "spam", "c": "2020-05-04T13:37:00"}
+   >>> Data(**from_json)  # ****
+   Data(a=3, b='spam', c=datetime.datetime(2020, 5, 4, 13, 37))
+
+
+Customize Value Serialization in ``asdict()``
+---------------------------------------------
+
+``attrs`` allows you to serialize instances of ``attrs`` classes to dicts using the `attr.asdict` function.
+However, the result can not always be serialized since most data types will remain as they are:
+
+.. doctest::
+
+   >>> import json
+   >>> import datetime
+   >>>
+   >>> @attr.frozen
+   ... class Data:
+   ...    dt: datetime.datetime
+   ...
+   >>> data = attr.asdict(Data(datetime.datetime(2020, 5, 4, 13, 37)))
+   >>> data
+   {'dt': datetime.datetime(2020, 5, 4, 13, 37)}
+   >>> json.dumps(data)
+   Traceback (most recent call last):
+     ...
+   TypeError: Object of type datetime is not JSON serializable
+
+To help you with this, `attr.asdict` allows you to pass a *value_serializer* hook.
+It has the signature
+
+.. function:: your_hook(inst: type, field: attr.Attribute, value: typing.Any) -> typing.Any
+   :noindex:
+
+.. doctest::
+
+   >>> def serialize(inst, field, value):
+   ...     if isinstance(value, datetime.datetime):
+   ...         return value.isoformat()
+   ...     return value
+   ...
+   >>> data = attr.asdict(
+   ...     Data(datetime.datetime(2020, 5, 4, 13, 37)),
+   ...     value_serializer=serialize,
+   ... )
+   >>> data
+   {'dt': '2020-05-04T13:37:00'}
+   >>> json.dumps(data)
+   '{"dt": "2020-05-04T13:37:00"}'
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -13,6 +13,7 @@ def asdict(
     filter=None,
     dict_factory=dict,
     retain_collection_types=False,
+    value_serializer=None,
 ):
     """
     Return the ``attrs`` attribute values of *inst* as a dict.
@@ -32,6 +33,10 @@ def asdict(
     :param bool retain_collection_types: Do not convert to ``list`` when
         encountering an attribute whose type is ``tuple`` or ``set``.  Only
         meaningful if ``recurse`` is ``True``.
+    :param Optional[callable] value_serializer: A hook that is called for every
+        attribute or dict key/value.  It receives the current instance, field
+        and value and must return the (updated) value.  The hook is run *after*
+        the optional *filter* has been applied.
 
     :rtype: return type of *dict_factory*
 
@@ -40,6 +45,7 @@ def asdict(
 
     ..  versionadded:: 16.0.0 *dict_factory*
     ..  versionadded:: 16.1.0 *retain_collection_types*
+    ..  versionadded:: 20.3.0 *value_serializer*
     """
     attrs = fields(inst.__class__)
     rv = dict_factory()
@@ -47,17 +53,28 @@ def asdict(
         v = getattr(inst, a.name)
         if filter is not None and not filter(a, v):
             continue
+        if value_serializer is not None:
+            v = value_serializer(inst, a, v)
         if recurse is True:
             if has(v.__class__):
                 rv[a.name] = asdict(
-                    v, True, filter, dict_factory, retain_collection_types
+                    v,
+                    True,
+                    filter,
+                    dict_factory,
+                    retain_collection_types,
+                    value_serializer,
                 )
             elif isinstance(v, (tuple, list, set)):
                 cf = v.__class__ if retain_collection_types is True else list
                 rv[a.name] = cf(
                     [
                         _asdict_anything(
-                            i, filter, dict_factory, retain_collection_types
+                            i,
+                            filter,
+                            dict_factory,
+                            retain_collection_types,
+                            value_serializer,
                         )
                         for i in v
                     ]
@@ -67,10 +84,18 @@ def asdict(
                 rv[a.name] = df(
                     (
                         _asdict_anything(
-                            kk, filter, df, retain_collection_types
+                            kk,
+                            filter,
+                            df,
+                            retain_collection_types,
+                            value_serializer,
                         ),
                         _asdict_anything(
-                            vv, filter, df, retain_collection_types
+                            vv,
+                            filter,
+                            df,
+                            retain_collection_types,
+                            value_serializer,
                         ),
                     )
                     for kk, vv in iteritems(v)
@@ -82,19 +107,36 @@ def asdict(
     return rv
 
 
-def _asdict_anything(val, filter, dict_factory, retain_collection_types):
+def _asdict_anything(
+    val,
+    filter,
+    dict_factory,
+    retain_collection_types,
+    value_serializer,
+):
     """
     ``asdict`` only works on attrs instances, this works on anything.
     """
     if getattr(val.__class__, "__attrs_attrs__", None) is not None:
         # Attrs class.
-        rv = asdict(val, True, filter, dict_factory, retain_collection_types)
+        rv = asdict(
+            val,
+            True,
+            filter,
+            dict_factory,
+            retain_collection_types,
+            value_serializer,
+        )
     elif isinstance(val, (tuple, list, set)):
         cf = val.__class__ if retain_collection_types is True else list
         rv = cf(
             [
                 _asdict_anything(
-                    i, filter, dict_factory, retain_collection_types
+                    i,
+                    filter,
+                    dict_factory,
+                    retain_collection_types,
+                    value_serializer,
                 )
                 for i in val
             ]
@@ -103,13 +145,19 @@ def _asdict_anything(val, filter, dict_factory, retain_collection_types):
         df = dict_factory
         rv = df(
             (
-                _asdict_anything(kk, filter, df, retain_collection_types),
-                _asdict_anything(vv, filter, df, retain_collection_types),
+                _asdict_anything(
+                    kk, filter, df, retain_collection_types, value_serializer
+                ),
+                _asdict_anything(
+                    vv, filter, df, retain_collection_types, value_serializer
+                ),
             )
             for kk, vv in iteritems(val)
         )
     else:
         rv = val
+        if value_serializer is not None:
+            rv = value_serializer(None, None, rv)
     return rv
 
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -373,7 +373,7 @@ def _collect_base_attrs(cls, taken_attr_names):
             if a.inherited or a.name in taken_attr_names:
                 continue
 
-            a = a._assoc(inherited=True)
+            a = a.assoc(inherited=True)
             base_attrs.append(a)
             base_attr_map[a.name] = base_cls
 
@@ -411,7 +411,7 @@ def _collect_base_attrs_broken(cls, taken_attr_names):
             if a.name in taken_attr_names:
                 continue
 
-            a = a._assoc(inherited=True)
+            a = a.assoc(inherited=True)
             taken_attr_names.add(a.name)
             base_attrs.append(a)
             base_attr_map[a.name] = base_cls
@@ -419,7 +419,9 @@ def _collect_base_attrs_broken(cls, taken_attr_names):
     return base_attrs, base_attr_map
 
 
-def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
+def _transform_attrs(
+    cls, these, auto_attribs, kw_only, collect_by_mro, field_transformer
+):
     """
     Transform all `_CountingAttr`s on a class into `Attribute`s.
 
@@ -451,6 +453,7 @@ def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
                 continue
             annot_names.add(attr_name)
             a = cd.get(attr_name, NOTHING)
+
             if not isinstance(a, _CountingAttr):
                 if a is NOTHING:
                     a = attrib()
@@ -498,8 +501,8 @@ def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
     AttrsClass = _make_attr_tuple_class(cls.__name__, attr_names)
 
     if kw_only:
-        own_attrs = [a._assoc(kw_only=True) for a in own_attrs]
-        base_attrs = [a._assoc(kw_only=True) for a in base_attrs]
+        own_attrs = [a.assoc(kw_only=True) for a in own_attrs]
+        base_attrs = [a.assoc(kw_only=True) for a in base_attrs]
 
     attrs = AttrsClass(base_attrs + own_attrs)
 
@@ -518,6 +521,8 @@ def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
         if had_default is False and a.default is not NOTHING:
             had_default = True
 
+    if field_transformer is not None:
+        attrs = field_transformer(cls, attrs)
     return _Attributes((attrs, base_attrs, base_attr_map))
 
 
@@ -574,9 +579,15 @@ class _ClassBuilder(object):
         collect_by_mro,
         on_setattr,
         has_custom_setattr,
+        field_transformer,
     ):
         attrs, base_attrs, base_map = _transform_attrs(
-            cls, these, auto_attribs, kw_only, collect_by_mro
+            cls,
+            these,
+            auto_attribs,
+            kw_only,
+            collect_by_mro,
+            field_transformer,
         )
 
         self._cls = cls
@@ -1001,6 +1012,7 @@ def attrs(
     collect_by_mro=False,
     getstate_setstate=None,
     on_setattr=None,
+    field_transformer=None,
 ):
     r"""
     A class decorator that adds `dunder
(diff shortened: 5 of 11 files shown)


tests/test_make.py
```diff
@@ -2131,3 +2138,58 @@ def __setstate__(self, state):
...
+        assert results == [
+            ('x', int, True),
+            ('y', str, True),
+        ]
+
+    @pytest.mark.skipif(PY2, reason="Auto-detection is Python 3-only.")
+    def test_hook_applied_modify_attrib(self):
+        """The transform hook must be able to modify attributes."""
+        def hook(name, ca, type):
+            ca.converter = type
```
wsanchez, line 2187: Hrm… if I were writing a hook like this, I'd be type hinting everything, and the type of `ca` here is `_CountingAttr`, which is private.  That seems like a problem.

IIRC we've avoided making `_CountingAttr` public, so this API would require us to at least define an interface, no?

sscherfke: Yes, that’s true. I also stumbled upon this when I was writing the docstring. Making an Interface seems like a reasonable idea. We wouldn’t to expose the actual `_CountingAttr` than but mypy would be happy nonetheless.

sscherfke:
I have a lot more uncommited code, but some (important) parts of it require at least Python 3.6.

When are you planning to drop support for Python 2 (and maybe even 3.5)?  This would make a few things easier and if that’s not too far into the future, I would gladly wait a little. :)

wsanchez:
> When are you planning to drop support for Python 2 (and maybe even 3.5)? This would make a few things easier and if that’s not too far into the future, I would gladly wait a little. :)

I suggest you file an Issue called "[RFC] Drop support for Python 2" and have that discussion there, as it's not specific to this ticket. (And perhaps another for 3.5.)

hynek:
I'm sorry I'm offline for the next week, but there's been a feature request asking for `_CountingAttr` to become public already so this might be two birds with one stone.

As for Python 2: https://www.attrs.org/en/stable/python-2.html

As much as I don't like it, it's one of our main features over dataclasses ATM.

_Once_ we go Python 3-only, it will definitely not be 3.5. :)

sscherfke:
What is your preferred approach for tests that run (and parse) only with python 3.6 and above?

hynek:
https://github.com/python-attrs/attrs/blob/3e3d84216c5c7b6b451610ebbe8f5a7ed9c3209b/conftest.py#L17

sscherfke:
This PR will only focus on the hooks. I have removed the converters and validators and will re-propose them in another PR when this one is accepted.

sscherfke:
`_CountingAttr` (or an interface for it) need to be made public for this PR.

What would be your prefered approach?  Create a new Interface that `_CountingAttr` inherits or rename `_CountingAttr` to `CountingAttr`? Should also be imported in `__init__.py` in that case?

wsanchez:
> `_CountingAttr` (or an interface for it) need to be made public for this PR.
> 
> What would be your prefered approach? Create a new Interface that `_CountingAttr` inherits or rename `_CountingAttr` to `CountingAttr`? Should also be imported in `__init__.py` in that case?

Again, we've resisted making `_CountingAttr` in the past because it's not an API we want to commit to long-term, so an interface that is limited to what we *are* willing to support seems best, leaving `_CountingAttr` itself as an implementation detail.

That assumes that @hynek is amenable to the resulting public interface.

hynek:
The main issue here is that I don't want it to be "counting" anymore if not necessary (ordered dicts), cf #643. I guess we'll have to extract a sub-interface?

sscherfke:
An Interface would suffice since it will (at least in this PR) only be used for documentation and (mypy) typing. I can try to extract an Interface.

sscherfke:
Hm, `_CountingAttr` is so "simple" and specific, that I don’t know which attributes/meths to extract into an interface.

wsanchez:
> Hm, `_CountingAttr` is so "simple" and specific, that I don’t know which attributes/meths to extract into an interface.

I'd start with: what exactly is the minimum needed for this PR to function?

sscherfke:
*I* only want to add a converter to the attrib, but instances of `_CountingAttr` (I’d call the interface `MutableAttr` since it is, in contrast to `Attribute`, mutable) are passed to the hook so it would not be very long-sighted to only add the converter attribute to the interface.  I think at least `converter`, `metadata`, `validator()` and `default()` should be made public.  Changing the other attributes in the hook might not make too much sense.

hynek:
I guess a fairer name would be `AttributeDefinition` but I wonder if we couldn't spin it in a way that the hook receives `Attributes` but is free to replace them with something else? The would be definitely cleaner and less invasive.

sscherfke:
I’m not sure if I understand what you mean.  The whole purpose of the hook is to be invasive and let the user mess with the attribute definition.

hynek:
I mean that due to stuff like resolving decorator-based defaults and validators we can’t really provide a good interface? My hope would be that the hook would get a list of `Attribute`s will fully resolved validators etc and would return a list of `Attributes` that may replace some of them or even remove some?

sscherfke:
Thanks for the clarification. I will think about this and see if it works.

hynek:
Yeah, please let me know what you think. The upsides are mainly that we don’t have to introduce a new “in-between type” that would have to be mutated by the users. That smells like bugs and edge cases to me. I’m open to add more helper APIs to `Attribute`, though!

sscherfke:
This was an excelent idea of you, @hynek.  :) The hook is now purely functional. It takes an `Attribute` and returns an `Attribute`.  The only thing that we’d need to change now is to make `Attribute._assoc` public in order to allow people to evolve an attribute.

sscherfke:
ping :)

sscherfke:
ping :)

hynek:
Yeah I'm sorry, I really had to focus on landing all the `__setattr__` stuff (#663 still remains) because I was procrastinating on this stuff for months and it was the last hurdle before starting to work on "import attrs", because it affects default values. Once that lands, I urgently need to get 20.1 out of the door this, the `__eq__` PR, and "import attrs" will be my next priorities for 20.2 which I hope to get out of the door _much_ faster.

I was basically running behind a carrot all year and had to draw a line now. Sorry it affected you. :(((

sscherfke:
No worries.  I know this problem.  Just wanted to make sure I’m not forgotten. :)

hynek:
No worries, I **really** want to open attrs to your use cases!

hynek:
(you may want to start rebasing tho, sadly the innards have quite changed :()

sscherfke:
I think the PR is now complete.

When this gets merge, I will create another one with some new validators, a default transform_hook, and a new section in the *Examples* documentation.

hynek:
At a first glance this look reasonable!

Just to double check: have you been able to implement your PoCs with this kind of code? Because as it stands here, there is no real justification for it so I just wanna make sure you’re 100% positive this is what you need and it’s sufficient. :)

sscherfke:
Yes, I already tested the hooks with an API client (JSON (de)serialization to/from attrs classes) and the same hook should work for my typed-settings PoC as well.

Instead of creating a second PR, I can add its contents directly to this one. This might make the benefits of this PR more clear.

sscherfke:
I’ve added a draft for additional validators and (auto) converters. This would allow us to use attrs directly for

- API clients (load attrs classes from json and serialize request data to json)
- Typed settings loader

Here is an example:

```python
from datetime import datetime
from enum import Enum
from typing import List, Set
import json

from attr.converters import auto_convert_hook
from attr.validators import le, maxlen
import attr


class LeEnum(Enum):
    spam = 'Le spam'
    eggs = 'Le eggs'


@attr.frozen(transform_hook=auto_convert_hook)
class Child:
    x: int = attr.ib()
    y: int = attr.ib(converter=float)


@attr.frozen(kw_only=True, transform_hook=auto_convert_hook)
class Parent:
    child: Child
    a: float
    b: float = attr.field(default=3.14, validator=le(2))
    c: LeEnum
    d: datetime
    e: 'List[Child]'
    f: Set[datetime]


DATA = {
    'a': '1',
    'b': '2',
    'c': 'Le spam',
    'd': '2020-05-04T13:37:00',
    'e': [
        {
            'x': '23',
            'y': '42',
        },
    ],
    'f': ['2020-05-04T13:37:00', '2020-05-04T13:37:00'],
    'child': {
        'x': '23',
        'y': '42',
    },
}
p = Parent(**DATA)
print(p)
d = attr.asdict(p, hook=attr.converters.serialize_hook)
print(d)

jd = json.dumps(d)
print(Parent(**json.loads(json.dumps(d))) == p)
```

The validators and converters should cover the most common use cases. I’d like to make the `auto_convert_hook()` more customizable, though, so that users could add additional types without having to create a completely new hook.

sscherfke:
There is no a `make_auto_converter()` factory that allows you to extend the bundled `auto_convert()` hook. 
I also added a few tests for this.

What do you think of the general direction this PR is taking? Shall I continue with adding tests and docs?

si14:
A note from a silent observer: this PR looks like a fabulous addition to `attrs` and I'm really looking forward to using it. Thanks for your great work, @sscherfke and @hynek !

src/attr/_make.py
```diff
@@ -484,6 +487,9 @@ def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
         for attr_name, ca in ca_list
     ]
 
+    if transform_hook is not None:
+        own_attrs = [transform_hook(cls, a) for a in own_attrs]
```
hynek, line 491: Out of curiosity: why did you choose to call them one by one and not all at once?

I have a gut feeling that an `own_atts = transform_hooks(cls, own_attrs)` wouldn't be just faster (less method calls) but also allow for more use-cases where you need see all attributes? 🤔

changelog.d/653.change.rst
```diff
@@ -0,0 +1,5 @@
+``attr.s()`` now has a *transform_hook* that is called for each
```
hynek, line 1: We use [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/). :)

docs/api.rst
```diff
@@ -24,7 +24,7 @@ Core
 
 .. autodata:: attr.NOTHING
 
-.. autofunction:: attr.s(these=None, repr_ns=None, repr=None, cmp=None, hash=None, init=None, slots=False, frozen=False, weakref_slot=True, str=False, auto_attribs=False, kw_only=False, cache_hash=False, auto_exc=False, eq=None, order=None, auto_detect=False, collect_by_mro=False, getstate_setstate=None, on_setattr=None)
+.. autofunction:: attr.s(these=None, repr_ns=None, repr=None, cmp=None, hash=None, init=None, slots=False, frozen=False, weakref_slot=True, str=False, auto_attribs=False, kw_only=False, cache_hash=False, auto_exc=False, eq=None, order=None, auto_detect=False, collect_by_mro=False, getstate_setstate=None, on_setattr=None, transform_hook=None)
```
hynek, line 27: Do you think we could come up with a better name for the hook? While through the lens of this PR it makes complete sense, I feel its a bit generic next to the other options. It also bleeds out the name of the function that calls it which doesn't make much sense to the users. Would you agree?

src/attr/_funcs.py
```diff
@@ -13,6 +13,7 @@ def asdict(
     filter=None,
     dict_factory=dict,
     retain_collection_types=False,
+    hook=None,
```
hynek, line 16: Same here, I'm afraid we need long arg names. :)

src/attr/converters.py
```diff
@@ -83,3 +86,117 @@ def default_if_none_converter(val):
             return default
 
     return default_if_none_converter
+
+
+def convert_attrs(cls):
```
hynek, line 91: This is not our style of writing docstrings. 😇

(Check out https://github.com/python-attrs/attrs/blob/master/.github/CONTRIBUTING.rst#code – I know your style is the CPython one but I can't stand it because the rationale per PEP 257 is that Guido's Emacs can't format paragraphs otherwise 🤪).

hynek (changes requested):
So overall I'm quite happy how it's turning out!

The following needs to be done:

- I think we need a better name for `transform_hook`, I explain it better in the comments.
- Please remove the converters/validators to keep this PR small. Your actual feature is literally a two line change but due to its impact on the core of attrs it ripples throughout the whole code base. 🙈 We'll do it in a separate PR. At this point I'd like to get it in and build on it.
- You can start adding basic docs, yes! Maybe a short chapter on its own? We'll have to edit/rewrite it probably but we gotta start somewhere and you're the best person for it since the whole thing has been started by your own motivations. :)
- I need to get our a bug release ASAP, but this PR is on the top of my big-things-for-attrs-queue.

I'm very excited by all of this!

src/attr/_make.py
sscherfke: This is a good point. When I started work on this feature, I called the hook in another place where I *needed* to call it once for each attrib. This is now no longer necessary and my default auto-convert-hook would actually benefit from making this one-call-for-all. :+1:

src/attr/converters.py
sscherfke: Alright, I will fix this. :)

docs/api.rst
sscherfke: Naming things is so hard! 🤪 Is something like `attrib_converter()` better?

src/attr/_funcs.py
sscherfke: `value_serializer()`?

sscherfke:
Ok, I'll rip out the validators and converters again and put them in a separate PR.

Would a new section in the `examples.rst` be a gould place for the docs or shall I create a new document?

src/attr/_make.py
hynek: What was the use-case for needing to call them once? Can't you simulate it using an intermediate function?

docs/api.rst
hynek: So this is awkward.

The reason why new-style APIs use `field` is because it's quite distinct to the soup of attrs, attribs, attr and whatnot.

Given we're marching towards a brighter future…maybe `field_transformer`? 😇

Summoning the creative juices of @wsanchez @glyph @euresti @Tinche etc 😅

src/attr/_funcs.py
hynek: Yeah, better I think.

hynek:
I think this is an advanced feature so either `extending.rst` or something new entirely is apt?

docs/api.rst
ambv: +1 to `field_transformer`

src/attr/_make.py
sscherfke: There never was a use case. It was just the natural way of doing it were I originally implemented it. I think there is no drawback in making it `hook(cls: type, fields: List[Attribute]) -> List[Attribute])`

hynek: Wonderful, pls do. Let’s provide the users with as much context as necessary.

sscherfke:
I renamed the hooks, (hopefully) fixed the docs format, added more documentation, removed the validators and converters, rebased on master, and updated *versionadded* labels to 20.3.0.

sscherfke:
Did some more polishing and fixed some formatting issues.

src/attr/_make.py
```diff
@@ -484,6 +487,9 @@ def _transform_attrs(cls, these, auto_attribs, kw_only, collect_by_mro):
         for attr_name, ca in ca_list
     ]
 
+    if field_transformer is not None:
```
hynek, line 490: One quick (but important) thing: why are you calling this so early? I would put it at the very end. There's no reason to ignore inherited attributes, is there?

BTW another reason why calling it on all attrs is much better: the transformer has to power to add or remove attributes if it wants to! 🎉

sscherfke: Do you mean like this`
```python
    if field_transformer is not None:
        attrs = field_transformer(cls, attrs)
    return _Attributes((attrs, base_attrs, base_attr_map))
```

sscherfke: > BTW another reason why calling it on all attrs is much better: the transformer has to power to add or remove attributes if it wants to! 🎉

The docs contain an example where I use this hook to remove all float fields from a class – because, who likes floats? 🤪

sscherfke: Seems to work. I also added a test for this.

sscherfke:
ping :cat2:

hynek:
> ping 🐈

Current status:

current status

sscherfke:
Oh, I thought you’re already back. Then have a lot of fun and I’ll be patient! 🏄

docs/extending.rst
```diff
@@ -159,3 +159,107 @@ Here are some tips for effective use of metadata:
...
+You do this by passing a *field_transformer* hook to :func:`attr.define()` (and its friends).
+Its main purpose is to automatically add converters to attributes based on their type to aid the development of API clients and other typed data loaders.
+
+| This hook must have the signature:
+| :code:`your_hook(cls: type, fields: List[Attribute]) -> List[Attribute]`
+
+- *cls* is your class right *before* it is being converted into an attrs class.
+  This means it does not yet have the ``__attrs_attrs__`` attribute.
+
+- *fields* is a list of all :class:`attr.Attribute` instances that will later be set to ``__attrs_attrs__``.  You can modify these attributes any way you want:
```
hynek, line 177:
```suggestion
- *fields* is a list of all :class:`attr.Attribute` instances that will later be set to ``__attrs_attrs__``. 
  You can modify these attributes any way you want:
```

docs/extending.rst
```diff
@@ -159,3 +159,107 @@ Here are some tips for effective use of metadata:
...
+   ...
+   >>> from_json = {"a": 3, "b": "spam", "c": "2020-05-04T13:37:00"}
+   >>> Data(**from_json)  # ****
+   Data(a=3, b='spam', c=datetime.datetime(2020, 5, 4, 13, 37))
+
+
+Change value serialization in ``asdict()``
+------------------------------------------
+
+Attrs allows you to serialize instances of attrs classes to dicts using the :func:`attr.asdict()` function.
```
hynek, line 229:
```suggestion
``attrs`` allows you to serialize instances of ``attrs`` classes to dicts using the `attr.asdict` function.
```

docs/extending.rst
```diff
@@ -159,3 +159,107 @@ Here are some tips for effective use of metadata:
...
+   ...
+   >>> data = attr.asdict(Data(datetime.datetime(2020, 5, 4, 13, 37)))
+   >>> data
+   {'dt': datetime.datetime(2020, 5, 4, 13, 37)}
+   >>> json.dumps(data)
+   Traceback (most recent call last):
+     ...
+   TypeError: Object of type datetime is not JSON serializable
+
+To help you with this, :attr:`attr.asdict()` allows you to pass a *value_serializer* hook.  It has the signature :code:`your_hook(inst: type, field: Attribute, value: Any) -> Any`:
```
hynek, line 249:
```suggestion
To help you with this, `attr.asdict` allows you to pass a *value_serializer* hook.
It has the signature

.. function:: your_hook(inst: type, field: attr.Attribute, value: typing.Any) -> typing.Any
   :noindex:
```

docs/extending.rst
```diff
@@ -159,3 +159,107 @@ Here are some tips for effective use of metadata:
...
+   ...     b: str
+   ...     c: datetime
+   ...
+   >>> from_json = {"a": 3, "b": "spam", "c": "2020-05-04T13:37:00"}
+   >>> Data(**from_json)  # ****
+   Data(a=3, b='spam', c=datetime.datetime(2020, 5, 4, 13, 37))
+
+
+Change value serialization in ``asdict()``
+------------------------------------------
```
hynek, line 227: I think Customize is better here:

```suggestion
Customize Value Serialization in ``asdict()``
---------------------------------------------
```

src/attr/_funcs.py
```diff
@@ -32,6 +33,9 @@ def asdict(
     :param bool retain_collection_types: Do not convert to ``list`` when
         encountering an attribute whose type is ``tuple`` or ``set``.  Only
         meaningful if ``recurse`` is ``True``.
+    :param callable value_serializer: A hook that is called for every attribute
+        or dict key/value.  It receives the current instance, field and value
+        and must return the (updated) value.
```
hynek, line 38:
```suggestion
    :param Optional[callable] value_serializer: A hook that is called for every
        attribute or dict key/value.  It receives the current instance, field and
        value and must return the (updated) value.
```

(I'm eyeballing the line length here, I hope it works out 🤪)

src/attr/_funcs.py
```diff
@@ -40,24 +44,36 @@ def asdict(
...
     ..  versionadded:: 16.1.0 *retain_collection_types*
+    ..  versionadded:: 20.3.0 *value_serializer*
     """
     attrs = fields(inst.__class__)
     rv = dict_factory()
     for a in attrs:
         v = getattr(inst, a.name)
+        if value_serializer is not None:
+            v = value_serializer(inst, a, v)
         if filter is not None and not filter(a, v):
```
hynek, line 55: Wouldn't it be better, if we ran the filters first to avoid unnecessary serialization? The order should be documented in any case.

src/attr/_make.py
```diff
@@ -1196,6 +1208,13 @@ def attrs(
         If a list of callables is passed, they're automatically wrapped in an
         `attr.setters.pipe`.
 
+    :param field_transformer:
```
hynek, line 1211:
```suggestion
    :param Optional[callable] field_transformer:
```

src/attr/_make.py
```diff
@@ -1196,6 +1208,13 @@ def attrs(
...
+        A function that is called with the original class object and all
+        fields: ``ft(cls: type, attribs: List[Attribute]) -> List[Attribute]``.
+        The hook is called right before the class gets transformed into an
+        attrs class.  The fields are already :class:`Attribute` instances.  You
+        can use this, e.g., to automatically add converters or validators to
+        fields.
```
hynek, line 1217: I'm gonna say we should shorten this to "what does it do and why it's interesting" and then link to extending:

```suggestion
        A function that is called with the original class object and all
        fields right before ``attrs`` finalizes the class.  You can use
        this, e.g., to automatically add converters or validators to
        fields based on their types.  See `transform-fields` for more details.
```

docs/extending.rst
```diff
@@ -159,3 +159,107 @@ Here are some tips for effective use of metadata:
     ...     x = typed(int, default=1, init=False)
     >>> attr.fields(C).x.metadata[MY_TYPE_METADATA]
     <class 'int'>
+
```
hynek, line 162: This is for a reference from the `attr.s` docstring below.

```suggestion


.. _transform-fields:
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+import attr
+
+
+class TestTransformHook:
+    """Tests for `attrs(tranform_value_serializer=func)`"""
+
+    def test_hook_applied(self):
+        """
+        The transform hook must be applied to all attributes.  Types will not
+        be available here (they only work with type annotations).
```
hynek, line 13: Wait, really? Why?

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+
+        @attr.s(auto_attribs=True, field_transformer=hook)
+        class C:
+            x: int
+            y: str = attr.ib()
+
+        assert results == [("x", int), ("y", str)]
+
+    def test_hook_applied_modify_attrib(self):
+        """The transform hook must be able to modify attributes."""
```
hynek, line 47:
```suggestion
        """
        The transform hook can modify attributes.
        """
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
+from datetime import datetime
+from typing import Dict, List
+
+import attr
+
+
+class TestTransformHook:
+    """Tests for `attrs(tranform_value_serializer=func)`"""
```
hynek, line 8:
```suggestion
    """
    Tests for ``attr.s(transform_value_serializer=func)``
    """
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+            x: int
+            y: str = attr.ib()
+
+        assert results == [("x", int), ("y", str)]
+
+    def test_hook_applied_modify_attrib(self):
+        """The transform hook must be able to modify attributes."""
+
+        def hook(cls, attribs):
+            return [a._assoc(converter=a.type) for a in attribs]
```
hynek, line 50: We should probably make `_assoc` public then? 🤔 Or at least document it?

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+        @attr.s(auto_attribs=True, field_transformer=hook)
+        class C:
+            x: int = attr.ib(converter=int)
+            y: float
+
+        c = C(x="3", y="3.14")
+        assert c == C(x=3, y=3.14)
+
+    def test_hook_remove_field(self):
+        """It must be possible to remove fields via the hook."""
```
hynek, line 61:
```suggestion
        """
        It is possible to remove fields via the hook.
        """
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+
+        @attr.s(auto_attribs=True, field_transformer=hook)
+        class C:
+            x: int
+            y: float
+
+        assert attr.asdict(C(2.7)) == {"y": 2.7}
+
+    def test_hook_add_field(self):
+        """It must be possible to remove fields via the hook."""
```
hynek, line 74:
```suggestion
        """
        It is possible to add fields via the hook.
        """
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+            return [a1, a2]
+
+        @attr.s(auto_attribs=True, field_transformer=hook)
+        class C:
+            x: int
+
+        assert attr.asdict(C(1, 2)) == {"x": 1, "new": 2}
+
+    def test_hook_with_inheritance(self):
+        """The hook must also receive all fields of base classes."""
```
hynek, line 88:
```suggestion
        """
        The hook receives all fields from base classes.
        """
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+        class Sub(Base):
+            y: int
+
+        assert attr.asdict(Sub(2)) == {"y": 2}
+
+
+class TestAsDictHook:
+    def test_asdict(self):
+        """
+        asdict() must call the hooks in attrs classes and in other
```
hynek, line 109:
```suggestion
        asdict() calls the hooks in attrs classes and in other
```

tests/test_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+        assert result == {
+            "a": {"x": 1, "y": ["2020-07-01T00:00:00"]},
+            "b": [{"x": 2, "y": ["2020-07-02T00:00:00"]}],
+            "c": {"spam": {"x": 3, "y": ["2020-07-03T00:00:00"]}},
+            "d": {"eggs": "2020-07-04T00:00:00"},
+        }
+
+    def test_asdict_calls(self):
+        """
+        The correct instances and attribute names must be passed to the hook.
```
hynek, line 147:
```suggestion
        The correct instances and attribute names are passed to the hook.
```

hynek (changes requested):
A lot of nitpicking but thanks to the magic of suggestions, it should be quick to get thru!

_collapses on the floor_

docs/extending.rst
```diff
@@ -159,3 +159,107 @@ Here are some tips for effective use of metadata:
...
+Automatic field transformation and modification
+------------------------------------------------
+
+Attrs allows you to automatically modify or transform the class' fields while the class is being created.
+You do this by passing a *field_transformer* hook to :func:`attr.define()` (and its friends).
+Its main purpose is to automatically add converters to attributes based on their type to aid the development of API clients and other typed data loaders.
+
+| This hook must have the signature:
+| :code:`your_hook(cls: type, fields: List[Attribute]) -> List[Attribute]`
```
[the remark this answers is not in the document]
sscherfke: The pipes just enforce manual line breaks. Your solution seems a lot better. :)

src/attr/_funcs.py
sscherfke: Narrator: It didn’t.

tests/test_hooks.py
sscherfke: This might no longer be true. I think that was the case with my original implementation. I will check this agian.

tests/test_hooks.py
sscherfke: Why not both? 😼

src/attr/_funcs.py
sscherfke: Good point will change and document this.

tests/test_hooks.py
sscherfke: `_assoc()` is currently being used in these places:
```
src/attr/hooks.py
50:                attrib = attrib._assoc(converter=converter)

src/attr/_make.py
376:            a = a._assoc(inherited=True)
414:            a = a._assoc(inherited=True)
504:        own_attrs = [a._assoc(kw_only=True) for a in own_attrs]
505:        base_attrs = [a._assoc(kw_only=True) for a in base_attrs]
2326:    def _assoc(self, **changes):
```
I don’t think making it public would cause any harm, but you are the expert here. 😸

sscherfke: I made it public and updated the docs. I also merged the additional docs from `api.rst` into the class’s docstring since the rendered docs looked and read really strange.

sscherfke:
Ohai, this is the weekly ping. 🙊

hynek (approved):
We should change `_assoc` to `evolve`, not `assoc` but I’ll do it myself, to not delay this any further. Thanks and thank you for the patience, this is a big one!

sscherfke:
Thanks for merging! ❤️ 

Next PR will arrive shortly 😬

PR #660, opened 2020-07-10
Add on_setattr hooks to attr.s and attr.ib

### LINKED ISSUE #645: [RFC] __setattr__ hooks
There's two things that people keep asking for:

1. validation on setting attributes
2. freezing single attributes

Those two features have something in common: they require `attrs` to write a `__setattr__` method.

I actually had 1 done when I implemented validators but I took it out again, because I didn't want to tamper with `__setattr__` too. But it totally makes sense to expect that validators run there too.

Now that argument has gone away thanks to frozen classes and `attrs` is in the `__setattr__` business. So it feels like the right thing to do, to implement it and make it default for _Operation `import attrs`_ (I hope this is legit the last part of the puzzle).

To allow for 2 too, I would suggest to add a hook called `on_setattr` (better names welcome) that takes a callable that is called with the instance, the attribute definition, and the new value.

To solve 2, the implementation would look like

```python
def frozen(_, __, ___):
    raise FrozenInstanceError
```

***

Open questions:

- what to do about `on_setattr` attributes in a frozen class (incl inheritance)
- what about converters? Maybe it should take a list/`and` like validator/converter do? They would need to work as a chain, returning values for the next one.

hynek:
This fixes #645 and a billion other issues that asked for freezing attributes and validation on assignment.

I'm not married to all the names and open to better suggestions.

Yes, it was more work to integrate than anticipated. 😳

---

While working on this I've also noticed that we take no precaution about `__setattr__` being inherited, which means that if you subclass a frozen class, you can't easily thaw it. This topic is somewhat related to this PR too but I didn't want to bloat it any further. It goes down to setting `__setattr__` to `object.__setattr__` if we don't write an own, but if `auto_detect` is true, only if the user didn't write one?


src/attr/setters.pyi
```diff
@@ -0,0 +1,12 @@
...
+from typing import TypeVar, Any, NewType, cast
+
+_T = TypeVar("_T")
+
+def pipe(*setters: _OnSetAttrType) -> _OnSetAttrType: ...
+def frozen(instance: Any, attribute: Attribute, new_value: Any) -> Any: ...
+def validate(instance: Any, attribute: Attribute, new_value: _T) -> _T: ...
+def convert(instance: Any, attribute: Attribute, new_value: _T) -> _T: ...
+
+DisableType = NewType("DisableType", object)
```
euresti, line 11: You might want to warn that this isn't a runtime type.  Maybe even make it `_DisableType` to enforce that.

changelog.d/645.change.rst
```diff
@@ -0,0 +1,5 @@
+It is now possible to specify hooks that are called whenever an attribute is set **after** a class has been instantiated.
```
euresti, line 1: Is this file necessary?

tests/test_functional.py
```diff
@@ -681,3 +683,183 @@ class C(object):
...
+            y = attr.ib()
+
+        pf = PartiallyFrozen("x", "y")
+
+        pf.y = "yyy"
+
+        assert "yyy" == pf.y
+
+        with pytest.raises(FrozenAttributeError):
+            pf.x = "xxx"
```
euresti, line 731: Just FYI.  Mypy will not catch this as an error.  Not important for this diff but something to consider.

src/attr/__init__.pyi
```diff
@@ -189,6 +199,7 @@ def attrs(
     order: Optional[bool] = ...,
     auto_detect: bool = ...,
     getstate_setstate: Optional[bool] = ...,
+    on_setattr: Optional[_OnSetAttrArgType] = ...,
```
euresti, line 202: Does this actually work?
```
@attr.s(on_setattr=DISABLE)
```

src/attr/__init__.pyi
```diff
@@ -41,6 +42,10 @@ _ConverterType = Callable[[Any], _T]
 _FilterType = Callable[[Attribute[_T], _T], bool]
 _ReprType = Callable[[Any], str]
 _ReprArgType = Union[bool, _ReprType]
+_OnSetAttrType = Callable[[Any, Any, Any], Any]
```
euresti, line 45: Could this be:
```
Callable[[Any, Attribute[_T], _T], _T]
```
?
It might just be enough to do:
```
Callable[[Any, Attribute[Any], Any], Any]
```

src/attr/setters.pyi
```diff
@@ -0,0 +1,12 @@
+from . import _OnSetAttrType, Attribute
+from typing import TypeVar, Any, NewType, cast
+
+_T = TypeVar("_T")
+
+def pipe(*setters: _OnSetAttrType) -> _OnSetAttrType: ...
+def frozen(instance: Any, attribute: Attribute, new_value: Any) -> Any: ...
+def validate(instance: Any, attribute: Attribute, new_value: _T) -> _T: ...
```
euresti, line 8: These could be:
```
def validate(instance: Any, attribute: Attribute[_T], new_value: _T) -> _T: ...
def convert(instance: Any, attribute: Attribute[_T], new_value: _T) -> _T: ...
```

src/attr/setters.pyi
```diff
@@ -0,0 +1,12 @@
+from . import _OnSetAttrType, Attribute
+from typing import TypeVar, Any, NewType, cast
+
+_T = TypeVar("_T")
+
+def pipe(*setters: _OnSetAttrType) -> _OnSetAttrType: ...
+def frozen(instance: Any, attribute: Attribute, new_value: Any) -> Any: ...
```
euresti, line 7: I wonder if this could be:
```
from typing import NoReturn

def frozen(instance: Any, attribute: Attribute[_T], new_value: _T) -> NoReturn: ...
```

src/attr/__init__.pyi
hynek: It does! Mostly because I didn't want to add a sharp edge for the sake of purity:

https://github.com/python-attrs/attrs/blob/a241f0f152a99e509f572b85d4dd42f82286f1dd/src/attr/_make.py#L800

tests/test_functional.py
hynek: Oh you mean mypy should be able to catch the presence of a frozen hook? That sounds tough but would be cool indeed.

changelog.d/645.change.rst
hynek: Unfortunately that's the only way to tell towncrier to link to another issue.

src/attr/__init__.pyi
hynek: What would those `_T`s buy us?

src/attr/setters.pyi
hynek: Yeah, thanks!

src/attr/__init__.pyi
euresti: The _T means that everywhere you see it, the type must be the same.  So mypy would warn you if your method tries to change the type or if the attribute doesn't match the new value.  For example

```
def foo(instance: Any, attribute: Attribute[bool], new_value: str) -> bool:
    return new_value == 'true'

@attr.s(on_setattr=foo)
```

However looking at it, this might be something you want to do.  So I think the correct type is
```
Callable[[Any, Attribute[_T], Any], _T]
```

Basically it forces your on_setattr method to return a type that's the same as the Attribute.

hynek: Is this something we want to enforce? 🤔 If you have a list, you could have intermediate results and only the last one has to return the correct one.

src/attr/setters.pyi
hynek: NoReturn is quite fresh in typing, isn't it? I remember pulling it from mypy_extensions or something? 🤔

src/attr/__init__.pyi
euresti: Oh wow.  Hadn't thought of that.  Can converters be chained in the same way?

src/attr/setters.pyi
euresti: Since this is a stub the only thing that matters is where it is in `typeshed`.

src/attr/__init__.pyi
hynek: They will be with the next release. At this point it's called `chain()` but I think `pipe()` is better and I will rename it before I release it.

Alternatively I could be convinced to rename pipe to chain here, but I feel like pipe is better at conveying the fact, that the data flows thru it?

src/attr/setters.pyi
hynek: Oh!

src/attr/__init__.pyi
hynek: I went with `Attribute[Any]` for now

hynek:
Since there was no more feedback, I'm YOLO-timeouting this. If someone wants to chime in before the release feel free.

hynek:
P.S. @euresti I took the liberty to relax `_ConverterType` based on your comments in a74c32b…feel free to yell at me if that's terrible and should be done differently.

src/attr/setters.pyi
```diff
@@ -0,0 +1,18 @@
...
+
+_T = TypeVar("_T")
+
+def frozen(
+    instance: Any, attribute: Attribute, new_value: Any
+) -> NoReturn: ...
+def pipe(*setters: _OnSetAttrType) -> _OnSetAttrType: ...
+def validate(instance: Any, attribute: Attribute[_T], new_value: _T) -> _T: ...
+
+# convert is allowed to return Any, because they cann be chained using pipe.
```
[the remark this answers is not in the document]
hynek: Good catch, fixed in d6ba983edb43ef04d1074109301901a1b8147a9d

hhnguyen-axon:
Hi @hynek, I wonder if we have any release plan for this fix? We're having some use-cases which requires this and we're eagerly looking for it to be released. If anything we can help to make it release sooner, can you let us know?

hynek:
#666 is the last remaining part of the puzzle. I hope I’ll be able to release this week.

hhnguyen-axon:
Sounds great, thanks so much @hynek

PR #663, opened 2020-08-04
Clear __setattr__ if it was inherited && written by us

### LINKED ISSUE #661: Restore __setattr__ when subclassing and we didn't write a new __setattr__
Currently if you do this:

```python
@attr.s(frozen=True)
class A:
    a = attr.ib()

@attr.s
class B(A):
    b = attr.ib()
```

B is frozen too, because we don't write a new `__setattr__` to overwrite the old one. Once #660 is merged, this will become an even bigger problem.

We have to restore `__setattr__` to `object.__setattr__` if we don't have a reason to overwrite the existing one ourselves.

`auto_detect=True` should be considered.

hynek:
Fixes #661 

Since the frozen behavior was deliberate, the changes should only affect new `on_setattr` code.

This is the final PR before 20.1.0.

I hope we can ship 20.2.0 with `import attrs`, equality, and maybe better serialization soon after.


hynek:
Looks like codecov is broken. :(

hynek:
👯 nobody cares 👯

Issue #487, opened 2019-01-20
[RFC] Inconvenient defaults?


hynek:
As part of “Project`import attrs`” (see also #408) we get a unique opportunity to change default options/behaviors that grew over the years but couldn't be fixed due to backward-compatibility restrictions.

The rough plan is create a new, friendlier API on top of `@attr.s` and `attr.ib()` that won't go anywhere.

The following is cast in stone:

- `auto_attribs=True`
- consistent (Python 3-style) hashing behavior on every Python version
- #428 

The following would be really nice to have:

- #368 but it might need too much thinking/research

What else is bugging you?

***

One thing that I just can't make up my mind is related to #223: should we make `slots=True` the default? I’m quite confident that in 99,9% of cases it's the right thing to do and will guide people to write better classes.

However on the other hand, our approach of rewriting classes breaks in certain scenarios, usually involving metaclasses.

So the question is, whether we want to tolerate a higher rate of bogus bug reports/help requests or make the default case nicer?

I welcome your input.

***

Finally a controversial idea: we could make `import attrs` Python 3 only. There isn't much baggage we'd get rid of but there is _some_ and 2020 is less than a year ahead. It would also allow us to embrace enums as part of our API.

sfermigier:
"we could make import attrs Python 3 only." -> +1. 

"It would also allow us to embrace enums as part of our API." -> That would be awesome.

berislavlopac:
I absolutely agree with your point on `slots=True`, but I would recommend not making it the default. I'm always surprised by how many people have no idea that this amazing feature exists; you need to at the very least explain in the documentation what it is anyway, so it can be used to recommend its use and guide people to write better classes. :slightly_smiling_face: 

I'm also in favour of Python3-only `import attrs`.

> What else is bugging you?

A question: My main issue with `attrs` has always been its decorator-based syntax and inability to instantiate using `__init__`. To be clear, it's my personal preference and not a general opinion on How Things Should Be Done [TM], but I've always liked the subclassing approach found in e.g. `schematics`. Also, I've often explained the concept of data classes (regardless of the implementation) as "ORM without the R", as it maps nicely with people's experience with Django and others. So I just have to ask, it there a possibility to have an optional method of implementing `attrs` as a parent/abstract class and/or a mixin?

Thanks for all the amazing work!

hynek:
> So I just have to ask, it there a possibility to have an optional method of implementing `attrs` as a parent/abstract class and/or a mixin?

[...]

Issue #408, opened 2018-07-13
Find better name for attr.dataclass


hynek:
So `attr.dataclass` was meant more or less as a joke, but I just caught myself using it too but the name is bad, because it’s confusing. It implies some kind of symmetry or compatibility to DCs which it doesn’t.

So here’s a nice bike shed, please help me finding a short, memorable, and meaningful color for it!

euresti:
`attr.auto`?

hynek:
> attr.auto?

That’s kinda my favorite too. It’s just sad that it doesn’t work with the attr prefix but that ship has somewhat sailed. I think I’ll have to re-visit the introduction of an `attrs` namespace.

hynek:
Which opens the question: if the namespace becomes `attrs` and you can do `from attrs import X`, what would be the best name for it? `auto` seems too short…OTOH `@attrs.auto` looks perfect. I wonder if in a perfect world, we could come up with something nice and symmetric to it – like `@attrs.explicit`?

The world around `attrs` has changed considerably since the invention of it's original names. It outgrew it’s scope more than I could *ever* predict when I was scheming with @glyph on IRC and the introduction of static types opened it up to new interesting use cases.

While I realize that introducing new APIs brings some confusion, there’s always the [cautionary tale](https://stackoverflow.com/questions/1755550/what-is-the-reasoning-behind-the-makefile-whitespace-syntax/1765566#1765566) of why Makefiles use tab characters.

***

After using it a bit, I believe that if Python has a future, static typing is unequivocally part of it and I want `attrs` to embrace it fully and not to feel like we’re dragging our heels. Obviously I do *not* intend to break backward compatibility but unless someone can talk me out of it, I’m very close to add a new, more forward-looking API and make it the primary one.

*I* for one, am very annoyed whenever I use the `attr` package name with names that don’t make any sense (like `attr.validators`) and the last time I got annoyed on each import, I abandoned the project in favor of `attrs`. ;)

hynek:
So I’ve been playing and I’m absolutely in love with *this* being the future canonical API for `attrs`:

```pycon
>>> import attrs
>>> import typing

>>> @attrs.auto
... class SomeClass:
...     a_number: int = 42
...     list_of_numbers: typing.List[int] = attrs.Factory(list)
...     a_number = attrs.attrib(converter=int)

>>> attrs.fields(SomeClass).a_number.type
<class 'int'>
```

***

[...]

PR #666, opened 2020-08-14
Add attr.define/mutable/frozen

hynek:
Prepare for import attrs y'all!

This is basically what we kinda agreed on in #408 and #487.

I'm open to comments and suggestions.

Also suggestions how to supply type stubs without a major copy/paste operation. 😳

### DIFF
--- a/src/attr/__init__.py
+++ b/src/attr/__init__.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 from functools import partial
 
 from . import converters, exceptions, filters, setters, validators
+from ._compat import PY2 as _PY2
 from ._config import get_run_validators, set_run_validators
 from ._funcs import asdict, assoc, astuple, evolve, has, resolve_types
 from ._make import (
@@ -39,7 +40,6 @@ s = attributes = attrs
 ib = attr = attrib
 dataclass = partial(attrs, auto_attribs=True)  # happy Easter ;)
 
-
 __all__ = [
     "Attribute",
     "Factory",
@@ -68,3 +68,8 @@ __all__ = [
     "validate",
     "validators",
 ]
+
+if not _PY2:
+    from ._auto import auto, frozen, mutable
+
+    __all__.extend((auto, frozen, mutable))
--- a/src/attr/__init__.py
+++ b/src/attr/__init__.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import, division, print_function
 
+import sys
+
 from functools import partial
 
 from . import converters, exceptions, filters, setters, validators
-from ._compat import PY2 as _PY2
 from ._config import get_run_validators, set_run_validators
 from ._funcs import asdict, assoc, astuple, evolve, has, resolve_types
 from ._make import (
@@ -69,7 +70,7 @@ __all__ = [
     "validators",
 ]
 
-if not _PY2:
-    from ._auto import auto, frozen, mutable
+if sys.version_info[:2] >= (3, 6):
+    from ._next_gen import define, frozen, mutable
 
-    __all__.extend((auto, frozen, mutable))
+    __all__.extend((define, frozen, mutable))
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -273,7 +273,7 @@ def define(
 ) -> Callable[[_C], _C]: ...
 
 mutable = define
-frozen = define  # XXX: not quite but good enough for now
+frozen = define  # they differ only in their defaults
 
 # TODO: add support for returning NamedTuple from the mypy plugin
 class _Fields(Tuple[Attribute[Any], ...]):
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -227,6 +227,53 @@ def attrs(
     getstate_setstate: Optional[bool] = ...,
     on_setattr: Optional[_OnSetAttrArgType] = ...,
 ) -> Callable[[_C], _C]: ...
+@overload
+def define(
+    maybe_cls: _C,
+    *,
+    these: Optional[Dict[str, Any]] = ...,
+    repr: bool = ...,
+    hash: Optional[bool] = ...,
+    init: bool = ...,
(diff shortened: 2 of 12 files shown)


lig:
My wishes:
* `kw_only` by default
* `auto_attribs` by default
* `attrs.Field()` for attributes
* `attrs.Class()` for class decorators
* allow to inherit something like `attrs.BaseClass` to be able to use `isinstance(obj, attrs.BaseClass)` later (basically allow to detect that we have attrs class)
* recursive `attrs.from_dict()`
* transparent builder API to simplify third party tools integration like converters between ORM models and attrs, serializer, etc
* regarding defaults, maybe allow building your own factory which provides required defaults to reuse it later.

hynek:
To be clear, I'm only looking for stuff that we might paint ourselves into the corner. Most of what you listed can be fixed at any time.

- kw_only=True is not happening, because I had it in [characteristic](https://github.com/hynek/characteristic) and people (including myself) hated it.
- ✔︎
- not sure what you mean? Just a different name? I'm indeed thinking to go for `attrs.field` or `attrs.Field` in the next APIs.
- you mean `@attrs.Class\nclass Foo`? After long deliberation on #408  it's gonna be `attrs.define`. Class class is repetitive.
- uh, `attr.has`?
- that's too complex, so we've delegated that to other packages like cattrs and try to help them get better at it in #653 
- that's interesting but out of scope for this issue
- you can do that with `functools.partial`. You just have to trick mypy into recognizing it but we've just merged some docs on that in #650

lig:
> * kw_only=True is not happening, because I had it in [characteristic](https://github.com/hynek/characteristic) and people (including myself) hated it.

:( it is how it is then. anyway, everybody needs different defaults

> * not sure what you mean? Just a different name? I'm indeed thinking to go for `attrs.field` or `attrs.Field` in the next APIs.

well, kinda. if it will be `import attrs` then `attrs.ib` doesn't work anymore. So, if `import attr` works as before then new API with `import attrs` needs something new and explicit.

> * you mean `@attrs.Class\nclass Foo`? After long deliberation on #408  it's gonna be `attrs.define`. Class class is repetitive.

Doesn't matter much, define works as well. `attrs.s` looks kinda "meh"

> * uh, `attr.has`?

Nice. Never noticed it. Maybe, missed in the docs and then didn't recognize it with autocomplete. The name feels a bot misleading to guess what it does straight away. Thanks for educating me on this:)

> * that's too complex, so we've delegated that to other packages like cattrs and try to help them get better at it in #653

Well, maybe the next point will help with it. However, it feels like it's not that complex and it would fully complement `asdict`. It's your call, no pressure here. `cattrs` worked for me.

> * that's interesting but out of scope for this issue

Fair enough. Shall I describe the idea in a separate issue?

> * you can do that with `functools.partial`. You just have to trick mypy into recognizing it but we've just merged some docs on that in #650

Making this work with fewer tricks would help. Docs are good, though:)


Thanks and keep up the good work!

hynek:
> > * not sure what you mean? Just a different name? I'm indeed thinking to go for `attrs.field` or `attrs.Field` in the next APIs.

> well, kinda. if it will be `import attrs` then `attrs.ib` doesn't work anymore. So, if `import attr` works as before then new API with `import attrs` needs something new and explicit.

Yeah, it's prob gonna be `attrs.field`.

> > * that's interesting but out of scope for this issue 

> Fair enough. Shall I describe the idea in a separate issue?

Sure!

> Thanks and keep up the good work!

💛

euresti:
This is neat. I will have to spend some time fixing the mypy plugin to get these to work.

src/attr/__init__.py
```diff
@@ -68,3 +69,8 @@
     "validate",
     "validators",
 ]
+
+if sys.version_info[:2] >= (3, 6):
```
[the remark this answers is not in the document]
hynek: I just feels more correct to compare tuples of identical cardinality instead of relying of implicit comparison rules.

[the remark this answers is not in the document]
s0undt3ch: Fair enough. Thanks!

hynek:
> This is neat.

Not sure if sarcasm. ;)

> I will have to spend some time fixing the mypy plugin to get these to work.

Do the annotations look OK so far? I don't think I can do anymore than release and wait for mypy, correct?

src/attr/_next_gen.py
```diff
@@ -0,0 +1,86 @@
...
+        return do_it(auto_attribs)
+
+    try:
+        return do_it(True)
+    except UnannotatedAttributeError:
+        return do_it(False)
+
+
+mutable = define
+frozen = partial(define, frozen=True, on_setattr=None)
```
euresti, line 86: The one caveat I noticed is that you can say:

```
@frozen(frozen=False)
```

But if you do that I'd definitely ask why you're doing that.

src/attr/__init__.pyi
```diff
@@ -227,6 +227,53 @@ def attrs(
...
+    auto_exc: bool = ...,
+    eq: Optional[bool] = ...,
+    order: Optional[bool] = ...,
+    auto_detect: bool = ...,
+    getstate_setstate: Optional[bool] = ...,
+    on_setattr: Optional[_OnSetAttrArgType] = ...,
+) -> Callable[[_C], _C]: ...
+
+mutable = define
+frozen = define  # XXX: not quite but good enough for now
```
euresti, line 276: From a type perspective this is fine.  I'd take out the XXX, and maybe mention that it's identical to the above with different defaults.

euresti:
Not sarcasm, I'm excited to not have to confuse people in my team with `@dataclass` (no, not that kind of dataclass)

src/attr/_next_gen.py
hynek: Yeah it's not intended as an official API but at this point I'm too lazy to make real APIs all around. We can re-elaborate when we do `import attrs`. :)

hynek:
I guess there's no point in procrastinating on this any further…merging and if nobody yells at me, I'll release 20.1 tomorrow.

PR #669, opened 2020-08-20
Add attr.field

hynek:
Almost forgot the last part of the NG API puzzle because it doesn't change substantially.

Does this look OK so far?


euresti:
Looks good. But you are not making my mypy job easy!!  :)  J/K progress is important.

hynek:
> Looks good. But you are not making my mypy job easy!! :) J/K progress is important.

Heh I assure you that I feel genuinely sorry whenever I do this. Your struggle is always on my mind. /o\

PR #681, opened 2020-09-01
Don't reset custom __setattr__ in slotted classes

### LINKED ISSUE #680: Using slots class overrides custom `__setattr__` in 20.1.0
In 19.3.0, it was possible to have a custom `__setattr__` on your `slots` class, but in 20.1.0 the custom `__setattr__` is replaced with the default one:

```python
import attr
@attr.s(slots=True)
class A:
    def __setattr__(self, key, value):
        print(f"{key}: {value}")
```

In 19.3.0:

```python
>>> A().a = 3
a: 3
```

In 20.1.0:

```python
>>> A().a = 3
AttributeError: 'A' object has no attribute 'a'
>>> A.__setattr__
<slot wrapper '__setattr__' of 'object' objects>
```

When `slots=False`, the same thing does not occur.

hynek:
The detection of whether or not we should be resetting `__setattr__` in slotted classes is currently wrong.

Fixes #680


src/attr/_make.py
```diff
@@ -669,24 +669,29 @@ def _create_slots_class(self):
...
-        # Traverse the MRO to check for an existing __weakref__ and
-        # __setattr__.
-        custom_setattr_inherited = False
+        # Check the bases if one of them has an attrs-made __setattr__ that
+        # needs to be reset.
+        # XXX: This can be confused by subclassing a slotted attrs class with
+        # XXX: a non-attrs class and subclass the resulting class with an attrs
+        # XXX: class.  See `test_slotted_confused for details.  For now that's
+        # XXX: OK with us.
+        for base_cls in self._cls.__bases__:
```
pganssle, line 678: I assume you had a good reason for this, but why `__bases__` and not `__mro__[1:-1]` here? It seems that that would fix the `test_slotted_confused` issue, and changing from `__bases__` to `__mro__[1:-1]` doesn't cause any other tests to fail.

I think it would be good to either add a test that demonstrates the behavior you're trying to preserve by using `__bases__` or switch to `__mro__`.

src/attr/_make.py
```diff
@@ -669,24 +669,29 @@ def _create_slots_class(self):
...
+            if not self._has_own_setattr and getattr(
```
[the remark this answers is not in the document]
hynek: Yes very true. I've let my dislike for indentation lead myself astray. ;)

src/attr/_make.py
```diff
@@ -669,24 +669,29 @@ def _create_slots_class(self):
...
+                base_cls, "__dict__", {}
```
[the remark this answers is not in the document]
hynek: So this is on the embarrassing side: it's not very long that I learned that classes actually always have `__dict__` and I somewhat just kept writing this kind of defensive code although it's not necessary. I'll handle the `__bases__` case tho, thanks.

tox.ini
```diff
@@ -1,6 +1,7 @@
 [pytest]
 addopts = -ra
 testpaths = tests
+xfail_strict = true
```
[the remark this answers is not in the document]
hynek: I suspected you might enjoy this one. ;)

src/attr/_make.py
hynek: Because `__attrs_own_setattr__` is a computed attribute.

Take this for instance:

```python
@attr.define
class A:
    x: int = attr.field(on_setattr=frozen)

@attr.define
class B(A):
    x: int = attr.field(on_setattr=NO_OP)
```

A has `__attrs_own_setattr__ = True` but B doesn't. If I walk the whole MRO, it would turn up True.

IIUC, my only option would be to walk the MRO and apply the full resolution logic myself – whose avoidance's the whole reason behind `__attrs_own_setattr__`?

src/attr/_make.py
```diff
@@ -669,24 +669,29 @@ def _create_slots_class(self):
             if k not in tuple(self._attr_names) + ("__dict__", "__weakref__")
         }
 
-        # Traverse the MRO to check for an existing __weakref__ and
-        # __setattr__.
-        custom_setattr_inherited = False
+        # Check the bases if one of them has an attrs-made __setattr__ that
```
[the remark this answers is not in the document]
hynek: I've went for:

```
        # If our class doesn't have an own __setattr__ (either from the user or
        # by us), check the bases, if one of them has an attrs-made
        # __setattr__, that needs to be reset.
```

Is that clear & OK?

src/attr/_make.py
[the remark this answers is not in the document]
pganssle: I would definitely not be embarrassed by that, the "classes always have `__dict__`" thing is not something I knew offhand, I just figured it out by dint of trying to understand what the code was doing in various situations (slots classes, built-ins, etc).

src/attr/_make.py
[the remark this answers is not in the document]
pganssle: I am slightly confused by "have an own `__setattr__`", because "an own X" is not a common idiom in my dialect of English (can't speak to BrE or European flavors), though I recognize that it's something of a term of art in this library. That said, the "If one of them has an attrs-made `__setattr__` that needs to be reset" clause removes all ambiguity, so it's not a big deal.

I would probably replace "have an own `__setattr__`" with "have its own `__setattr__`" or even "have its own implementation of `__setattr__`", but I don't think leaving it as-is would meaningfully impact my ability to understand the comment.

src/attr/_make.py
pganssle: > A has `__attrs_own_setattr__ = True` but B doesn't. If I walk the whole MRO, it would turn up True.

From my test, it seems like B *does* have `__attrs_own_setattr__ == True`, is this not intentional?

I think I've come up with a test that covers the situation you're trying to guard against, though:

```python
    def test_custom_setattr_subclass(self):
        @attr.define
        class A(object):
            x = attr.field(on_setattr=setters.frozen)

        @attr.define
        class B(A):
            x = attr.field(on_setattr=setters.NO_OP)
            def __setattr__(self, key, value):
                raise SystemError

        @attr.define
        class C(B):
            pass

        with pytest.raises(SystemError):
            C(1).x = 2
```

Probably worth adding that as a regression test so that no one shows up with a PR fixing the `xfail`-ing test by switching from `__bases__` to `__mro__[1:-1]` or something of that nature.

src/attr/_make.py
```diff
@@ -669,24 +669,28 @@ def _create_slots_class(self):
...
 
-        # Traverse the MRO to check for an existing __weakref__ and
-        # __setattr__.
-        custom_setattr_inherited = False
+        # If our class doesn't have an own __setattr__ (either from the user or
+        # by us), check the bases, if one of them has an attrs-made
+        # __setattr__, that needs to be reset.
+        # XXX: This can be confused by subclassing a slotted attrs class with
+        # XXX: a non-attrs class and subclass the resulting class with an attrs
+        # XXX: class.  See `test_slotted_confused for details.  For now that's
```
pganssle, line 677:
```suggestion
        # XXX: class.  See `test_slotted_confused` for details.  For now that's
```

Just noticed this unclosed backtick.

pganssle (dismissed):
Putting the LGTM on this one even though there are un-addressed comments because I think that my remaining comments are "nice to haves" ­— I think the "meat" of this PR looks good.

If you do make changes and want me to take another look, let me know.

src/attr/_make.py
hynek: > > A has `__attrs_own_setattr__ = True` but B doesn't. If I walk the whole MRO, it would turn up True.
> 
> From my test, it seems like B _does_ have `__attrs_own_setattr__ == True`, is this not intentional?

No it's not, since B doesn't have an own setattr. 🙈

> I think I've come up with a test that covers the situation you're trying to guard against, though:
> 
> ```python
>     def test_custom_setattr_subclass(self):
>         @attr.define
>         class A(object):
>             x = attr.field(on_setattr=setters.frozen)
> 
>         @attr.define
>         class B(A):
>             x = attr.field(on_setattr=setters.NO_OP)
>             def __setattr__(self, key, value):
>                 raise SystemError
> 
>         @attr.define
>         class C(B):
>             pass
> 
>         with pytest.raises(SystemError):
>             C(1).x = 2
> ```

I'm pretty sure we do have such a test; I just wonder why anything works at all.

hynek: Does f144cfa9a880ff6be576e2361c34ae8cb8e9614a make any sense to you?

src/attr/_make.py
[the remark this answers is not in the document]
pganssle: Perfect!

src/attr/_make.py
pganssle: That looks good, but (other than the strict XPASS), the test suite still succeeds when you use `for base_cls in self._cls.__mro__[1:-1]:`.

The only test that looks roughly like the one I suggested is [this one](https://github.com/python-attrs/attrs/blob/bfd7bb49b467e7ab586deb475c23974a91df4dab/tests/test_setattr.py#L232-L257), but the key to my test is that the "root" base class has an attrs-generated `setattr` and the intermediate class has a user-generated `__setattr__`. It's interesting, even with the `__bases__` approach, it seems that there's some problem when some of these are slots classes and others aren't:

```python
    @pytest.mark.parametrize("a_slots", [True, False])
    @pytest.mark.parametrize("b_slots", [True, False])
    @pytest.mark.parametrize("c_slots", [True, False])
    def test_setattr_inherited_do_not_reset_intermediate(self, a_slots, b_slots, c_slots):
        @attr.define(slots=a_slots)
        class A(object):
            x = attr.ib(on_setattr=setters.frozen)

        @attr.define(slots=b_slots)
        class B(A):
            x = attr.ib(on_setattr=setters.NO_OP)
            def __setattr__(self, key, value):
                raise SystemError

        @attr.define(slots=c_slots)
        class C(B):
            pass

        with pytest.raises(SystemError):
            C(1).x = 3
```

With the `__mro__` approach, all 8 of these tests fail, but with the `__bases__` approach, we still get failures whenever `C` is _not_ a slots class. I assume that this is not intentional, and that `__setattr__` inheritance _should not_ be reset in this way? I think maybe this "check the base classes for a `__setattr__`" logic may need to be extended to non-slots classes.

src/attr/_make.py
pganssle: I think this got lost in the shuffle, this still has an unclosed backtick.

hynek: WTH, I clicked commit suggestion. 😳

src/attr/_make.py
hynek: Argl, yeah so the problem is that we treat "user wrote a `__setattr__`" and "we wrote a `__setattr__`" the same way which is obviously wrong. We only ought to reset the latter one. This requires some deeper work.

hynek: Paul! PAUL! I think I've got it! Please check out https://github.com/python-attrs/attrs/pull/681/commits/c5b690ff93904bfdc716e656d84bd7007faf32ee

Now the tests also fail if you use `__mro__[1:-1]` AFAICT.

(I've also noticed that we've always used `__bases__` in the call to `type()` so no need to be defensive with `getattr()` unless someone complains)

pganssle (approved):
Nice, I think this should cover it (or at least covers the test scenarios I came up with; this stuff is complicated enough that I doubt I'll ever be confident that it will "survive contact with the enemy").

Definitely the subject for another PR (since the scope of this one has already significantly expanded), but I wonder if it would make sense to add some `hypothesis` tests that generate inheritance hierarchies and assert some properties of the subclasses. I imagine doing so might be a bit complicated, but there do seem to be some useful properties we could assert here.

hynek:
Thanks Paul!

PR #685, opened 2020-09-02
Make `zope` a semi-optional test dependency

pganssle:
For the admittedly small number of people who don't care about support for `zope.interface`, it is unnecessary extra work to support even a test-only dependency on `zope`.

The `zope`-specific tests are relatively contained, however, and with judicious use of pytest fixtures, we can make it so that `zope` is not a required dependency when skipping tests marked with `pytest.mark.zope`.

This allows users to test everything but the `zope`-specific tests with no `zope` dependency with:

```
$ TOX_AP_TEST_EXTRAS="tests_no_zope" tox -e <env> -- -m 'not zope'
```

I realize that this is unusual and maybe not a maintenance burden you want to carry upstream, but it would smooth the process of importing `attrs` upstream into a certain large company's monorepo... I'm fine with a response that's basically "Enough with your weird nonsense, Paul".

I didn't bother with documentation or a news fragment yet because I figured there's a high likelihood that this wouldn't be accepted.

# Pull Request Check List

### DIFF
--- a/changelog.d/685.change.rst
+++ b/changelog.d/685.change.rst
@@ -1,0 +1,1 @@
+``zope.interface`` is now a "soft dependency" when running the test suite; if ``zope.interface`` is not installed when running the test suite, the interface-related tests will be automatically skipped.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ multi_line_output=3
 use_parentheses=true
 
 known_first_party="attr"
-known_third_party=["attr", "hypothesis", "pytest", "setuptools", "six", "zope"]
+known_third_party=["attr", "hypothesis", "pytest", "setuptools", "six"]
 
 
 [tool.towncrier]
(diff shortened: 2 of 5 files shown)


hynek:
Would you mind inverting the logic and just make the Zope tests conditional on the presence of Zope.interface? That seems to be the more common approach nowadays anyways.

pganssle:
> Would you mind inverting the logic and just make the Zope tests conditional on the presence of Zope.interface? That seems to be the more common approach nowadays anyways.

Yeah, that's easy enough — the main difference between that approach and this one is that with this approach, it would be very hard to _accidentally_ disable the zope-related tests, whereas if you always disable the zope-related testing whenever `zope.interface` isn't present, anyone who fails to declare a test dependency on `zope.interface` will silently start skipping tests.

I figured this is a niche enough thing to want that it's better to require enthusiastic opt-in, but maybe it's not that big a deal if these tests sometimes get accidentally disabled in downstream testers?

hynek:
I don’t think anybody cares at all TBH. We’re just dragging it along for backward-compat.

pganssle:
OK, I've changed it to an auto-skipping fixture and added some docstrings and a changelog. If you think there's somewhere else that this should be explicitly documented, let me know. It didn't seem like it would fit much into `CONTRIBUTING.rst`, since this is more aimed at downstream redistributors than at contributors.

hynek (approved):

hynek:
Thanks!

Issue #668, opened 2020-08-20
Feedback to next generation APIs


hynek:
This is the issue to discuss the defaults and behavior of `attr.define()`, `attr.mutable()`, and `attr.frozen()` before they are finalized inside the new `attrs` namespace.

Feedback that we find justified will be spun out into separate issues.

Known issues:

- mypy doesn't know about the new-style APIs yet and there's nothing _we_ can do about it. However you can use [this mypy plugin](https://gist.github.com/hynek/1e3844d0c99e479e716169034b5fa963#file-attrs_ng_plugin-py) for now (for more details see https://www.attrs.org/en/stable/extending.html#wrapping-the-decorator). PR is submitted in https://github.com/python/mypy/pull/9396.
- `__eq__` and `__ne__` aren't correctly auto-detected (fixed) #670 
- Unannotated attributes fail if `define` is called with arguments/parantheses (fixed) #673
- Inheriting from frozen classes requires to manually set `on_setattr` to None (fixed). https://github.com/python-attrs/attrs/issues/668#issuecomment-678689981)

rouge8:
Overall I love the new APIs. When updating some code to use `@attr.define` instead of `@attr.s(auto_attribs=True)`, I was surprised by one behavior of slotted classes -- you don't seem to be able to use `unittest.mock.patch.object()` on methods on instances of slotted classes. It's not a result of anything in attrs (in the code snippet below I don't use attrs at all), but it might be worth noting in [the docs](https://www.attrs.org/en/stable/glossary.html#term-slotted-classes) since making `slots=True` the default behavior means attrs will be how many people encounter this for the first time.

```python
import unittest.mock

class Slotted:
    __slots__ = ()

    def method(self):
        return 1

s = Slotted()
assert s.method() == 1

with unittest.mock.patch.object(s, "method", return_value=3):
    assert s.method() == 3

# Errors with:
# Traceback (most recent call last):
#   File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/mock.py", line 1490, in __enter__
#     setattr(self.target, self.attribute, new_attr)
# AttributeError: 'Slotted' object attribute 'method' is read-only
```

hynek:
Ah yeah, thanks for that reminder, I will add it to the glossary!

Drino:
I've played around with new generation API a little, trying to migrate a part of my project onto it to feel the taste. 

[...]

PR #687, opened 2020-09-04
NG: make frozen classes comfortably subclassable

hynek:
This is a fix for an inconvenience that @euresti mentioned in https://github.com/python-attrs/attrs/issues/668#issuecomment-678689981.

`on_setattr=validate` gets into the way if you want to pass `define(frozen=True)` or if you subclass a frozen class using `define`.

The new logic should catch these problems.

### DIFF
--- a/src/attr/_next_gen.py
+++ b/src/attr/_next_gen.py
@@ -82,7 +82,7 @@ def define(
         """
         nonlocal frozen, on_setattr
 
-        had_on_setattr = on_setattr is not None
+        had_on_setattr = on_setattr not in (None, setters.NO_OP)
 
         # By default, mutable classes validate on setattr.
         if frozen is False and on_setattr is None:
@@ -98,7 +98,6 @@ def define(
                         "(frozen-ness was inherited)."
                     )
 
-                frozen = True
                 on_setattr = setters.NO_OP
                 break
 
--- a/src/attr/_next_gen.py
+++ b/src/attr/_next_gen.py
@@ -10,7 +10,7 @@ from functools import partial
 from attr.exceptions import UnannotatedAttributeError
 
 from . import setters
-from ._make import NOTHING, attrib, attrs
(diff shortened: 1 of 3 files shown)


src/attr/_next_gen.py
```diff
@@ -73,11 +74,37 @@ def do_it(cls, auto_attribs):
...
+        # and disable on_setattr.
+        for base_cls in cls.__bases__:
+            if base_cls.__setattr__ is _frozen_setattrs:
+                if had_on_setattr:
+                    raise ValueError(
+                        "Frozen classes can't use on_setattr "
+                        "(frozen-ness was inherited)."
+                    )
+
+                frozen = True
```
euresti, line 101: Just 2 small questions.  
Couldn't this be just
```
on_setattr = None
```

I believe frozen will be inherited anyway right?  Or are you figuring well we've already spent time detecting it so we might as well use it?

src/attr/_next_gen.py
```diff
@@ -73,11 +74,37 @@ def do_it(cls, auto_attribs):
...
     def wrap(cls):
-        # Making this a wrapper ensures this code runs during class creation.
+        """
+        Making this a wrapper ensures this code runs during class creation.
+
+        We also ensure that frozen-ness of classes is inherited.
+        """
+        nonlocal frozen, on_setattr
+
+        had_on_setattr = on_setattr is not None
```
euresti, line 85: Shouldn't we check for NO_OP here too?

euresti (approved):
Just 2 small questions. But otherwise it looks good.

src/attr/_next_gen.py
hynek: Astute as always!

src/attr/_next_gen.py
hynek: You're right, let's keep it auto-detected.

PR #700, opened 2020-10-12
[RFC] kw_only python 2 backport

Drino:
Hello!

Thank you for bringing an awesome package and supporting Python 2 in 2020!

In my experience I find kw_only attributes really addictive, as they allow to inherit from classes ignoring attributes with defaults. Unfortunately, they don't work on Python 2, so it's actually impossible to write some compatible code. 

Even writing `__init__` by hand:
```python
@attr.s(kw_only=not PY2)
class Base(object):
    kw_only1 = attr.ib(default=None)
    kw_only2 = attr.ib(default=42)

@attr.s(init=not PY2)
class Child(Base):
    attr1 = attr.ib()
    attr2 = attr.ib()

    if PY2:
        def __init__(self, attr1, attr2, **kwargs):
            super(Child, self).__init__(**kwargs)
            self.attr1, self.attr2 = attr1, attr2
```
Will fail, as `attrs` disallows mandatory attributes after an attribute with a default value or factory. Even this non-working code snippet seems as boilerplate, so in this PR I propose a backport of kw_only `attr.ib` to Python 2.

There are four approaches to implementation that I've tested:
1. Just move kw_only attributes to the end of `__init__`. It's the fastest approach, on my benckmark (3 positional args, 3 kw_only args) it's about ~100ns for argument parsing. However, it doesn't allow kw_only attributes without default values and also doesn't make them real kw_only.
2. Pop from a dictionary of kwargs. This is the slowest approach with a somewhat-reasonable error message, ~400ns for argument parsing.
3. On every `__init__` call create a function that takes kw_only attributes and returns a tuple of unpacked arguments (implemented in this PR). It's slightly faster than previous approach, ~390ns for argument parsing.
4. Cache that function on class instance - it saves ~40-50ns, giving 340ns. I plan to rewrite this PR to use this approach. 

Both approaches with function have error message lacking some information, however it's a default py2 error and the constructor call should be somewhere in the stacktrace.

Is this a desired feature or should it not happen? Should I add some fancy error handling? 

P.S. I'm unfamiliar with changelog.d - should I just add an rst file to the dir? 

# Pull Request Check List

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -146,7 +146,7 @@ On Python 3 it overrides the implicit detection.
 Keyword-only Attributes
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes:
+You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes (even on Python 2):
 
 .. doctest::
 
@@ -159,8 +159,6 @@ You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-
     TypeError: A() missing 1 required keyword-only argument: 'a'
     >>> A(a=1)
     A(a=1)
-
-``attrs`` supports ``kw_only`` attributes even on Python 2, though error messages are default python error messages and may be a bit confusing.
 
 ``kw_only`` may also be specified at via ``attr.s``, and will apply to all attributes:
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -146,7 +146,7 @@ On Python 3 it overrides the implicit detection.
 Keyword-only Attributes
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-When using ``attrs`` on Python 3, you can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes:
+You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes:
 
 .. doctest::
 
@@ -159,6 +159,8 @@ When using ``attrs`` on Python 3, you can also add `keyword-only <https://docs.p
     TypeError: A() missing 1 required keyword-only argument: 'a'
     >>> A(a=1)
     A(a=1)
+
+``attrs`` supports ``kw_only`` attributes even on Python 2, though error messages are default python error messages and may be a bit confusing.
 
 ``kw_only`` may also be specified at via ``attr.s``, and will apply to all attributes:
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -146,7 +146,7 @@ On Python 3 it overrides the implicit detection.
 Keyword-only Attributes
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes (even on Python 2):
+You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes:
 
 .. doctest::
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1947,10 +1947,8 @@ if PY2:
         """
         assert kw_only_args
         lines = ["try:"]
-        lines += [
-            "    " + _unpack_kw_only_py2(*arg.split("="))
-            for arg in kw_only_args
-        ]
+        for arg in kw_only_args:
+            lines.append("    " + _unpack_kw_only_py2(*arg.split("=")))
         lines += """\
 except KeyError as _key_error:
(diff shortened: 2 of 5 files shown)


hynek:
I honestly don’t care about the micro-performance on Python 2 since it’s only run on class creation anyways. So I say keep it as simple as possible.

But it looks really gnarly…could you isolate it into a function please (not the much simpler Python 3 path, I don’t want the function call overhead in the common case)?

Speaking of performance: please avoid the usage of str.format if you can. I know attrs is still full of it but it’s slow af and I don’t want it in new code. 🙏

You can find details on changelog.d in our contribution guide: https://www.attrs.org/en/latest/contributing.html#changelog

Drino:
@hynek thank you for the answer!

As we don't care about performance in this case I rewritten it to poping from dict and added fancy error-handling - now error-message is much closer to python2/python3 native ones for `__init__()`.  

Am I correct that this change doesn't need a testing strategy?

Drino:
@hynek could you please help me with the failing codecov check? 

I don't understand, why the coverage decreased. I've added two functions, both are PY2-only and hidden under `if PY2:`. The only branching depends on presence of default value, but I we have tests for both cases (`kw_only` with default and `kw_only` without default).

docs/examples.rst
```diff
@@ -146,7 +146,7 @@ On Python 3 it overrides the implicit detection.
 Keyword-only Attributes
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-When using ``attrs`` on Python 3, you can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes:
+You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes (even on Python 2):
```
hynek, line 149: There's no need to spell Py2 out, we don't do it for other features either. :)

```suggestion
You can also add `keyword-only <https://docs.python.org/3/glossary.html#keyword-only-parameter>`_ attributes:
```

src/attr/_make.py
```diff
@@ -1924,6 +1925,45 @@ def _assign_with_converter(attr_name, value_var, has_on_setattr):
...
+            attr_name,
+            attr_name,
+            arg_default,
+        )
+
+    def _unpack_kw_only_lines_py2(kw_only_args):
+        """
+        Unpack all *kw_only_args* from _kw_only dict and handle errors.
+        """
+        assert kw_only_args
```
hynek, line 1948: I don't think we do something like this in attrs anywhere else?

src/attr/_make.py
```diff
@@ -1924,6 +1925,45 @@ def _assign_with_converter(attr_name, value_var, has_on_setattr):
...
+            arg_default = ""
+        return "%s = _kw_only.pop('%s'%s)" % (
+            attr_name,
+            attr_name,
+            arg_default,
+        )
+
+    def _unpack_kw_only_lines_py2(kw_only_args):
+        """
+        Unpack all *kw_only_args* from _kw_only dict and handle errors.
```
hynek, line 1946: Could you please elaborate/outline how it works? Because it's not entirely clear to me and I don't want to figure it out whenever I look at the code. :)

(leave the first like. Just add a blank line and explain what you're doing.)

src/attr/_make.py
```diff
@@ -1924,6 +1925,45 @@ def _assign_with_converter(attr_name, value_var, has_on_setattr):
...
+        """
+        assert kw_only_args
+        lines = ["try:"]
+        lines.extend(
+            "    " + _unpack_kw_only_py2(*arg.split("="))
+            for arg in kw_only_args
+        )
+        lines += """\
+except KeyError as _key_error:
+    _msg = '__init__() missing required keyword-only argument: %s'
```
hynek, line 1956: I don't think creating a variable (`_msg`) is necessary here?

src/attr/_make.py
```diff
@@ -1924,6 +1925,45 @@ def _assign_with_converter(attr_name, value_var, has_on_setattr):
...
+    raise TypeError(_msg % _key_error)
+if _kw_only:
+    _msg = '__init__() got an unexpected keyword argument %r'
```
hynek, line 1959: same

hynek (changes requested):

src/attr/_make.py
Drino: I had problems with linter here, as line was too long.
Put blacked code in triple-quotes.

src/attr/_make.py
Drino: Removed

docs/examples.rst
Drino: I found it a little bit confusing, as in the same sentence we have a link to official python3 docs for python3 term, but fixed.

src/attr/_make.py
Drino: Done, hope it's not too confusing.

hynek (approved):

hynek:
Thanks, let's hope it doesn't fall apart in production. :)

PR #702, opened 2020-10-16
Make the SVG logo use the current foreground color

pradyunsg:
Okay, the PR template is nice!

This is mostly just me testing out whether this rough idea I had actually works.


pradyunsg:
Nope! `` tags don't let `currentColor` flow through apparently.

PR #706, opened 2020-10-21
Refactor Hypothesis strategy for self-tests

Zac-HD:
This PR refactors two Hypothesis strategies which are used to test `attrs`, in one commit each:

1. In `simple_classes()`, we draw boolean flags for `frozen`, `slots`, and `weakref` *even if we will not use them*.  This means that from Hypothesis' perspective the data structure does not depend on whether the flags will be used, which in turn makes mutation and shrinking logic a little more efficient.

2. Refactoring `_create_hyp_nested_strategy()` to use the `@st.composite` decorator.  This makes the code considerably shorter, and IMO the imperative style is also much easier to read in this case.

### DIFF
--- a/tests/strategies.py
+++ b/tests/strategies.py
@@ -46,14 +46,8 @@ def maybe_underscore_prefix(source):
         to_underscore = not to_underscore
 
 
-def _create_hyp_class(attrs):
-    """
-    A helper function for Hypothesis to generate attrs classes.
-    """
-    return make_class("HypClass", dict(zip(gen_attr_names(), attrs)))
-
-
-def _create_hyp_nested_strategy(simple_class_strategy):
+@st.composite
+def _create_hyp_nested_strategy(draw, simple_class_strategy):
     """
     Create a recursive attrs class.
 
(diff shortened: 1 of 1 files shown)


hynek (approved):

hynek:
ZOMG Zac, thank you so much for looking at it! I honestly don't fully understand our Hypothesis code but I'm very aware, that it massively slows down our test runs (especially test_funcs is embarrassingly slow). A full test run (sans coverage) clocks in around 3mins on my notebook and that's after I already simplified it with my limited knowledge.

Unfortunately, I don't really understand the code and its complexities. So if you have the time to have a deeper look to improve and speed-up our hypothesis code, I'd be super grateful!

Zac-HD:
Very happy to help :slightly_smiling_face: 

---------

Having looked at the tests, I'm not sure whether it's good or bad news to report that everything looks good - there's no obvious problems to fix.  This is within a normal performance range for generating nested data (classes with attributes which themselves have attributes, including perhaps dicts of metadata...).

If runtime is an issue, you could have a [settings profile](https://hypothesis.readthedocs.io/en/latest/settings.html#settings-profiles) which runs e.g. 30 examples (~3x faster than the default `max_examples=100`) in local development, and turn it up to 100 or even 1000 in CI where the feedback time is not such a problem.

Running `tox -e py37 -- --durations=20 -k test_funcs --hypothesis-show-statistics -m hypothesis` shows that for the slow functions, >95& of the total runtime is spent generating data (including e.g. calling `attrs` constructors).  If you run this a few times you'll also notice that runtime for each test is quite variable, because total runtime tends to be dominated by the relatively few cases which generate much larger inputs than average.  You can stabilise this for profiling by running in deterministic mode with the `--hypothesis-seed=n` argument.

Options are some combination of 

1. Just live with the current performance.
2. Spend some quality time with a profiler, and pick a trade off between test effectiveness and test runtime that you like better.
3. Implement the profiles trick and explicitly run for a long time locally if nervous.
   (and [HypoFuzz](https://hypofuzz.com/) has free-for-OSS licences as well, if you'd like to try out the beta)

I'd choose (1) if it's a minor niggle, or (3) if performance is a serious issue for your workflow.

PR #710, opened 2020-11-02
Infer type annotations from converters

nosewings:


docs/init.rst
```diff
@@ -301,6 +301,21 @@ Arguably, you can abuse converters as one-argument validators:
...
+You can override this by passing the ``type`` argument or by giving an explicit type annotation.
+
+.. doctest::
+
+   >>> def str2int(x: str) -> int:
+   ...     return int(x)
+   >>> @attr.s
+   ... class C(object):
+   ...     x = attr.ib(converter=str2int)
+   ...     y = attr.ib(converter=str2int, type=int)
```
hynek, line 314: I don't think this is correct behavior. The type parameter is equivalent to using a type annotation. IOW: it's what's stored on the class, _not_ what the initializer is accepting.

nosewings: That makes sense, and I've updated PR to reflect that. It's worth noting, though, that the `type` argument now becomes a no-op if a converter is present. I suppose that was already the case, though.

The only issues I see now are:
1. If `inspect.signature` finds that the converter doesn't take any arguments, an `IndexError` will be thrown at line 2208. In most cases, an error would have occurred anyway -- but it would have happened later, and it would have been more readable (e.g., `f() takes 0 positional arguments but 1 was given`).
2. I haven't special-cased `pipe`, because I don't see a good way to do that without modifying the definition of `pipe` (e.g., turning it into a callable class).

Drino: 2 - why don't just do `update_wrapper` in `pipe`? 
```
if not PY2 and converters:
    pipe_converter.__anno
    functools.update_wrapper(pipe_converter, converters[0])
```

Return-type annotation may be inaccurate, but the arguments annotations would anyway flow into `__init__`. To fix return ones, something like this can be done:
```
    if 'return' in converters[-1].__annotations__:
        pipe_converter.__annotations__['return'] = converters[-1].__annotations__['return']
    else:
        del converters[-1].__annotations__['return']
```

but I'm not sure if this approach is good enough.

hynek: Hm the update_wrapper is intriguing, but the problem is that it also copies over `__name__` etc which would make it confusing.

But I guess we could do just a `pipe_converter.__annotations__ = converters[0].__annotations__` and then the handling of return you've outlined?

Drino: You can just take only annotations on update_wrapper:
```
In [2]: def test_me(a: int, b: str): pass                                                                                                                                                                  

In [3]: def test_wrapper(*args, **kwargs): pass                                                                                                                                                            

In [4]: update_wrapper(test_wrapper, test_me, assigned=('__annotations__',))                                                                                                                                   
Out[4]: <function __main__.test_wrapper(a: int, b: str)>
```

Please, note that `update_wrapper`also copies signature information (which is important in this case, as annotations are parsed from signature) and just copying `__annotations__` wouldn't do the trick.

hynek: Sure, why not then.

nosewings:
I went ahead and implemented inference for `pipe` and `optional` (`default_if_none` isn't possible without knowing the type of the default argument). I don't think the `update_wrapper` approach will work, because, e.g., the function returned by `pipe` takes exactly one positional argument named `val`, not any other arguments that the first wrapped converter might take. Instead, I had to use `inspect` and copy over the necessary annotations manually.

nosewings:
I suppose the other option would be to make `pipe` (and `optional`) take `*args, **kwargs`; then the `update_wrapper` approach would make sense (I think).

hynek:
@Drino would you mind having a look?

docs/init.rst
```diff
@@ -301,6 +301,20 @@ Arguably, you can abuse converters as one-argument validators:
    ValueError: invalid literal for int() with base 10: 'x'
 
 
+If a converter's first argument has a type annotation, that type will appear in the signature for ``__init__``.
+A converter will override an explicit type annotation or ``type`` argument.
```
Drino, line 305: nit: This changes current behaviour even if converters do not have type annotations. Though, these changes apply to `typing.get_type_hints(cls.__init__)`, not `typing.get_type_hints(cls)` and I'm not sure anybody will actually get anything broken by that.

src/attr/_make.py
```diff
@@ -2762,4 +2786,36 @@ def pipe_converter(val):
 
         return val
 
+    if not PY2:
+        if not converters:
+            # If the converter list is empty, pipe_converter is the identity.
+            A = typing.TypeVar("A")
```
Drino, line 2792: nit: I don't like how it works with empty pipe, as it will put `TypeVar("A")` in `__init__` annotation. Can't figure a much better solution, though.

Drino:
I'm OK with parsing only argument type (e.g. it may be better for 3-argument converters when they will be added), though in my personal opinion `*args, **kwargs` approach is a little bit more elegant.

Added a couple of pesky comments, but anyway it looks good to me.

@nosewings thank you for the great PR :)

docs/init.rst
hynek: Hm if that's so important, maybe it would warrant a section in `types.rst`. Changelogs are terrible for long-term knowledge keeping. But I don't want to put that on poor @nosewings.

hynek:
Thanks everyone!

PR #712, opened 2020-11-04
Fix exception chaining on PyPy

### LINKED ISSUE #703: Frozen Exceptions cannot be thrown from Exception Handler in PyPy
Consider the following code, which generates some random exception, which is then rethrown as a custom attrs exception class (in this case suppressing the context using [PEP 409](https://www.python.org/dev/peps/pep-0409/)'s `from None`):
```python
import attr

@attr.s(frozen=True, slots=True, auto_exc=True)
class MyException(BaseException):
    msg:str = attr.ib()

try:
    next(iter([]))
except StopIteration:
    raise MyException("test") from None
```
In CPython, this works as expected:
```
  File "scratch.py", line 10, in <module>
    raise MyException("test") from None
__main__.MyException: test
```
On PyPy, this yields a very weird, different exception:
```
  File "scratch.py", line 8, in <module>
    next(iter([]))
StopIteration

During handling of the above exception, another exception occurred:

  File "scratch.py", line 10, in <module>
    raise MyException("test") from None
  File ".venv-pypy3/site-packages/attr/_make.py", line 528, in _frozen_setattrs
    raise FrozenInstanceError()
attr.exceptions.FrozenInstanceError
```
I assume this is related to how `__context__` and `__cause__` are set when (re-)raising exceptions (see also the linked PEP for some info on that). The PEP says "To support raise Exception from None, `__context__` will stay as it is, but `__cause__` will start out as Ellipsis and will change to None when the raise Exception from None method is used.", so it's weird that only PyPy seems to be wrong here (which is also the reason I'm reporting this against attrs and not PyPy). Maybe CPython doesn't completely adhere to that and doesn't change the value in this case - or changes it bypassing the high-level `__setattr__` API. Note that leaving the `from None` out doesn't change the weird behaviour. Maybe the `__setattr__` frozen check should be skipped for `__context__` and `__cause__` of exceptions?

The versions of the interpreters I'm using are:
```
$ pypy3 -V
Python 3.6.9 (78d4c48fa091, Apr 30 2020, 07:55:31)
[PyPy 7.3.1 with GCC 10.0.1 20200328 (Red Hat 10.0.1-0.11)]
$ python3.9 -VV
Python 3.9.0 (default, Oct  6 2020, 00:00:00) 
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)]
```

hynek:
Fixes #703

### DIFF
--- a/changelog.d/703.change.rst
+++ b/changelog.d/703.change.rst
@@ -1,0 +1,1 @@
+``raise from`` now works on frozen classes on PyPy.
--- a/changelog.d/712.change.rst
+++ b/changelog.d/712.change.rst
@@ -1,0 +1,1 @@
+``raise from`` now works on frozen classes on PyPy.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ source = ["src", ".tox/*/site-packages"]
 [tool.coverage.report]
 show_missing = true
 exclude_lines = [
+    "pragma: no cover",
     # PyPy is unacceptably slow under coverage.
     "if PYPY:",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,10 @@ source = ["src", ".tox/*/site-packages"]
 
 [tool.coverage.report]
 show_missing = true
+exclude_lines = [
+    # PyPy is unacceptably slow under coverage.
+    "if PYPY:",
+]
 
 
 [tool.black]
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -91,7 +91,7 @@ if PY2:
         res.data.update(d)  # We blocked update, so we have to do it like this.
         return res
 
-    def just_warn(*args, **kw):  # pragma: nocover
+    def just_warn(*args, **kw):  # pragma: no cover
         """
         We only warn on Python 3 because we are not aware of any concrete
         consequences of not setting the cell on Python 2.
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -132,7 +132,7 @@ def make_set_closure_cell():
     """
     # pypy makes this easy. (It also supports the logic below, but
     # why not do the easy/fast thing?)
-    if PYPY:  # pragma: no cover
+    if PYPY:
 
         def set_closure_cell(cell, value):
             cell.__setstate__((value,))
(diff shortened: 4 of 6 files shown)


N-Coder:
Looks good:
```
$ pypy3 -m venv venv-pypy3
$ source venv-pypy3/bin/activate
(venv-pypy3) $ pip install git+https://github.com/python-attrs/attrs.git@pypy-exc-chain
Collecting git+https://github.com/python-attrs/attrs.git@pypy-exc-chain
  Cloning https://github.com/python-attrs/attrs.git (to revision pypy-exc-chain) to /tmp/pip-req-build-fevtddhv
  Running command git clone -q https://github.com/python-attrs/attrs.git /tmp/pip-req-build-fevtddhv
  Running command git checkout -b pypy-exc-chain --track origin/pypy-exc-chain
  Switched to a new branch 'pypy-exc-chain'
  Branch 'pypy-exc-chain' set up to track remote branch 'pypy-exc-chain' from 'origin'.
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: attrs
  Building wheel for attrs (PEP 517) ... done
  Created wheel for attrs: filename=attrs-20.3.0.dev0-pp373-none-any.whl size=49552 sha256=26c81e0c48c2f4470901fd8cfb9b9ae6990c3d28185d6fcb8acacdefc4101a79
  Stored in directory: /tmp/pip-ephem-wheel-cache-ccltyff7/wheels/14/a6/cc/b466b3f48ab1403375e5b30128ef59ff27a61e824aa62c0bfe
Successfully built attrs
Installing collected packages: attrs
Successfully installed attrs-20.3.0.dev0
WARNING: You are using pip version 19.3.1; however, version 20.2.4 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
(venv-pypy3) $ python
Python 3.6.9 (78d4c48fa091, Apr 30 2020, 07:55:31)
[PyPy 7.3.1 with GCC 10.0.1 20200328 (Red Hat 10.0.1-0.11)] on linux
>>>> import attr
>>>> 
>>>> @attr.s(frozen=True, slots=True, auto_exc=True)
.... class MyException(BaseException):
....     msg:str = attr.ib()
....     
>>>> try:
....     next(iter([]))
.... except StopIteration:
....     raise MyException("test") from None
....     
  File "<stdin>", line 4, in <module>
MyException: test
>>>> 
```
Same for CPython 3.8.6. Thanks! <3

hynek:
Thank you for your quick turnaround!

PR #719, opened 2020-11-16
logo: Split `path` into multiple paths

Turbo87:
Passing SVG files through `svgo` often makes them harder to edit for applications like InkScape or Sketch. This commit splits the single `path` in the files into multiple paths to resolve this issue. The file is now slightly larger, but should be a lot more maintainable. If more compression is needed then svgo should be applied in a build step, but the repository should contain the raw source file.

see https://twitter.com/hynek/status/1328230837150773254 😉


sscherfke:
LGTM.  Tested both files with Inkscape 1.0.1 on macOS.

hynek:
But do the ones from the repo make any problems? As I've written on Twitter: my problem are the originals I got from logojoy. The ones in the repo should work AFAIK?

Turbo87:
@hynek sorry, apparently I misunderstood your original tweet. if the SVGs from this repo work for you then why not use them instead of the logojoy originals?

> But do the ones from the repo make any problems?

when I imported them in Sketch they seemed broken, but in Inkscape they indeed appeared to work. the new variants work fine in both applications.

hynek:
> why not use them instead of the logojoy originals?

I do! But I needed the other variant for dark mode, so I had to go back to the originals.

> when I imported them in Sketch they seemed broken, but in Inkscape they indeed appeared to work. the new variants work fine in both applications.

Well, that sounds like a good argument to me. Thanks!

PR #728, opened 2020-12-06
Implement @converter decorator

taliastocks:
Implement @converter decorator as described in #240. If you like this approach, I'll be happy to update the documentation etc.

This is a less sophisticated approach than #404, which has been in draft since 2018 and supports self arguments.

# Pull Request Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .cache
 .coverage*
 .hypothesis
+.idea/
 .mypy_cache
 .pytest_cache
 .tox
@@ -11,3 +12,5 @@ dist
 docs/_build/
 htmlcov
 pip-wheel-metadata
+Pipfile
+Pipfile.lock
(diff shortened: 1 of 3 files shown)


taliastocks:
(I'm not sure I understand why the tox lint is failing for python 3.8.)

hynek:
Oof, firstly thanks for putting in the effort! I'll start with answering your question: you can click on "Details" of the failing check, then you can see that "ERROR:   lint: commands failed". So you click the chevron next to "tox: lint" and there you can see that black failed:

```
  black....................................................................Failed
  - hook id: black
  - files were modified by this hook
  
  reformatted /home/runner/work/attrs/attrs/tests/test_make.py
```

Arguably this should be much more obvious, but I'm not aware of anything in GitHub Actions that would help… :(

---

As for #404, while it looks a bit stalled now, it's entirely _my_ fault that it's in draft since 2018: I kinda decided I didn't want it in. I changed my mind and we've revived it only at the end of October and Kyle's last push is only a little longer ago than a month.  I would really like to see #404 finished and not merge subsets of his work.

I'm not 100% clear on what's stuck there rn (I'm busy getting a new structlog release out), but do you think you could maybe help out over there so we can finish the whole thing?

Issue #725, opened 2020-11-30
Calling super.__init__ before attrs __init__ (or, __attrs_pre_init__)


indigoviolet:
Hi,

(Thanks for your work on attrs, it's fantastic, I'm a happy sponsor)

I'm subclassing `pytorch`'s `nn.Module` and it has some "magic" treatment of members that are of certain types, which boils down to it [requiring](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L825) invoking `super().__init__()` *before* assigning instance variables, so I can't just use `__attrs_post_init__` for this.

I've read https://github.com/python-attrs/attrs/issues/695#issuecomment-702617747, https://github.com/python-attrs/attrs/issues/640#issuecomment-615281984, https://github.com/python-attrs/attrs/issues/167 etc and I'm generally in agreement about composition vs. inheritance, but I'd love to be able to use the convenience of attrs in this situation: a library class that I need to subclass but have no control over. 

Naively, it seems to me that this could be satisfied by having an `__attrs_pre_init__` hook that receives all the constructor arguments.  I assume you have considered this - would you be open to a PR adding it (it seems relatively straightforward following the code for `__attrs_post_init__`)?

hynek:
Thank you for being a sponsor, happy to boot!

As stated in https://github.com/python-attrs/attrs/issues/393#issuecomment-546710431, I think it would be overall a nicer design to allow for telling `attrs` to write its `__init__` into a `__attrs_init__` that the user calls themselves?

That said I'm somewhat OK with doing `__attrs_pre_init__` too, I'm just busy catching up with my issues on `structlog` right now – unfortunately I don't seem to have the mental capacity to do large-scale work on two big FOSS projects at once. 🙃

Drino:
I would like to notice that while `__attrs_init__` is a great feature, it forces user either to lose `__init__` hints from IDE or to write boilerplate `__init__` signature.

hynek:
IOW types alone make it worth having both in any case?

indigoviolet:
Yes, my intention was to move current `__init__`  to be `__attrs_init__`, then autogenerate an `__init__` like:

```
def __init__(self, ...args as before...):
     self.__attrs_init__(...args again...)
     self.__attrs_post_init__()
```

hynek:
[...]

PR #731, opened 2020-12-13
__attrs_init__()

indigoviolet:
Here's a first pass for #725. @hynek, please take a look to see if you like this approach, and I can then do all the polish work below. Tests were passing at least locally, for py38. ~Please ignore all the `poetry`-related noise, I'll get rid of that.~

It's a bit ugly for a few reasons:

1.  I wanted to have `{args}` in both the auto-generated `__init__` and `__attrs_init__`, but didn't want to repeat or call that code twice; and it seems non-trival to factor out.

2. We have to always generate `__attrs_init__` (so that the user-provided `__init__` can refer to it), but conditionally generate `__init__`  as before. 

3. Because of (2), `Factory` now calls `_make_init` (even though it has defined `__init__`), but `_make_init` refers to `Factory`, so we need a small hack for this.

I think that `__attrs_pre_init__` will be cleaner to implement, and I'm not sure I fully see the advantages of `__attrs_init__` over `__attrs_pre_init__`. Thoughts?

# Pull Request Check List

-  [x] Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``).


.envrc
```diff
@@ -0,0 +1 @@
+layout_poetry
```
indigoviolet, line 1: ignore this file

.python-version
```diff
@@ -0,0 +1 @@
+3.8.1
```
indigoviolet, line 1: ignore this file

pyproject.toml
```diff
@@ -61,3 +61,36 @@ profile = "attrs"
         directory = "change"
         name = "Changes"
         showcontent = true
+
```
indigoviolet, line 64: ignore all changes here

indigoviolet:
Looks like I need to handle kw-only args from py2 in some way.  (basically unpacking `_kw_only` in `__init__` before passing to `__attrs_init__`

.python-version
pganssle: If you do this frequently to projects, it's probably a good idea to put it in your global .gitignore rather than committing it and pushing it as part of the PR.

indigoviolet: Actually I usually want it checked in. Just not in this case.

indigoviolet: (Removed the noise though)

hynek:
First an important lifehack: the solution to your `.envrc` problem is called `.git/info/exclude` which is a directory-specific ignore file. I use it everywhere too.

---

As for the PR: it is quite intrusive, are you sure you wouldn't be fine with `__attrs_pre_init__` for now? I think that should be a lot easier to implement and has the mentioned upside of having proper types. I wanna do both but the length of the PoC is somewhat daunting.

indigoviolet:
I'm fine with `__attrs_pre_init__` as mentioned in the issue and summary above, I was under the impression that you preferred this approach. I will abandon this and replace with a different PR for that instead.

Note, however, that this PR does maintain the types for `__init__`. In fact that accounts for much of its intrusiveness, since we need to do the same generation for both `__init__`
and `__attrs_init__`.

hynek:
>  I was under the impression that you preferred this approach.

Your impression is 100% on point! It's just that we need both anyways and `__attrs_pre_init__` seems like the lower hanging fruit?

> Note, however, that this PR does maintain the types for `__init__`.

Maybe I misunderstand, but the point of `__attrs_init__` as original intended is that `attrs` does _not_ write an `__init__` at all and provides the user with with an `__attrs_init__` instead.

IOW instead of:

```python
@attr.define
class C:
    x: int

    def __attrs_pre_init__(self):
        do_a()
    def __attrs_post_init__(self):
        do_b()
```

The user can write:

```python
@attr.define(init=some_magic_value_we_need_to_agree_on)
class C:
    x: int
    def __init__(self, x: int) -> None:
        do_a()
        self.__attrs_init__()
        do_b()
```

Are we on the same page at all or am I being confusing?

indigoviolet:
> Maybe I misunderstand, but the point of `__attrs_init__` as original intended is that `attrs` does _not_ write an `__init__` at all and provides the user with with an `__attrs_init__` instead.

As I understand it now, your idea is: based on the `init` flag, we will either write `__init__` (as before), or instead name that method (minus the call to `__attrs_post_init__`) as `__attrs_init__`. That makes sense to me, and I will update accordingly.

I was doing something slightly different: always providing an `__attrs_init__` and a default `__init__` implementation that calls `__attrs_init__()` and `__attrs_post_init__()`, which the user can override -- thus removing the need for the `init` flag. But your version makes more sense, and I think it will be simpler, since I'm only renaming that function. 

Separately:

1) why do you think it is useful to have `__attrs_pre_init__()` + `__attrs_post_init__()` as well as  `__attrs_init__()`? It seems like these two have overlapping functionality - when do you anticipate someone using `__attrs_init__()` and their own `__init__()` instead of the pre/post init hooks? I can think of two situations: (a) they want `__init__` to take different arguments than their declared attr.ibs() would generate or (b) they need the arguments to do their pre-init work (and we won't pass arguments to `__attrs_pre_init__()`).

2) Currently `init=True` means an `__init__` is generated, and `init=False` means `__init__` is not. Can we treat `init=False` as indicating that they want an `__attrs_init__`?

indigoviolet:
This is now ready for review; not sure how to appease the codecov check, it became red after I added some docs.

indigoviolet:
@hynek just want to check that this is on your radar. no rush but if you happen to know when you might get to it, i won't bug you until then :)

hynek:
Yes it is, sorry. I need to get out a structlog release before I can focus on bigger attrs work.

indigoviolet:
ping?

src/attr/_make.py
```diff
@@ -1299,6 +1326,8 @@ def attrs(
     .. versionadded:: 20.1.0 *getstate_setstate*
     .. versionadded:: 20.1.0 *on_setattr*
     .. versionadded:: 20.3.0 *field_transformer*
+    .. versionchanged:: 20.4.0
```
hynek, line 1329:
```suggestion
    .. versionchanged:: 21.1.0
```

😇

src/attr/_make.py
```diff
@@ -1929,10 +1961,10 @@ def _make_init(
         unique_filename,
     )
 
-    __init__ = locs["__init__"]
-    __init__.__annotations__ = annotations
+    init = locs["__attrs_init__"] if attrs_init else locs["__init__"]
```
hynek, line 1964: IIRC correctly, the reason why we used `__init__` as a variable name, was such that `__name__` etc are correct? 🤔  But I don't think it works that way if I think about it…

src/attr/_make.py
```diff
@@ -2120,7 +2153,7 @@ def fmt_setter_with_converter(
         )
         arg_name = a.name.lstrip("_")
 
-        has_factory = isinstance(a.default, Factory)
+        has_factory = Factory is not None and isinstance(a.default, Factory)
```
hynek, line 2156: I don't understand this and why you set `Factory` globally to `None`. Could you explain please?

src/attr/_make.py
```diff
@@ -2277,7 +2310,7 @@ def fmt_setter_with_converter(
             names_for_globals[val_name] = a.validator
             names_for_globals[attr_name] = a
 
-    if post_init:
+    if post_init and not attrs_init:
```
hynek, line 2313: So this is a valid decision, but could you explain why you decided for it? I think I'd tend to leave it inside of `__attrs_init__` but am happy to be convinced otherwise.

src/attr/_make.py
```diff
@@ -2727,11 +2762,15 @@ def make_class(name, attrs, bases=(object,), **attributes_arguments):
         raise TypeError("attrs argument must be a dict or a list.")
 
     post_init = cls_dict.pop("__attrs_post_init__", None)
-    type_ = type(
-        name,
-        bases,
-        {} if post_init is None else {"__attrs_post_init__": post_init},
-    )
+    user_init = cls_dict.pop("__init__", None)
```
hynek, line 2765: Why do you remove it and put it back later? That seems like something that could break under certain conditions and seems ineffective?

hynek (changes requested):
Sorry for the delays and thank you for your patience! Overall this is all high quality, but we need to figure out some edge cases together.

src/attr/_make.py
indigoviolet: Ya, the retval from this function is assigned to the `_cls_dict`  in `_ClassBuilder`: https://github.com/indigoviolet/attrs/blob/vi-attr-init/src/attr/_make.py#L891, https://github.com/indigoviolet/attrs/blob/vi-attr-init/src/attr/_make.py#L910

The variable name we used for the retval doesn't affect things (the contents of `locs` would have already fixed `__name__` etc).

src/attr/_make.py
indigoviolet: I'm just following the pattern that was already present for `post_init`/`__attrs_post_init__`: https://github.com/indigoviolet/attrs/blob/vi-attr-init/src/attr/_make.py#L2764

It looks like we use the [`body` to construct `type_`](https://github.com/indigoviolet/attrs/blob/vi-attr-init/src/attr/_make.py#L2773), and then we pass the [`cls_dict` object to `_attrs`](https://github.com/indigoviolet/attrs/blob/vi-attr-init/src/attr/_make.py#L2797) -- I presume we can't just leave these methods in `cls_dict`. I haven't dug into this.

src/attr/_make.py
indigoviolet: As this code is written, if init=True, the user will end up writing an `__init__` like 

```
def __init__(self, *args, **kwargs):
     ... # pre-init stuff
     self.__attrs_init__(*args, **kwargs)
     self.__attrs_post_init__()
```
In theory this allows the user to do something in between `__attrs_init__` and `__attrs_post_init__`, but that doesn't make a lot of sense since they can just do that in `__attrs_post_init__`. It's probably worse to do it like this because users will forget to call post-init. 

I will change this back to be called by `__attrs__init__`.

src/attr/_make.py
indigoviolet: `Factory` is an `attrs` class, so it calls this code while it is being declared. If we don't forward-declare `Factory`, this will lead to some sort of undeclared variable error (I forget exactly what the error text was). 

However, we must obviously abide by the constraint of not using `factory=` inside the declaration of `Factory`, so that `has_factory` could never be True while `Factory` was being constructed; by the time a user class is being constructed, `Factory is not None`, so the code falls back to checking `isinstance(a.default, Factory)`.

Open to other suggestions here.

src/attr/_make.py
hynek: OK. I'm all for leaving doors open, but as you say: we don't lose anything here. And I already smell the confused bug reports about post init not being called…

src/attr/_make.py
hynek: Ah shit.

How about we pull this on Factory:

https://github.com/indigoviolet/attrs/blob/0effbc2852c6d22e6d351d10b56ed5bc12d001d8/src/attr/_make.py#L2546-L2568

The biggest downside of working on attrs is not being able to use attrs. 🤪

indigoviolet: Hmm- I'm still a bit confused by that - does Factory need to be frozen like `Attribute`? Does it need a get/set state, a cmp? Not sure I understand the consequences here completely.

hynek: I'm not saying you should copy paste. What I mean is that I'd like you to re-create Factory (as it is right now) using the `_add_XXX` primities like `Attribute`. Sorry for being fuzzy.

indigoviolet:
Yes, I understood that, but I'm not exactly sure which features used in the
Attribute class are needed to correctly create the Factory class. For
example, does it need to be frozen, does it need to support pickling etc.

src/attr/_make.py
indigoviolet: Yes, I understood that, but I'm not exactly sure which features used in the Attribute class are needed to correctly create the Factory class. For example, does it need to be frozen, does it need to support pickling etc.

hynek: I would say go for the same properties as now:

- slots
- hash
- eq
- repr

I think we can drop cmp; that makes no sense.

It could be frozen but that would probably break someone so please don't.

hynek: Maybe we should also add a test that its pickleable, if it doesn't already exist.

indigoviolet: ok, implemented your suggestion; lmk if I missed something. I think there's already a [test for pickling](https://github.com/python-attrs/attrs/blob/master/tests/test_functional.py#L272-L294).

indigoviolet: Looks like mypy tests are failing on master, see https://github.com/python-attrs/attrs/pull/748

hynek (approved):
Cool,thanks! This is an exciting feature!

We should add some docs to init.rst but that doesn't need to happen here.

hynek:
So, since you're knee deep in this code: do you think it would be feasible to also add `__attrs_pre_init__`?

I still think that's more useful for most people, that just want to call `super()` and not lose the typing information we deliver. 🤔

PR #744, opened 2020-12-28
RFC: Bring in the attrs plugin from mypy

euresti:
Note: The lint tests are failing because mypy code is 120 vs 79 here.  I plan to fix those but there are some bikesheds to decide first:

1. Is the incantation:
```
[mypy]
plugins = attr.mypy
```
what we want?  It's how https://github.com/samuelcolvin/pydantic/tree/master/pydantic ships its plugin and it seemed neater than how sqlalchemy-stubs does it.  (They ship a separate module but it looks like this)
```
[mypy]
plugins = sqlmypy
```

2. I know people have been clamoring to fix the "How do I make my method an attr.s method". Are we ok with landing this as is and then in a later revision adding that, possibly breaking whatever hacks people have already written?  i.e. do we need to worry about any backwards compatibility issues?
3. Do we wish to preserve any of the history from the mypy repo?  
4. Any licensing issues?

Tagging the following people since you contributed to the attrs plugin on mypy:
@ilevkivskyi @msullivan @JukkaL @Zac-HD @tkukushkin @Bogdanp
@natemcmaster @Michael0x2a @markusschmaus @ceh

Please let me know if this is a huge problem.  And let me know if you want me to somehow preserve the history of the file changes (or just your contribution).  This is completely new territory for me.  I'm not trying to just steal people's hard work. I'm just trying to make it easier for attrs to maintain the plugin outside of the regular mypy release cycle.

# Pull Request Check List

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -12,6 +12,9 @@ def pytest_configure(config):
     )
     settings.load_profile("patience")
 
+    # Use this mypy file for the tests.
+    config.option.mypy_ini_file = "mypy-pytest.ini"
+
 
 collect_ignore = []
 if sys.version_info[:2] < (3, 6):
--- a/mypy-pytest.ini
+++ b/mypy-pytest.ini
@@ -1,0 +1,2 @@
+[mypy]
+plugins = attr.mypy
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,3 +1,4 @@
 [mypy]
 disallow_untyped_defs = True
 check_untyped_defs = True
+plugins = attr.mypy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,8 @@ exclude_lines = [
 [tool.black]
 line-length = 79
 
+[tool.pytest.ini_options]
+addopts = "--mypy-ini-file=mypy-pytest.ini"
 
 [tool.interrogate]
 verbose = 2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,8 +23,6 @@ exclude_lines = [
 [tool.black]
 line-length = 79
 
-[tool.pytest.ini_options]
-addopts = "--mypy-ini-file=mypy-pytest.ini"
 
 [tool.interrogate]
 verbose = 2
--- a/src/attr/mypy.py
+++ b/src/attr/mypy.py
@@ -1,0 +1,21 @@
+from typing import Callable, Optional
+
+from mypy.plugin import ClassDefContext, Plugin
+from mypy.typeops import TypingType
+
+
+def plugin(version: str) -> "TypingType[Plugin]":
+    """
+    `version` is the mypy version string
+
+    We might want to use this to print a warning if the mypy version
+    being used is newer, or especially older, than we expect (or need).
+    """
+    return AttrsPlugin
+
+
+class AttrsPlugin(Plugin):
+    def get_class_decorator_hook(
+        self, fullname: str
+    ) -> Optional[Callable[[ClassDefContext], None]]:
+        return None
--- a/src/attr/mypy.py
+++ b/src/attr/mypy.py
@@ -1,7 +1,910 @@
-from typing import Callable, Optional
+"""Plugin for supporting the attrs library (http://www.attrs.org)"""
 
-from mypy.plugin import ClassDefContext, Plugin
-from mypy.typeops import TypingType
+from functools import partial
+from typing import Callable, Dict, Iterable, List, Optional, Tuple, cast
+
+import mypy.plugin  # To avoid circular imports.
+
+from mypy.exprtotype import TypeTranslationError, expr_to_unanalyzed_type
+from mypy.fixup import lookup_qualified_stnode
+from mypy.nodes import (
+    ARG_NAMED,
+    ARG_NAMED_OPT,
+    ARG_OPT,
+    ARG_POS,
+    MDEF,
+    Argument,
+    AssignmentStmt,
+    CallExpr,
+    Context,
+    Decorator,
+    Expression,
+    FuncDef,
+    JsonDict,
+    ListExpr,
+    MemberExpr,
+    NameExpr,
+    OverloadedFuncDef,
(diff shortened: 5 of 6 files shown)


Tinche:
> I know people have been clamoring to fix the "How do I make my method an attr.s method". Are we ok with landing this as is and then in a later revision adding that, possibly breaking whatever hacks people have already written? i.e. do we need to worry about any backwards compatibility issues?

As one of the people clamoring for this I think taking a little while to get this right is more important than backwards comp right now.

hynek:
> I know people have been clamoring to fix the "How do I make my method an attr.s method". Are we ok with landing this as is and then in a later revision adding that, possibly breaking whatever hacks people have already written? i.e. do we need to worry about any backwards compatibility issues?

Unfortunately I'm busy until next year, but could you elaborate what this means please? Would embracing this break my attrs-ng plugin?

(I like attr.mypy better I think)

euresti:
> Would embracing this break my attrs-ng plugin?
Yes and no.  Your plugin currently does this:

```
from mypy.plugin import Plugin
from mypy.plugins.attrs import attr_attrib_makers, attr_dataclass_makers

attr_dataclass_makers.add("attr.define")
attr_dataclass_makers.add("attr.mutable")
attr_dataclass_makers.add("attr.frozen")
attr_attrib_makers.add("attr.field")
```

Which will pretty much do nothing because it's trying to patch `mypy.plugins.attrs`
However the plugin in `attr.mypy` actually supports attrs-NG so there's no need for this.  

But what I'm talking about is the `attr_dataclass_makers.add("attr.define")` itself.  Maybe we change it so that you can configure it using mypy.ini config file or some other thing like `register_attr_dataclass_maker(...)` or something like that.  How important is it to preserve this "hack" for backwards compatibility.  (And do we even want to answer this right now?)

hynek:
I really don't want to break anything and that would actually break almost all of my work projects. :D

But let's back up a little. If I understand it correctly, people will have to opt **in** to use our plugin and otherwise they'll fall back to mypy, is that correct? That would mean that we wouldn't break anything and people would have to touch their mypy.ini anyways, correct?

I mean mid-term we could ask mypy to automatically recognize _our_ plugin, but ISTM we have some leeway thanks to this?

euresti:
> I really don't want to break anything and that would actually break almost all of my work projects. :D

I think I lost the meaning of "that" in this sentence.  Could you elaborate what would break almost all your work projects?

> But let's back up a little. If I understand it correctly, people will have to opt in to use our plugin and otherwise they'll fall back to mypy, is that correct? That would mean that we wouldn't break anything and people would have to touch their mypy.ini anyways, correct?

We could also remove the plugin from mypy.  In fact we should otherwise we'll get version skew bug reports which would be really annoying.  In mypy we could error out when it sees an attr.s and tell you to install the plugin.  

> I mean mid-term we could ask mypy to automatically recognize our plugin, but ISTM we have some leeway thanks to this?

Are you suggesting that something in mypy detect that the code is using attrs and load our plugin?   This could be possible but I think it might have a lot of weird issues because it changes when our plugin gets intialized by the mypy plugin system and who knows what else.

Is having to modify mypy.ini a deal breaker?

JukkaL:
I don't have a strong opinion on this, but I have a few concerns:

* The plugin uses a bunch of undocumented, internal mypy APIs that are not necessarily stable. It's possible that updating mypy could break the attrs plugin (at least older versions of the plugin, if we are careful to keep attrs and mypy in sync). This is my main concern. Maybe we'd continue to test the plugin in mypy CI to catch regressions -- but this would effectively turn these APIs into stable APIs.
* It may become unclear which mypy versions are supported by the plugin, and the plugin may accidentally break on older mypy versions that are still in active use. It would be nice if attrs CI could test the plugin using the oldest supported mypy version and the latest one. If somebody tries to use the plugin on an unsupported mypy version, the plugin probably shouldn't crash but suggest updating to a more recent mypy version.
* As proposed above, mypy should start suggesting to install attrs and enabling the plugin, as otherwise users could silently lose some type checking precision. In case a mypy user doesn't want to use the plugin shipped with attrs for some reason, we can make it possible to ignore the error by giving it a unique mypy error code (this needs to be documented, however).

hynek:
Hey Jukka, thanks for chiming in!

> Maybe we'd continue to test the plugin in mypy CI to catch regressions -- but this would effectively turn these APIs into stable APIs.

Correct me if I'm wrong, but aren't they de-facto stable already? How are pydantic and sqlalchemy stubs handle this?

> It would be nice if attrs CI could test the plugin using the oldest supported mypy version and the latest one. If somebody tries to use the plugin on an unsupported mypy version, the plugin probably shouldn't crash but suggest updating to a more recent mypy version.

That seems reasonable.

> As proposed above, mypy should start suggesting to install attrs and enabling the plugin, as otherwise users could silently lose some type checking precision. In case a mypy user doesn't want to use the plugin shipped with attrs for some reason, we can make it possible to ignore the error by giving it a unique mypy error code (this needs to be documented, however).

Long-term I would really like to make it opt-out, not opt-in. I.e. mypy picks up attrs's plugin automatically and there should be a switch to disable that – would you find that reasonable?

hynek:
> > I really don't want to break anything and that would actually break almost all of my work projects. :D
> I think I lost the meaning of "that" in this sentence. Could you elaborate what would break almost all your work projects?

Not having _some_ attrs plugin active by default.

> > I mean mid-term we could ask mypy to automatically recognize our plugin, but ISTM we have some leeway thanks to this?
> 
> Are you suggesting that something in mypy detect that the code is using attrs and load our plugin? This could be possible but I think it might have a lot of weird issues because it changes when our plugin gets intialized by the mypy plugin system and who knows what else.

Yeah, that's what I'm suggesting. If it's not possible, so be it but it would be really nice if we could make that work.

> Is having to modify mypy.ini a deal breaker?

Not a deal breaker per se, but it's definitely a usability regression (especially over dataclasses) that I would like to get rid of eventually. Also it's breaking backward-compat.

PR #760, opened 2021-02-16
Fix issue with get_type_hints(cls.__init__) and refactor

### LINKED ISSUE #593: Deferred type annotations are evaluated in the wrong execution context
To reproduce, run the following program:
```python
import attr
from typing import List, get_type_hints

@attr.s
class C:
    x = attr.ib(type='List[int]')

print(get_type_hints(C.__init__))
```

#### Expected result
```
{'return': <class 'NoneType'>, 'x': typing.List[int]}
```

#### Actual result
```
  File "example.py", line 8, in <module>
    print(get_type_hints(C.__init__))
  File "C:\Python37\lib\typing.py", line 1001, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "C:\Python37\lib\typing.py", line 260, in _eval_type
    return t._evaluate(globalns, localns)
  File "C:\Python37\lib\typing.py", line 464, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name 'List' is not defined
```

This annotation should be evaluated in the context of the module where it's defined, where `List` is in globals.

#### Version info
```
attrs 19.3.0
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
```

euresti:
This fixes #593 

I factored out the call to eval(compile... and added a generic `_make_method` method to remove repeated code.

# Pull Request Check List

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -294,14 +294,11 @@ def _compile_and_eval(script, globs, locs=None, filename=""):
     eval(bytecode, globs, locs)
 
 
-def _make_method(name, script, filename, globs=None, locs=None):
+def _make_method(name, script, filename, globs=None):
     """
-    Create the method with the script give and return the method object.
-
-    Note: This will mutate locals.
+    Create the method with the script given and return the method object.
     """
-    if locs is None:
-        locs = {}
+    locs = {}
     if globs is None:
         globs = {}
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1951,6 +1951,10 @@ def _make_init(
     )
     locs = {}
     bytecode = compile(script, unique_filename, "exec")
+    if cls.__module__ in sys.modules:
+        # This makes typing.get_type_hints(CLS.__init__) resolve string types.
+        globs.update(sys.modules[cls.__module__].__dict__)
+
     globs.update({"NOTHING": NOTHING, "attr_dict": attr_dict})
 
     if needs_cached_setattr:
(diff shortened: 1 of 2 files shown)


hynek:
We totally can switch to exec.

It was only removed by your boss in https://github.com/python-attrs/attrs/pull/87 because it’s a bit ugly re: Python 2.

If it helps you in any way: feel free to re-introduce it!

euresti:
After reading a bunch of docs I think we can leave well enough alone and keep going with compile/eval.  Unless @glyph has strong opinions about it now.  

So do we think having the correct "globals" on every method is necessary?  If so I can factor out the eval/compile into a helper.

hynek:
I don’t think that any other method changes it’s signature based on the attributes.  However, I think I’d prefer a nice and documented function anyways if it’s not too much trouble!

euresti:
Right. The only other uses of compile/eval are are hash, the cmp methods and the Attributes class creator, which don't have annotations. 

I'll try to make a helper anyway just to keep them together.

euresti:
I'm not sure why it's saying the tests weren't successful when they were here:

image

hynek:
So since codecov is once again giving us a hard time, I've run it manually and got this missing coverage:

```
src/attr/_make.py             911      0    463      2    99%   303->305, 1957->1961
```

Any chance of fixing it? 😇

euresti:
My pleasure.

hynek:
Thank you so much!

PR #763, opened 2021-02-19
Add Python 3.10 support

### LINKED ISSUE #716: Python 3.10 compatibility
Hello.

We have already started testing RPM packages in Fedora with Python 3.10a2 and attrs is currently not compatible.

The problem is caused by the implementation of [PEP 563](https://www.python.org/dev/peps/pep-0563/#abstract) — type annotations are now stored as strings which cause asserts like this one to fail:

```python
class TestAnnotations:
    """
    Tests for types derived from variable annotations (PEP-526).
    """

    def test_basic_annotations(self):
        """
        Sets the `Attribute.type` attr from basic type annotations.
        """

        @attr.s
        class C:
            x: int = attr.ib()
            y = attr.ib(type=str)
            z = attr.ib()

        assert int is attr.fields(C).x.type
```

The value stored in `attr.fields(C).x.type` is a string "int" instead of a class int.

It seems it was discussed before in https://github.com/python-attrs/attrs/issues/424

hynek:
Python 3.10 has string types everywhere and that has unmasked a bunch of
bugs/edge cases in our code.

Especially the hooks code need a resolving helper for string types. I'm leaving
that for a separate PR.

Same for test_mypy.yml

Fixes #716

cc @euresti @ambv @frenzymadness @sscherfke

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -386,9 +386,11 @@ def _is_class_var(annot):
     """
     annot = str(annot)
 
-    return annot.startswith(_classvar_prefixes) or annot[1:].startswith(
-        _classvar_prefixes
-    )
+    # Annotation can be quoted.
+    if annot.startswith(("'", '"')) and annot.endswith(("'", '"')):
+        annot = annot[1:-1]
+
+    return annot.startswith(_classvar_prefixes)
 
 
 def _has_own_attribute(cls, attrib_name):
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -384,7 +384,11 @@ def _is_class_var(annot):
     annotations which would put attrs-based classes at a performance
     disadvantage compared to plain old classes.
     """
-    return str(annot).startswith(_classvar_prefixes)
+    annot = str(annot)
+
+    return annot.startswith(_classvar_prefixes) or annot[1:].startswith(
+        _classvar_prefixes
+    )
 
 
 def _has_own_attribute(cls, attrib_name):
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -25,13 +25,12 @@ class TestAnnotations:
         Sets the `Attribute.type` attr from basic type annotations.
         """
 
+        @attr.resolve_types
         @attr.s
         class C:
             x: int = attr.ib()
             y = attr.ib(type=str)
             z = attr.ib()
-
-        attr.resolve_types(C)
 
         assert int is attr.fields(C).x.type
         assert str is attr.fields(C).y.type
@@ -61,12 +60,11 @@ class TestAnnotations:
         Sets the `Attribute.type` attr from typing annotations.
         """
 
+        @attr.resolve_types
         @attr.s
         class C:
             x: typing.List[int] = attr.ib()
             y = attr.ib(type=typing.Optional[str])
-
-        attr.resolve_types(C)
 
         assert typing.List[int] is attr.fields(C).x.type
         assert typing.Optional[str] is attr.fields(C).y.type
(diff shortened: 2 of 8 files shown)


tests/test_annotations.py
```diff
@@ -382,30 +402,41 @@ def noop():
...
+        sys.version_info[:2] == (3, 6), reason="Does not work on 3.6."
+    )
     @pytest.mark.parametrize("slots", [True, False])
-    @pytest.mark.parametrize("classvar", _classvar_prefixes)
-    def test_annotations_strings(self, slots, classvar):
+    def test_annotations_strings(self, slots):
         """
         String annotations are passed into __init__ as is.
+
+        It fails on 3.6 due to a bug in Python.
```
euresti, line 413: What bug?

src/attr/_make.py
```diff
@@ -366,7 +366,11 @@ def _is_class_var(annot):
     annotations which would put attrs-based classes at a performance
     disadvantage compared to plain old classes.
     """
-    return str(annot).startswith(_classvar_prefixes)
+    annot = str(annot)
```
euresti, line 369: I'd prefer something safer here like:
```
    if annotation.startswith(("'", '"')) and annotation.endswith(("'", '"')):
        annotation = annotation[1:-1]
```
To handle the "double quoted" annotations.

tests/test_annotations.py
```diff
@@ -31,14 +31,16 @@ class C:
             y = attr.ib(type=str)
             z = attr.ib()
 
+        attr.resolve_types(C)
```
euresti, line 34: You can also use resolve_types as a decorator which might make it easier to read.  (But you can't use the decorator if there are forward references)

tests/test_annotations.py
```diff
@@ -487,10 +518,6 @@ class C:
             y = attr.ib(type=str)
             z = attr.ib()
 
-        assert "int" == attr.fields(C).x.type
```
euresti, line 490: Why did you remove these?

tests/test_dunders.py
```diff
@@ -624,9 +624,8 @@ def test_init(self, slots, frozen):
         with pytest.raises(TypeError) as e:
             C(a=1, b=2)
 
-        assert (
+        assert e.value.args[0].endswith(
```
euresti, line 627: Interesting. What does it start with now?

tests/test_annotations.py
```diff
@@ -31,14 +31,16 @@ class C:
...
         assert int is attr.fields(C).x.type
         assert str is attr.fields(C).y.type
         assert None is attr.fields(C).z.type
-        assert C.__init__.__annotations__ == {
+        assert {
             "x": int,
             "y": str,
-            "return": None,
-        }
+            "return": type(None),
```
euresti, line 42: I would probably define a constant called `NoneType = type(None)`. Or I might define a helper that tests this. 

```
def assert_init_annotation(cls, **annotations):
    annotations["return"] = type(None)
    assert annotations == typing.get_type_hints(cls.__init__)

assert_init_annotation(C, x=int, y=str)
```

tests/test_annotations.py
hynek: `TypeError: Forward references must evaluate to types. Got typing.ClassVar[int].`

src/attr/_make.py
hynek: Agreed. I think I was just tired/frustrated.

tests/test_annotations.py
hynek: Because they don't work across Python versions (they become typing.ForwardRef or something like that). I figured it's more important it ultimately gives us what we need?

tests/test_dunders.py
hynek: The class names. So in this case `C.`.

tests/test_annotations.py
hynek: Excellent idea!

hynek:
The tests have totally finished and succeeded. _sigh_

euresti (approved):
Changes look good!

PR #768, opened 2021-02-25
Document comparison

hynek:
These are some narrative docs for #627, I'm open to suggestions to make it juicier.

Here's the built version: https://attrs--768.org.readthedocs.build/en/768/comparison.html

@Julian @botant

### DIFF
--- a/docs/comparison.rst
+++ b/docs/comparison.rst
@@ -42,4 +42,4 @@ It is then used as a key function like you may know from `sorted`:
 This is especially useful when you have fields with objects that have atypical comparison properties.
 Common examples of such objects are `NumPy arrays <https://github.com/python-attrs/attrs/issues/435>`_.
 
-Please note that they are set *independently* because *order* is `False` by default in `modern APIs <prov>`.
+Please note that *eq* and *order* are set *independently*, because *order* is `False` by default in `modern APIs <prov>`.
--- a/docs/comparison.rst
+++ b/docs/comparison.rst
@@ -1,7 +1,7 @@
 Comparison
 ==========
 
-By default, two instances of ``attrs`` classes is equal, if all their fields are equal.
+By default, two instances of ``attrs`` classes are equal, if all their fields are equal.
 For that, ``attrs`` writes ``__eq__`` and ``__ne__`` methods for you.
 
 Additionally, if you pass ``order=True`` (default if you use the `attr.s` decorator), ``attrs`` will also create a full set of ordering methods that are based on the defined attributes: ``__le__``, ``__lt__``, ``__ge__``, and ``__gt__``.
--- a/docs/comparison.rst
+++ b/docs/comparison.rst
@@ -22,12 +22,17 @@ As with other features, you can exclude attributes from being involved in compar
    >>> C(1, 2) == C(1, 3)
    True
 
-Additionally you can also pass a *callable* instead of a bool to both ``eq`` and ``order``
+Additionally you can also pass a *callable* instead of a bool to both *eq* and *order*.
 It is then used as a key function like you may know from `sorted`:
 
 .. doctest::
 
    >>> import attr
+   >>> @attr.s
+   ... class S(object):
+   ...     x = attr.ib(eq=str.lower)
+   >>> S("foo") == S("FOO")
+   True
    >>> @attr.s
    ... class C(object):
    ...     x = attr.ib(order=int)
--- a/docs/comparison.rst
+++ b/docs/comparison.rst
@@ -22,8 +22,10 @@ As with other features, you can exclude attributes from being involved in compar
    >>> C(1, 2) == C(1, 3)
    True
 
-Additionally you can also pass a *callable* instead of a bool to both ``eq`` and ``order``.
+Additionally you can also pass a *callable* instead of a bool to both ``eq`` and ``order``
 It is then used as a key function like you may know from `sorted`:
+
+.. doctest::
 
    >>> import attr
    >>> @attr.s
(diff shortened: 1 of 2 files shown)


botant:
Looks good to me, simple and direct.

1) There is a typo in the first sentence:
>> By default, two instances of attrs classes **is** equal, if all their fields are equal. For that, attrs writes __eq__ and __ne__ methods for you.

2) I'd add a case insensitive comparison, since this was one of the requests.

3) I'd mention pandas *DataFrame*s as well as numpy *array*s. Is there a way to provide an example without having to import those libraries?

4) I'd make it clear that specifying a *callable* for `eq` does not mean the same key will be used for `order`; the key must be specified explicitly.

hynek:
I believe I have addressed everything except 3. I'd be happy to add a fake doctest but someone else has to write it for me because I've never used neither NumPy nor Pandas. 😳

botant:
I just realized that we were going to provide a wrapping class, since numpy and pandas do not follow the key model. In both cases, you need to use a comparison function that takes 2 arguments. This was discussed in the original PR. 

Shall we provide a function like `attrs.cmp.using(key: Callable) -> Callable`? The output `callable` is a function that returns a wrapper object that uses `key` as `eq`. Or something like that. 

Alternatively, and probably easier, we can *not* mention numpy in this release, and add it back via a different PR.

hynek:
Hm, I think it would be nicer to publish all at once if it isn't too much work. We had decided on the API IIRC and it shouldn't be too complex to add? 🤔 Just adding the APIs that landed without NumPy support seems to be lacking the main effect? :)

botant:
That works as well.  I'll work on it later today, if that is okay?

botant:
There is another point on `cmp`, and that might be what you meant to ask in the other PR.

Should we de-deprecate it and keep it permanently as syntactic sugar? That would save users from having to wirte `attr.ib(eq=int, order=int)` for a nicer looking `attr.ib(cmp=int)`.

Just a thought.

hynek:
> That works as well. I'll work on it later today, if that is okay?

It's not like I can cut your salary. ;) If we get it done by tomorrow it's be more than great. RedHat wants a release but there's some minor stuff I need to address first anyways.

> Should we de-deprecate it and keep it permanently as syntactic sugar?

Yes we should. It was very obvious to me while writing the docs. :)

botant:
How about something along these lines?  The same logic used to validate `cmp`, `eq` and `order` in `attr.ib()` would be used here to regulate the usage of the arguments.

`attr.cmp.using(eq: Callable = None, order: Callable = None, cmp: Callable = None, enforce_same_class: bool = True) -> Callable`

```python
def using(eq: Callable = None, order: Callable = None, cmp: Callable = None, enforce_same_class: bool = True) -> Callable:

    @attr.s(eq=False, order=False)
    class Comparable(object):
        value = attr.ib()
        def __eq__(self, other):
            pass
        def __lt__(self, other):
            pass
        ...

    return Comparable
```

Note: I think we should let the user choose whether equal classes should be enforced, since the whole point of the PR is to let the user customize the operation.

hynek:
If everything is done by one function, what's the point of the module? Storing helpers?

botant:
No particular reason, I just thought it looked good.

hynek:
I'm not opposed to it, I'd just like to think it through. I could live with a helper for Numpy/Pandas for instance, if it's possible without importing them.

botant:
I wouldn't add specific helpers for numpy/pandas.  Too much trouble and too little added value, if all the user needs is `using(key=np.array_equal)` or something similar.  Plus, I'm afraid that one particular helper might not cover all use cases related to matrices, time series, etc.  I think one simple function is the way to go.

docs/comparison.rst
```diff
@@ -0,0 +1,45 @@
+Comparison
+==========
+
+By default, two instances of ``attrs`` classes are equal, if all their fields are equal.
```
Julian, line 4:
```suggestion
By default, two instances of ``attrs`` classes are equal if all their fields are equal.
```

docs/comparison.rst
```diff
@@ -0,0 +1,45 @@
...
+   ...     x = attr.ib(eq=str.lower)
+   >>> S("foo") == S("FOO")
+   True
+   >>> @attr.s
+   ... class C(object):
+   ...     x = attr.ib(order=int)
+   >>> C("10") > C("2")
+   True
+
+This is especially useful when you have fields with objects that have untypical comparison properties.
```
Julian, line 42:
```suggestion
This is especially useful when you have fields with objects that have atypical comparison properties.
```

docs/comparison.rst
```diff
@@ -0,0 +1,45 @@
...
+A common example being `NumPy arrays <https://github.com/python-attrs/attrs/issues/435>`_.
```
Julian, line 43:
```suggestion
Common examples of such objects are `NumPy arrays <https://github.com/python-attrs/attrs/issues/435>`_.
```

docs/comparison.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Please note that they are set *independently* because *order* is `False` by default in `modern APIs <prov>`.
```
Julian, line 45: I'm not sure I understand this sentence -- maybe it's just the word `they`? Maybe worth rewriting in slightly more words to clarify what thing is referred to, actually now rereading a fourth time this means to say that `order` and `eq` are independently controlled?

Julian (approved):
Cool!

docs/comparison.rst
hynek: Yes, the sentence used to be higher up. I'll clarify it but it will change a bit anyways, once botant comes through with the cmp de-deprecation. :)

hynek:
> Maybe `attr.cmp_using`?

Oooof sorry I lost the thread on this one! Yes please, I’d like to push 21.1.0 but realized this is still open. Any chance to get it done sometime next week?

botant:
Working on this now.

PR #770, opened 2021-02-25
Fix make_class for dynamic generic classes

### LINKED ISSUE #756: attr.make_class does not work with Python>=3.7
In attr.make_class function line 2818: `type_ = type(name, bases, body)` cannot be used for dynamic generic class creation beyond python 3.7 and leads to an error `TypeError: type() doesn't support MRO entry resolution; use types.new_class()`.

The error can be recreated with:
```
from typing import TypeVar, Generic
from types import new_class
import attr

MyTypeVar = TypeVar("MyTypeVar")
MyParent = new_class("MyParent", (Generic[MyTypeVar],), {})


attr.make_class('test', {'id': attr.ib(type=str)}, (MyParent[int],))
```

As mentioned in the error it can be fixed by replacing `type_ = type(name, bases, body)` with `types.new_class(name, bases, body)`. 

This issue has been discussed here in detail: [https://bugs.python.org/issue33188](https://bugs.python.org/issue33188).

hynek:
Fixes #756

Does this look OK to you @avinashmnit30?

### DIFF
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -28,14 +28,13 @@ if PY2:
     def isclass(klass):
         return isinstance(klass, (type, types.ClassType))
 
+    new_class = type
+
     # TYPE is used in exceptions, repr(int) is different on Python 2 and 3.
     TYPE = "type"
 
     def iteritems(d):
         return d.iteritems()
-
-    def new_class(name, bases, body):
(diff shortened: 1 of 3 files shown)


hynek:
Yuk, looks like it's more complicated than assumed.

Issue #766, opened 2021-02-25
Field hooks are too clunky with Python 3.10 / string annotations


hynek:
Currently I had to exclude `test_hooks.py` on Python 3.10, because it would be too painful to make them work with string annotations.

The problem is that there is no good helper to transform `"str"` into `str`. We're supposed to be able to `eval()` the string.

I've talked to @ambv and he suggested the following:

> If faced with a string, first do `eval(str, globalns, localns)`. Then on any type you get, do `typing._eval_type(type, globalns, localns)` which will deal with recursed ForwardRefs.

e.g.

```pycon
>>> class G(Generic[T]):
...   def method(self) -> 'List["T"]':
...     return 1
...
>>> G.method.__annotations__
{'return': 'List["T"]'}
>>> eval(G.method.__annotations__['return'], globals(), locals())
typing.List[ForwardRef('T')]
>>> return_type=_
>>> typing._eval_type(return_type, globals(), locals())
typing.List[~T]
```

`typing._eval_type` has been part of the stdlib since 3.5 and I don't expect it to go anywhere.

---

I'm not sure how to call that function. `attr.resolve_type()` sounds good but that's a bit too close to `attr.resolve_types()` so I'm open to better suggestions.

---

I'm not sure we can get it done by 21.1.0 but it should be on our short term agenda.

cc @sscherfke @euresti

euresti:
Another option is to do
```
            import typing
            hints = typing.get_type_hints(cls)
```
The hints will have the evalled rhs of the colon

example: `{'x': <class 'int'>, 'y': <class 'str'>}`

You can't call attr.resolve_types because `cls` is not an attrs class yet.

hynek:
What I mean is this:

```
>>> import typing
>>> typing.get_type_hints("str")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/hynek/.asdf/installs/python/3.7.9/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 1005, in get_type_hints
    'or function.'.format(obj))
TypeError: 'str' is not a module, class, method, or function.
```

There is just no good way for this.

euresti:
I meant do something like this:
```
         def hook(cls, attribs):
            import typing
            hints = typing.get_type_hints(cls)
            results[:] = [(a.name, hints.get(a.name, a.type)) for a in attribs]
            return attribs
```

hynek:
Ah right, you're correct as usual.

[...]

PR #772, opened 2021-02-27
Fix test_hooks for Python 3.10 and str annotations

hynek:
This makes all tests pass but it shows how clunky the hooks become, once string annotations comes into play. So I wouldn't say it's a resolution for #766, but just a step towards clarity.


hynek:
better fixed by #774

PR #773, opened 2021-02-27
Undeprecate cmp

botant:
# Pull Request Check List

- [x] Removed **tests** for cmp deprecation warning.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -17,6 +17,7 @@ from ._compat import (
     isclass,
     iteritems,
     metadata_proxy,
+    new_class,
     ordered_dict,
     set_closure_cell,
 )
@@ -2916,7 +2917,8 @@ def make_class(name, attrs, bases=(object,), **attributes_arguments):
     if user_init is not None:
         body["__init__"] = user_init
 
-    type_ = type(name, bases, body)
+    type_ = new_class(name, bases, {}, lambda ns: ns.update(body))
+
     # For pickling to work, the __module__ variable needs to be set to the
     # frame where the class is created.  Bypass this step in environments where
     # sys._getframe is not defined (Jython for example) or sys._getframe is not
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -195,7 +195,6 @@ def attrib(
 
     :param cmp: Setting to ``True`` is equivalent to setting ``eq=True,
         order=True``. Can also be set to a ``callable``.
-        Deprecated in favor of *eq* and *order*.
     :type cmp: a `bool` or a `callable`.
 
     :param Optional[bool] hash: Include this attribute in the generated
@@ -251,7 +250,8 @@ def attrib(
     .. versionadded:: 19.2.0 *eq* and *order*
     .. versionadded:: 20.1.0 *on_setattr*
     .. versionchanged:: 20.3.0 *kw_only* backported to Python 2
-    .. versionchanged:: 21.1.0 *eq* and *order* also accept a custom callable.
+    .. versionchanged:: 21.1.0 *eq* and *order* also accept a custom callable
+    .. versionchanged:: 21.1.0 *cmp* undeprecated
     """
     eq, eq_key, order, order_key = _determine_attrib_eq_order(
         cmp, eq, order, True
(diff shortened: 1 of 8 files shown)


botant:
Not sure which changelog file I should add. I thought it was 773, but there is already a 787.

src/attr/_make.py
```diff
@@ -2605,10 +2595,8 @@ def from_counting_attr(cls, name, ca, type=None):
     @property
     def cmp(self):
         """
-        Simulate the presence of a cmp attribute and warn.
+        Simulate the presence of a cmp attribute.
         """
-        warnings.warn(_CMP_DEPRECATION, DeprecationWarning, stacklevel=2)
```
hynek, line 2610: I've been thinking and I think we should keep this deprecation.

The behavior of cmp is rather muddy, because it's a derivation of eq and order.

Now cmp becomes the officially blessed way of setting eq and order but this attribute doesn't make sense to me.

Especially because it returns a bool, while eq and order can be callables.

Does that make sense to you?

botant: I can see where you're coming from.  We have three arguments: `eq`, `order` and `cmp`, so one of them is redundant for sure.  At the same time, `eq` and `order` aren't entirely *clean*, because `order` requires `eq`.

My preference would be to undeprecate `cmp` to avoid `attr.ib(eq='str.lower', order='eq.lower')`. I think `attr.ib(cmp='eq.lower')` feels and looks a bit nicer.

For the record, in my applications I will use `attr.ib(eq=key_func(... ), order=False)`. Very often `order` won't make sense with complex objects. In fact, I'm a lot more annoyed by `order` being `True` by default than I am by `cmp` vs `eq`.

Having said that, I completely understand that `cmp` deprecation has been on the cards for a long time, and that undeprecating it may feel like a regression.  Happy to follow your lead.

hynek: I think/hope we've misunderstood each other!

What I mean is that we should undeprecate cmp for `attr.s`/`attr.attrs` (and add it to `attr.define` for that matter, but that's a different PR) since it's syntactic sugar for setting eq/order and very useful.

---

But I want to remove it eventually from the Attribute class. That's why I've commented specifically on the change in the cmp property.

botant: Oh I see, you want to keep the deprecation specifically in the `cmp` property of `Attribute` - not else where? Sorry, my bad.  Would we keep it in `Attribute`'s `init`?

> Especially because it returns a bool, while eq and order can be callables.

`eq` and `order` are also booleans. The callables are stored under `eq_key` and `order_key`.

botant: Do we need a new deprecation text?

```python
_CMP_DEPRECATION = (
    "The usage of `cmp` is deprecated and will be removed on or after "
    "2021-06-01.  Please use `eq` and `order` instead."
)
```

hynek: If it's not used, probably not?

hynek: > Would we keep it in Attribute's init?

You mean the deprecation? Yes. I would like to remove cmp from Attribute for good eventually. I want it only to be a syntactic sugar, nothing more.

hynek:
> Not sure which changelog file I should add. I thought it was 773, but there is already a 787.

I'm confused, what is about 787? :) Your correct issue number is 773 so make it `773.deprecation.rst`

Don't spend too much time on it, I will edit it anyways because I need to add some apologies.

src/attr/_make.py
botant: All done (I think). Apologies for the confusion.

botant:
> I'm confused, what is about 787? :) Your correct issue number is 773 so make it `773.deprecation.rst`

Done.

botant:
Out of curiosity, what is keeping us from removing `cmp` from `Attribute`'s `init`? As far as I understand, attributes are not supposed to be instantiated manually, and `cmp` is already not used there anyway.

How about the property? Is it even possible to quantity the amount of damage that would be caused if we removed it?

Just curious to understand how you manage breaking changes.

hynek:
> Out of curiosity, what is keeping us from removing `cmp` from `Attribute`'s `init`? As far as I understand, attributes are not supposed to be instantiated manually, and `cmp` is already not used there anyway.

Yes, but people have historically done so nevertheless. :(

> How about the property? Is it even possible to quantity the amount of damage that would be caused if we removed it?

It's not! Which is why SemVer is bullshit (blog post incoming) but I'm trying to be extra careful and diligent.

They're both gonna move on to a better place in June.

> Just curious to understand how you manage breaking changes.

Poorly, but I'm trying my best. :)

hynek (approved):

hynek:
Thanks!

botant:
I'll submit another PR with the helper function soon.

PR #780, opened 2021-03-11
Narrower typing on Factory called with takes_self

dargueta:
This changes the annotations for `attr.Factory` so that the acceptable type of the callable changes based on `takes_self`.
Theoretically, MyPy will let you do this:

```py
attr.ib(default=attr.Factory(list, takes_self=True)
attr.ib(default=attr.Factory((lambda self: []), takes_self=False)
```

The change in annotations will prevent that from happening.

As written I *think* these annotations would require the type checker to be running on Python 3.8 because of the use of `Literal`. There are a couple ways around this, if necessary.

# Pull Request Check List

### DIFF
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -4,6 +4,7 @@ from typing import (
     Dict,
     Generic,
     List,
+    Literal,
     Mapping,
     Optional,
     Sequence,
@@ -64,9 +65,15 @@ NOTHING: object
 def Factory(factory: Callable[[], _T]) -> _T: ...
 @overload
 def Factory(
-    factory: Union[Callable[[Any], _T], Callable[[], _T]],
-    takes_self: bool = ...,
+    factory: Callable[[Any], _T],
+    takes_self: Literal[True],
 ) -> _T: ...
+@overload
+def Factory(
+    factory: Callable[[], _T],
+    takes_self: Literal[False],
+) -> _T: ...
+
 
 class Attribute(Generic[_T]):
     name: str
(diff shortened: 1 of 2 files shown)


dargueta:
Looks like it's blowing up because of https://github.com/python/typed_ast/issues/156 which was only fixed four days ago in https://github.com/python/typed_ast/pull/155. There's been no release yet, so for the moment MyPy is broken on CPython 3.10.0a6 and later. I think using 3.10.0a5 or lower might fix the issue?

src/attr/__init__.pyi
```diff
@@ -64,9 +65,15 @@ NOTHING: object
 def Factory(factory: Callable[[], _T]) -> _T: ...
 @overload
 def Factory(
```
euresti, line 67: Since Literal doesn't exist in older pythons (and will get turned to Any) you'll probaby want to use
```
if sys.version_info >= (3, 8):
```
to keep the old overload.

Also if you use 
```
import typing
if sys.version_info >= (3, 8):
    typing.Literal
```

You avoid the error message on import.

euresti (changes requested):
Thanks for doing this!  Lovely to see more people get involved in the type annotations.

I think the problem won't be which python you're running but which python you're type-checking for.  

For example: 
```
[mypy]
python_version=3.7
```
or `--python-version 3.7`

What appears to happen in that case is that `Literal` doesn't exist and the type of `takes_self` gets an Any.  This sees happens silently by the way (at least when I ran it I got no error messages).  I left comments on how to fix this.

I wonder if we could convince mypy to allow using `Literal` in stub files even if they aren't present in the current python. Literal is just so useful.

euresti (approved):

hynek:
Pre-commit’s isort is failing for `__init__.pyi`.

hynek:
Thanks!

PR #783, opened 2021-03-22
Don't cancel all builds if one fails

hynek:


hynek:
Thanks @hugovk!

hugovk:
You're welcome!

PR #785, opened 2021-04-04
Added missing type argument for `Attribute` class when used in type stubs

erictraut:
With this change, all exported symbols from the library have well-defined and fully-known types (a "type completeness score" of 100% according to pyright).


hynek:
Thanks Eric! And thanks for reaching out, I’ll get back to you by Tuesday by the latest!

PR #786, opened 2021-04-06
Make NG APIs official

### LINKED ISSUE #668: Feedback to next generation APIs

hynek:
I think it's time to go this step! I've been using these APIs in production _exclusively_ and the feedback to them was overwhelmingly positive. They are also officially supported by Mypy without any plugin hacks.

Fixes #668

### DIFF
--- a/changelog.d/668.change.rst
+++ b/changelog.d/668.change.rst
@@ -1,0 +1,5 @@
+The instant favorite `next-generation APIs <https://www.attrs.org/en/stable/api.html#next-gen>`_ are not provisional anymore!
+
+They are also officially supported by Mypy as of their `0.800 release <https://mypy-lang.blogspot.com/2021/01/mypy-0800-released.html>`_.
+
+We hope the next release will already contain an (additional) importable package called ``attrs``.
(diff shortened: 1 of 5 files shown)


Tinche (approved):

Tinche:
Just this Friday I gave an internal learning session talk to my team about the NG APIs (and some new cattrs features), advising them to start using them going forward :)

PR #787, opened 2021-04-11
New attr.cmp_using helper function

botant:
# Pull Request Check List

    - [?] ...and used in the stub test file `tests/typing_example.py`.

### DIFF
--- a/src/attr/_cmp.py
+++ b/src/attr/_cmp.py
@@ -2,7 +2,8 @@ from __future__ import absolute_import, division, print_function
 
 import functools
 
-from ._make import _make_ne, attrib, attrs
+from ._compat import new_class
+from ._make import _make_ne
 
 
 _operation_names = {"eq": "==", "lt": "<", "le": "<=", "gt": ">", "ge": ">="}
@@ -43,29 +44,12 @@ def cmp_using(
     .. versionadded:: 21.1.0
     """
 
-    @attrs(slots=True, eq=False, order=False)
-    class Comparable(object):
-        value = attrib()
-        _requirements = []
-
-        def _is_comparable_to(self, other):
-            """
-            Check whether `other` is comparable to `self`.
-            """
-            for func in self._requirements:
-                if not func(self, other):
-                    return False
-            return True
-
-    if class_name is None:
-        Comparable.__qualname__ = Comparable.__name__
-    else:
-        Comparable.__name__ = class_name
-        Comparable.__qualname__ = class_name
-
-    # Add same type requirement.
-    if require_same_type:
-        Comparable._requirements.append(_check_same_type)
+    body = {
+        "__slots__": ["value"],
+        "__init__": _make_init(),
+        "_requirements": [],
+        "_is_comparable_to": _is_comparable_to,
+    }
 
     # Add operations.
     num_order_fucntions = 0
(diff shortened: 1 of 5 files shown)


botant:
Hello @hynek, I've added the function `cmp_using` in this pull request.

It is still draft for one reason.  I've used `functools.total_ordering` so complete the ordering functions, and I found that the functions such as `_ge_from_lt` are defined directly in the `functools` module, and because of that I can't set their `__qualname__` to get well-behaved methods.

My question therefore is "how much does this matter"? If this this matters, then we need an alternative to total_ordering.

Let me know what you think.

Thanks

botant:
Example of numpy usage:

```{python}
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.
PyDev console: using IPython 7.20.0
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
import numpy as np
import attr
@attr.s(eq=True, order=False)
class A:
    value = attr.ib()
    
a1 = A(np.array([[1,2],[3,4]]))
a2 = A(np.array([[1,2],[3,4]]))
a1 == a2
  File "/home/antonio/.venv/attrs/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3427, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-74942f9c531d>", line 1, in <module>
    a1 == a2
  File "<attrs generated eq __main__.A>", line 4, in __eq__
    return  (
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

@attr.s(eq=True, order=False)
class B:
    value = attr.ib(eq=attr.cmp_using(eq=np.array_equal))
    
<ipython-input-8-6aa555853f34>:3: SyntaxWarning: You have customized the behavior of `eq` but not of `order`.  This is probably a bug.
  value = attr.ib(eq=attr.cmp_using(eq=np.array_equal))
b1 = B(np.array([[1,2],[3,4]]))
b2 = B(np.array([[1,2],[3,4]]))
b1 == b2
Out[11]: True
```

botant:
Finally, I believe we should update the following warning: 

```{python}
@attr.s(eq=True, order=False)
class B:
    value = attr.ib(eq=attr.cmp_using(eq=np.array_equal))
    
<ipython-input-8-6aa555853f34>:3: SyntaxWarning: You have customized the behavior of `eq` but not of `order`.  This is probably a bug.
  value = attr.ib(eq=attr.cmp_using(eq=np.array_equal))
```

If it is possible to tell that `order=False` for the entire class, then we can skip the warning altogether. If not, we can suggest setting `order=False`.

hynek:
Oh god that's a lot more code than I hoped for. 😅

Firstly, I don't think the qualname stuff matters at all, but it seems like your Comparable creation is more complicated than it needs to be.

Have you considered building a dict and then using `new_class()` (we have a helper in _compat for Python 2) instead of patching an existing class?

Also you broke almost all builds. :D

botant:
Got it, I'll simplify it then.  :-)

PS: the builds are breaking intentionally, as I committed a failing test.
That is why I marked the PR as draft.

hynek:
_cough_ 😇

botant:
So... here is a version using `new_class`, passing tests and 100% coverage.

I even installed Python 2.7 to be sure everything worked. :smile:

hynek:
@botant  thanks! would you mind removing the stray white space  edits from the pyi file? Or are they on purpose?

@Julian since IIRC this is implementing an idea of yours, would you mind having a look if it’s what you had in mind?

botant:
From `__init__`? I found that weird as well, but I assumed it was okay since it was formatted by `black`.

Happy to rollback if you want.

hynek:
Yeah, `__init__.pyi`. I think we might be still excluding pyi files from black? I’ll need to re-investigate.  Anyhow, the removed empty lines are nice, please give them back. :)

botant:
Sure thing.

src/attr/_cmp.py
```diff
@@ -0,0 +1,150 @@
...
+
+    body = {
+        "__slots__": ["value"],
+        "__init__": _make_init(),
+        "_requirements": [],
+        "_is_comparable_to": _is_comparable_to,
+    }
+
+    # Add operations.
+    num_order_fucntions = 0
```
Julian, line 55:
```suggestion
    num_order_functions = 0
```

Julian:
Nice! This looks great to me yeah!

src/attr/_cmp.py
botant: Good catch!  It was just a typo, not a bug.

botant:
> Yeah, `__init__.pyi`. I think we might be still excluding pyi files from black? I’ll need to re-investigate. Anyhow, the removed empty lines are nice, please give them back. :)

Done!

hynek:
Thanks! I'll write some docs and look into the warning thing!

Julian:
Big props for sticking with this @botant.

botant:
Thanks a lot @hynek and @Julian, it's been a pleasure to collaborate. Great experience (for me :smile:).

Issue #788, opened 2021-04-12
auto_attribs + Factory + validator doesn't work


belm0:
```python3
import attr
from functools import partial
import numpy as np

@attr.attrs(auto_attribs=True)
class Foo:
    position: np.ndarray = attr.Factory(partial(np.zeros, 3))
    #position: np.ndarray = attr.attrib(default=attr.Factory(partial(np.zeros, 3)))  # this works

    @position.validator
    def check_position(self, attribute, value):
        if value.size != 3:
            raise ValueError
```

```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "<input>", line 6, in Foo
AttributeError: 'Factory' object has no attribute 'validator'
```

version: 20.2.0

hynek:
The decorator syntax only works with attr.ibs: https://www.attrs.org/en/stable/init.html#decorator

We could technically fudge it into factories, but that wouldn't fix all other use cases and would only cause more confusion.

belm0:
the docs for `auto_attribs`:
> If you assign a value to those attributes (e.g. x: int = 42), that value becomes the default value like if it were passed using attr.ib(default=42). Passing an instance of Factory also works as expected.

Yet passing a Factory instance did not work as expected.  Please document the decorator limitation here.

PR #789, opened 2021-04-12
document auto_attribs Factory limitation

belm0:
Raised in #788

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1336,7 +1336,8 @@ def attrs(
         If you assign a value to those attributes (e.g. ``x: int = 42``), that
         value becomes the default value like if it were passed using
         ``attr.ib(default=42)``.  Passing an instance of `Factory` also
-        works as expected.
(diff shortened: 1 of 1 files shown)


hynek:
Hey, I've decided to add a big banner, not just the single factory use case. See https://github.com/python-attrs/attrs/commit/685f865d50eca222b4c0326e5aabe3b54cb2c0b3 and https://www.attrs.org/en/latest/api.html#attr.s

Sorry and thank you!

belm0:
@hynek I'm not sure the merged documentation will be that effective at avoiding confusion.  It's still stated that "passing an instance of Factory is fine", only to contradict that later with the warning (hurried users may not make it that far).  The warning itself is somewhat roundabout, and I'm not sure the user will immediately associate `must use attr.ib` with `can't assign Factory instance`.

As far as repo process, I think it would have been better to have a dialogue on my open PR.  Even though you're the library author, it's still worth taking advantage of peer review.

PR #791, opened 2021-04-13
docs: more explicit about Factory caveats

belm0:
The intent is to avoid catching readers off guard from the contradiction between "assigning Factory will work as expected" and the subsequent warning.

Raised in #788


hynek (approved):

hynek:
Thanks :)

PR #800, opened 2021-05-04
Remove warning if eq is customized & order is not

hynek:
Since this warning is raised from the attr.ib/attr.field call, we cannot verify
whether order is on at all for the whole class. Fixing this would require to
check _after_ the class is assembled which is both additional complexity as
well as performance overhead. Unless we get inundated by bug reports, this
doesn't seem worth it.

@botant @Julian

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1073,11 +1073,6 @@ _CMP_DEPRECATION = (
     "2021-06-01.  Please use `eq` and `order` instead."
 )
 
-_EQ_ORDER_CUSTOMIZATION = (
-    "You have customized the behavior of `eq` but not of `order`.  "
-    "This is probably a bug."
-)
-
 
 def _determine_attrs_eq_order(cmp, eq, order, default_eq):
     """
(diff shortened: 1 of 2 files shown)


botant:
Alternatively, we could leave the warning and suggest that the user explicitly sets `order=False` to avoid the warning.  I don't know which one is worse.

hynek:
I don't think I want to put any strain on _all_ users just to save a _few_ people from using it wrong.

botant:
> I don't think I want to put any strain on _all_ users just to save a _few_ people from using it wrong.

What I mean is that *if* setting `eq` but not `order` is likely to be a bug for most users, then we should leave the warning.

On the other hand, *if* setting `eq` but not `order` is likely to be intentional for most users, then *maybe* we should remove the warning.

My own view is that this will be a bug more often than not, but I might be wrong.

Another consideration is the relative value of a nuisance for some vs a protection for others. It depends on how annoying the nuisance is, and the value of the warning.

This is all subjective, of course.

hynek:
The nice thing is that we can always add the warning later, if people run into this problem frequently.

The problem is that making that warning properly is non-trivial, and having stray warnings devalues warnings to some kind of car alarm. I want every warning to be actionable.

In any case, this code is going out. If it shows to be a problem, we'll have to add the warning to ClassBuilder I guess. 🤔 I just don't want to delay 21.1 any further.

Issue #804, opened 2021-05-06
`attr.evolve` no longer supports setting an attr object as attribute


ProgVal:
Hi,

I don't know if this is an intentional side-effect of #759, but `attr.evolve` no longer works in the non-recursive recursive case if the argument is an attr class.

Here is a snippet to reproduce it:

```py
import attr

@attr.s
class Cls1:
    param1 = attr.ib()


@attr.s
class Cls2:
    param2 = attr.ib()


obj2a = Cls2(param2="a")
obj2b = Cls2(param2="b")

obj1a = Cls1(param1=obj2a)
obj1b = attr.evolve(obj1a, param1=obj2b)

print(obj1b)
```

Before fe6eb3120013833d1927374eb5e826b72815a775, it printed this: `Cls1(param1=Cls2(param2='b'))`

But after this commit, it raises this error:

```
Traceback (most recent call last):
  File "test.py", line 17, in <module>
    obj1b = attr.evolve(obj1a, param1=obj2b)
  File "/home/dev/.local/lib/python3.7/site-packages/attrs-21.2.0.dev0-py3.7.egg/attr/_funcs.py", line 347, in evolve
    changes[init_name] = evolve(value, **changes[init_name])
TypeError: evolve() argument after ** must be a mapping, not Cls2
```

Thanks!

staaam:
same here. Issue is in attr._funcs.evolve at:
```
    for a in attrs:
        ...
        elif has(value):
            # Evolve nested attrs classes
            changes[init_name] = evolve(value, **changes[init_name])
```
`value` - is old value (param1 from example above)
`changes[init_name]` - is new value (obj2b from above) 
but for recursive evolve it should be dict
probably the best way to make it backward compatible will be to check
`elif has(value) and isinstance(changes[init_name], collections.Mapping):`

hynek:
Yeah, we'll have to revert #759 and push out 21.2.

I'm gonna wait a bit with the release to see if there's other breakages that should be fixed.

hynek:
21.2 with the fix is on PyPI.

PR #805, opened 2021-05-06
Fix issue with recursive evolve when attrs used

staaam:
Fix issue https://github.com/python-attrs/attrs/issues/804

### DIFF
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import copy
 
-from ._compat import iteritems
+from ._compat import Mapping, iteritems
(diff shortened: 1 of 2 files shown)


ProgVal:
It does fix the issue, but I wonder if this is really what it should do. What if someone actually wants to replace an attr class with a dict?

Is it too late to revert `attr.evolve`, and add a new function for recursive updates?

staaam:
> It does fix the issue, but I wonder if this is really what it should do. What if someone actually wants to replace an attr class with a dict?
> 
> Is it too late to revert recursion from `attr.evolve`, and add a new function for recursive updates?

good point.. probably better to make separate call for evolve_recursive

hynek:
I guess we could just copy-paste the reverted evolve version and call it evolve_recursive and call it a day?

PR #806, opened 2021-05-06
Revert recursive evolve

### LINKED ISSUE #804: `attr.evolve` no longer supports setting an attr object as attribute

hynek:
Seems like the current implementation breaks a non-trivial amount of use cases.

So this reverts it and adds a regression test – lmk if I got it right (also the description of the test).

I welcome suggestions how to do it properly in separate PRs.

@ProgVal @staaam @sscherfke 

Fixes #804


ProgVal:
LGTM, thanks

You might also want to add a regression test where `obj2b` is a dict; to make sure one can still replace an attr class with a dict. But it seems like a pretty niche use-case, so it might not be worth it.

hynek:
Can you quickly provide one please?

ProgVal:
```
import attr

@attr.s
class Cls1:
    param1 = attr.ib()


@attr.s
class Cls2:
    param2 = attr.ib()


obj2a = Cls2(param2="a")
obj2b = {"foo": 42, "param2": 42}

obj1a = Cls1(param1=obj2a)
obj1b = attr.evolve(obj1a, param1=obj2b)

print(obj1b)
```

hynek:
Thanks!

hynek:
(If someone finds the time to contribute a copy-pasted `attr.evolve_recursive()` with appropriate tests/docs until tomorrow we could push a release that doesn't take anything away. 😇 Unfortunately, I'm busy for the rest of the day.)

sscherfke:
I’m afraid I’m too busy with home schooling and work, too. :-/

PR #815, opened 2021-05-14
Add __match_args__ to support match case destructuring in Python 3.10

tirkarthi:
# Pull Request Check List

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,13 +15,6 @@ Changes for the upcoming release can be found in the `"changelog.d" directory <h
 
 .. towncrier release notes start
 
-21.3.0.dev0 (2021-05-14)
-No significant changes.
-
-
-----
-
-
 21.2.0 (2021-05-07)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,13 @@ Changes for the upcoming release can be found in the `"changelog.d" directory <h
 
 .. towncrier release notes start
 
+21.3.0.dev0 (2021-05-14)
+No significant changes.
+
+
+----
+
+
 21.2.0 (2021-05-07)
 -------------------
 
--- a/changelog.d/815.changes.rst
+++ b/changelog.d/815.changes.rst
@@ -1,2 +1,3 @@
-Add ``__match_args__`` to generated class to support match case destructuring
-in Python 3.10. This can be controlled by ``match_args`` argument to ``@attrib.s``.
+``__match_args__`` are now generated to support Python 3.10's
+`Structural Pattern Matching <https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching>`_.
+This can be controlled by ``match_args`` argument to the class decorators.
--- a/changelog.d/815.changes.rst
+++ b/changelog.d/815.changes.rst
@@ -1,0 +1,2 @@
+Add ``__match_args__`` to generated class to support match case destructuring
+in Python 3.10. This can be controlled by ``match_args`` argument to ``@attrib.s``.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1369,7 +1369,7 @@ def attrs(
     :param bool collect_by_mro: Setting this to `True` fixes the way ``attrs``
        collects attributes from base classes.  The default behavior is
        incorrect in certain cases of multiple inheritance.  It should be on by
-       default but is kept off for backward-compatability.
+       default but is kept off for backward-compatibility.
 
        See issue `#428 <https://github.com/python-attrs/attrs/issues/428>`_ for
        more details.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -966,6 +966,13 @@ class _ClassBuilder(object):
 
         return self
 
+    def add_match_args(self):
+        self._cls_dict["__match_args__"] = tuple(
+            field.name
+            for field in self._attrs
+            if field.init and not field.kw_only
+        )
+
     def add_attrs_init(self):
         self._cls_dict["__attrs_init__"] = self._add_method_dunders(
             _make_init(
@@ -1192,6 +1199,7 @@ def attrs(
     getstate_setstate=None,
     on_setattr=None,
     field_transformer=None,
+    match_args=True,
 ):
     r"""
     A class decorator that adds `dunder
@@ -1407,6 +1415,10 @@ def attrs(
         this, e.g., to automatically add converters or validators to
         fields based on their types.  See `transform-fields` for more details.
 
+    :param bool match_args:
+        If `True` it sets __match_args__ in the class to support PEP 634. It is a
+        tuple of __init__ parameter names that are only positional arguments.
+
     .. versionadded:: 16.0.0 *slots*
     .. versionadded:: 16.1.0 *frozen*
     .. versionadded:: 16.3.0 *str*
@@ -1555,6 +1567,9 @@ def attrs(
                     "Invalid value for cache_hash.  To use hash caching,"
                     " init must be True."
                 )
+
+        if match_args and "__match_args__" not in cls.__dict__:
+            builder.add_match_args()
 
         return builder.build_class()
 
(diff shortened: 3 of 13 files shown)


tirkarthi:
Sorry for the lint errors. Hopefully it's all fine now.

src/attr/_make.py
```diff
@@ -1562,6 +1576,9 @@ def wrap(cls):
                     " init must be True."
                 )
 
+        if match_args and "__match_args__" not in cls.__dict__:
```
hynek, line 1579: Please use [`_has_own_attribute`](https://github.com/python-attrs/attrs/blob/20c2d4fc0c72a844e12899d5d5e0adbc4741ce2c/src/attr/_make.py#L403) for the second check.

I'm aware that the implementation is currently obtuse and depending on how Larry's typing proposal will go, we'll fix it centrally eventually.

src/attr/_make.py
```diff
@@ -1413,6 +1421,11 @@ def attrs(
         this, e.g., to automatically add converters or validators to
         fields based on their types.  See `transform-fields` for more details.
 
+    :param bool match_args:
+        If `True` it sets __match_args__ in the class to support PEP 634. It
+        is a tuple of __init__ parameter names that are only positional
+        arguments.
```
hynek, line 1427:
```suggestion
        If `True`, set ``__match_args__`` on the class to support `PEP 634
        <https://www.python.org/dev/peps/pep-0634/>`_ (Structural Pattern
        Matching). It is a tuple of all positional-only ``__init__`` parameter
        names.
```

src/attr/_make.py
hynek: This behavior is also somewhat inconsistent with the rest of `attrs`, where we overwrite unless `auto_detect` is set to True (default in new APIs). Is this the behavior of dataclasses?

hynek: It also doesn't make sense to add it on Pythons older than 3.10, no?

changelog.d/815.changes.rst
```diff
@@ -0,0 +1,2 @@
+Add ``__match_args__`` to generated class to support match case destructuring
+in Python 3.10. This can be controlled by ``match_args`` argument to ``@attrib.s``.
```
hynek, line 2:
```suggestion
``__match_args__`` are now generated to support Python 3.10's `Structural Pattern Matching <https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching>`_.
This can be controlled by ``match_args`` argument to the class decorators.
The argument is ignored for Python versions older than 3.10.
```

hynek (changes requested):
Thanks and welcome to the team!

Additionally to the inline comments, please:

- needs to be added to _next_gen.py
- it would be nice if you could adapt the test docstrings to https://jml.io/pages/test-docstrings.html but I can do it myself too later. Mostly just drop the `Test`
- as Tin mentioned, we need some tests. Please add a new file like `test_pattern_matching.py`, add your tests, and exclude the file on older Python versions in conftest.py – there's some examples how to do it!

CHANGELOG.rst
```diff
@@ -15,6 +15,13 @@ Changes for the upcoming release can be found in the `"changelog.d" directory <h
 
 .. towncrier release notes start
 
+21.3.0.dev0 (2021-05-14)
```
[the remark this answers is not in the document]
tirkarthi: Sorry, I ran towncrier by mistake and committed it. Reverted now.

tirkarthi:
Changes done :

* Updated docs. Removed "Test" from docstrings and updated a few. Feel free to makes changes.
* Added `test_pattern_matching.py` on Python 3.10.
* Disabled black since black doesn't support formatting pattern matching : https://github.com/psf/black/issues/2242
* Disabled flake8 since pyflakes has an issue with keyword argument matching : https://github.com/PyCQA/pyflakes/issues/634

> needs to be added to _next_gen.py

I am not sure about the changes to be made here since `match_args` is not a signature change in `define` and `field`

src/attr/_make.py
tirkarthi: In the initial implementation of the pattern matching PEP there was no flag to disable `__match_args__` generation but there was a request raised later on a flag to opt-out though `__match_args__ = ()` being set manually was proposed. Discussion : https://bugs.python.org/issue43764

hynek:
`define`, `mutable`, and `frozen` all wrap `attr.s`, so the argument needs to be added to them (and their type stubs) too?

tirkarthi:
Thanks, added in https://github.com/python-attrs/attrs/pull/815/commits/0cd51b485e560d201b7703c37eeb1b1d854cd8b6

tirkarthi:
It seems pre-commit  uses Python 3.8 and interrogate fails to parse the Python 3.10 syntax. Excluding the file in pyproject.toml for interrogate doesn't seem to ignore it. debug_statement action of pre-commit also uses the syntax and fails.

black seems not to exclude the files as per pre-commit pinned version. Upgrading black gives some more formatting changes on unrelated files as it seems black has changed the formatting.

hynek:
I think you'll have to block the match tests to black and interrogate in `.pre-commit-config.yaml`.

tirkarthi:
As noted in https://stackoverflow.com/questions/61032281/exclude-some-files-on-running-black-using-pre-commit, it seems that pre-commit gets all python files and just passes it to the command and using `--exclude` for black doesn't help. I have bumped to Python 3.10 for interrogate and debug_statement. For black I have added exclude from pre-commit end. `tox -e lint` seems to pass.

```
tox -e lint
lint installed: appdirs==1.4.4,cfgv==3.2.0,distlib==0.3.1,filelock==3.0.12,identify==2.2.4,nodeenv==1.6.0,pre-commit==2.12.1,PyYAML==5.4.1,six==1.16.0,toml==0.10.2,virtualenv==20.4.6
lint run-test-pre: PYTHONHASHSEED='2125636111'
lint run-test: commands[0] | pre-commit run --all-files
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
interrogate..............................................................Passed
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Debug Statements (Python)................................................Passed
Check Toml...............................................................Passed
_______________________________________________________________________________ summary ________________________________________________________________________________
  lint: commands succeeded
  congratulations :)

```

tirkarthi:
codecov gives me 403, I am not sure about the code coverage error.

hynek:
codecov is probably failing, because python 3.8 fails and doesn't upload.

now we have the problem, that the linter _needs_ python 3.10 and that's not present in the 3.8 container

I guess there's two options:

1. block match from interrogate
2. move lint to py310

I guess you _could_ try option 2 (change the `[gh-actions]` section in tox.ini) but if that fails for other reasons (or you don't want to deal with it), just go with 1.

tirkarthi:
Thanks, I guess option 2 makes more sense since eventually the python version has to be upgraded to support parsing. I have pushed commit to bump basepython to 3.10 in lint step.

tirkarthi:
tox and gh-actions are slightly new to me so I got confused. I pushed the lint job from py38 to py310 and also made basepython 3.10 in lint step. I guess this should run link in python 3.10 container with python 3.10 as base. Please correct me if I am wrong.

hynek:
Looks good, but I think you still add the tuple on Python <3.10? (On my phone for the rest of the day)

tirkarthi:
Yes, I am adding `__match_args__` tuple always. I don't see any backwards compatibility issues with the same. I will be curious how dataclasses backport also deals with the change too. I can add a version check to add tuple only for Python 3.10+ if you want. In that case I guess match_args=True will always be there in signature and it's just that tuple generation becomes python 3.10+?

hynek:
Yeah I don’t have any comparability problems in mind – it just seems wasteful to set an attribute on all classes even tho they can’t use it.

And yes, just add something to the tune of  “this argument is ignored on Python versions before 3.10” to the doc string and call it a day.

hynek (approved):
You've suffered enough and I have to approve every CI run so I'd rather do the small stuff myself than shooing you around.

Thanks for playing along!

tirkarthi:
Thanks @hynek for the review and merge :)

Issue #654, opened 2020-06-19
[RFC] Drop support for older Python versions (2.7 and 3.5)


sscherfke:
I propose that we end support for Python 2.7 and Python 3.5.

Python 2.7 is already EOL and the last patch for 3.5 will be released [this July](https://www.python.org/dev/peps/pep-0478/) (if I interpret the PEP correctly).

According to [pypistats.org](https://pypistats.org/packages/attrs), the share of py27 is approx. 11% and declining.  Py35 has approx. 7%.

Setting py36 as minimum version would allow us to use these features:

- [f-Strings](https://www.python.org/dev/peps/pep-0498/)
- [dict ordered](https://www.python.org/dev/peps/pep-0468/)
- [type annotation syntas](https://www.python.org/dev/peps/pep-0526/)

These features would in turn allow us to:

- Use f-Strings everywhere. =)
- Make `auto_attribs=True` the default (and deprecate this kwarg altogether) (#487)
- Implement (optional) auto-convertion/-validation based on type annotations (#424, #649)
- Write test with attrs classes that directly use type annotations (auto_attrs) w/o worrying about syntax errors on older versions
- Use inline type annotations and get rid of the pyi files which could increase readability and maintainability of the code.
- Get rid of compatibilty code which would also increase maintainability.

This would require a new major release. The next one will be v20 which is, as chance would have it, a nice round number. ;-)

wsanchez:
Per `pypistats python_minor attrs --last-month`:

```console
| category | percent | downloads  |
|----------|--------:|-----------:|
| 3.7      |  38.26% |  9,540,224 |
| 3.6      |  27.61% |  6,882,498 |
| 2.7      |  13.54% |  3,376,839 |
| 3.8      |  12.10% |  3,017,537 |
| 3.5      |   8.00% |  1,993,390 |
| null     |   0.31% |     76,454 |
| 3.4      |   0.14% |     35,484 |
| 3.9      |   0.04% |      9,089 |
| 3.10     |   0.00% |        295 |
| 3.3      |   0.00% |        139 |
| 2.6      |   0.00% |         81 |
| 3.1      |   0.00% |          1 |
| Total    |         | 24,932,031 |

Date range: 2020-05-01 - 2020-05-31
```

hynek:
I'm afraid 2.7 support is one of our main advantages over dataclasses and I don't want to leave those users in the rain (I'd have to lie if I said, that I have zero Python 2.7 apps in prod).

[...]

PR #819, opened 2021-05-20
Initial implementation of a faster repr

Tinche:
Initial implementation, CHANGELOG to follow.

Benchmark used:
```
pyperf timeit -g -s "from attr import make_class, attrib; A = make_class('A', ['a', 'b', 'c', 'd', 'e']); a = A(1, 2, 3, 4, 5)" "repr(a)"
```

Old code:
```
Mean +- std dev: 2.09 us +- 0.07 us
```

New code:
```
Mean +- std dev: 957 ns +- 27 ns
```

### DIFF
--- a/changelog.d/819.changes.rst
+++ b/changelog.d/819.changes.rst
@@ -1,1 +1,1 @@
-The generated `__repr__` is significantly faster on Pythons with F-strings.
+The generated ``__repr__`` is significantly faster on Pythons with F-strings.
--- a/changelog.d/819.changes.rst
+++ b/changelog.d/819.changes.rst
@@ -1,0 +1,1 @@
+The generated `__repr__` is significantly faster on Pythons with F-strings.
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -8,6 +8,7 @@ import warnings
 
 PY2 = sys.version_info[0] == 2
 PYPY = platform.python_implementation() == "PyPy"
+HAS_F_STRINGS = sys.version_info[:2] >= (3, 7)
 PY310 = sys.version_info[:2] >= (3, 10)
 
 
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -8,7 +8,11 @@ import warnings
 
 PY2 = sys.version_info[0] == 2
 PYPY = platform.python_implementation() == "PyPy"
-HAS_F_STRINGS = sys.version_info[:2] >= (3, 7)
+HAS_F_STRINGS = (
+    sys.version_info[:2] >= (3, 7)
+    if not PYPY
+    else sys.version_info[:2] >= (3, 6)
+)
 PY310 = sys.version_info[:2] >= (3, 10)
 
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1905,7 +1905,6 @@ if HAS_F_STRINGS:
                 "%s={%s!r}" % (name, accessor)
                 if r == repr
                 else "%s={%s_repr(%s)}" % (name, name, accessor)
-                )
             )
             attribute_fragments.append(fragment)
         repr_fragment = ", ".join(attribute_fragments)
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1935,8 +1935,10 @@ if HAS_F_STRINGS:
         )
         lines.append("  finally:")
         lines.append("    working_set.remove(id(self))")
-        script = "\n".join(lines)
-        return _make_method("__repr__", script, unique_filename, globs=globs)
+
+        return _make_method(
+            "__repr__", "\n".join(lines), unique_filename, globs=globs
+        )
 
 
 else:
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1886,7 +1886,7 @@ if HAS_F_STRINGS:
             if a.repr is not False
         )
         globs = {
-            f"{name}_repr": r
+            name + "_repr": r
             for name, r, _ in attr_names_with_reprs
             if r != repr
         }
@@ -1896,12 +1896,16 @@ if HAS_F_STRINGS:
         attribute_fragments = []
         for name, r, i in attr_names_with_reprs:
             accessor = (
-                f"self.{name}" if i else f'getattr(self, "{name}", NOTHING)'
+                "self." + name
+                if i
+                else 'getattr(self, "' + name + '", NOTHING)'
             )
             fragment = (
-                f"{name}={{{accessor}!r}}"
+                "{name}={{{accessor}!r}}".format(name=name, accessor=accessor)
                 if r == repr
-                else f"{name}={{{name}_repr({accessor})}}"
+                else "{name}={{{name}_repr({accessor})}}".format(
+                    name=name, accessor=accessor
+                )
             )
             attribute_fragments.append(fragment)
         repr_fragment = ", ".join(attribute_fragments)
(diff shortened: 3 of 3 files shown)


src/attr/_compat.py
```diff
@@ -8,6 +8,7 @@
 
 PY2 = sys.version_info[0] == 2
 PYPY = platform.python_implementation() == "PyPy"
+HAS_F_STRINGS = sys.version_info[:2] >= (3, 7)
```
hynek, line 11: I think you can add PyPy 3.6 here too!

hynek:
Looks like you’ll need a conditionally imported `repr_fstrings.py` or something.

Tinche:
> Looks like you’ll need a conditionally imported `repr_fstrings.py` or something.

I think it just means I can't use f-strings while preparing the function. I can just use string concatenation or `.format()` instead, not a big deal.

hynek:
y u hate Python 2 Tin

sscherfke:
Maybe it’s time for #654 🙊 👼

hynek:
Still getting a depressing amount of Python 2 downloads 😢

sscherfke:
The current versions that still support Python 2 won’t go away. :)

Many of the new features are related to Python-3-only features (e.g., typing, pattern matching) and are not benefical for Python 2 users anyways. So I don’t realy see the problem with new versions only supporting python 3.6 (or 3.7) upwards. :)

Tinche:
@hynek I don't understand the problem though, I don't think I touched that line?

Tinche:
Oh it's the trailing comma? FML

Tinche:
Ok, looks like just some line length warnings which I can fix. Ready for a full review.

hynek:
I'm sure you "can" – but will you, too? :D

Tinche:
Of course, when I get a free minute

hynek:
Looks like one more line: https://codecov.io/gh/python-attrs/attrs/src/f37fc134b66c14b03a09fc09ca59f188537e1911/src/attr/_make.py#L1976

Tinche:
> Looks like one more line: https://codecov.io/gh/python-attrs/attrs/src/f37fc134b66c14b03a09fc09ca59f188537e1911/src/attr/_make.py#L1976

Hm, if I remove this, 3.5 and 3.6 start failing.

hynek:
Guess we have to add 3.6 to coverage then

src/attr/_make.py
```diff
@@ -1873,64 +1874,132 @@ def _add_eq(cls, attrs=None):
...
+                working_set = set()
+                _already_repring.working_set = working_set
+
+            if id(self) in working_set:
+                return "..."
+            real_cls = self.__class__
+            if ns is None:
+                qualname = getattr(real_cls, "__qualname__", None)
+                if qualname is not None:
+                    class_name = qualname.rsplit(">.", 1)[-1]
```
hynek, line 1976:
```suggestion
                if qualname is not None:  # pragma: no cover
                    # This case only happens on Python 3.5 and 3.6. We exclude
                    # it from coverage, because we don't want to slow our test
                    # suite by running them under coverage too for this one
                    # line.
                    class_name = qualname.rsplit(">.", 1)[-1]
```

changelog.d/819.changes.rst
```diff
@@ -0,0 +1 @@
+The generated `__repr__` is significantly faster on Pythons with F-strings.
```
hynek, line 1:
```suggestion
The generated ``__repr__`` is significantly faster on Pythons with F-strings.
```

src/attr/_make.py
```diff
@@ -1873,64 +1874,136 @@ def _add_eq(cls, attrs=None):
...
 
+if HAS_F_STRINGS:
+
+    def _make_repr(attrs, ns, cls):
+        unique_filename = "repr"
+        # Figure out which attributes to include, and which function to use to
+        # format them. The a.repr value can be either bool or a custom
+        # callable.
+        attr_names_with_reprs = tuple(
+            (a.name, repr if a.repr is True else a.repr, a.init)
```
hynek, line 1885:
```suggestion
            (a.name, (repr if a.repr is True else a.repr), a.init)
```

This took me *minutes* to decipher. 🙈

src/attr/_make.py
```diff
@@ -1873,64 +1874,136 @@ def _add_eq(cls, attrs=None):
...
+        globs["NOTHING"] = NOTHING
+        attribute_fragments = []
+        for name, r, i in attr_names_with_reprs:
+            accessor = (
+                "self." + name
+                if i
+                else 'getattr(self, "' + name + '", NOTHING)'
+            )
+            fragment = (
+                "{name}={{{accessor}!r}}".format(name=name, accessor=accessor)
```
hynek, line 1905: look if you wanna optimize, optimize! 🤪

```suggestion
                "%s={%s!r}" % (name, accessor)
```

src/attr/_make.py
```diff
@@ -1873,64 +1874,136 @@ def _add_eq(cls, attrs=None):
...
+                if r == repr
+                else "{name}={{{name}_repr({accessor})}}".format(
+                    name=name, accessor=accessor
```
hynek, line 1908:
```suggestion
                else "%s={%s_repr(%s)}" % (name, name, accessor)
```

src/attr/_make.py
```diff
@@ -1873,64 +1874,136 @@ def _add_eq(cls, attrs=None):
...
+        lines.append("    working_set = {id(self),}")
+        lines.append("    _already_repring.working_set = working_set")
+        lines.append("  else:")
+        lines.append("    if id(self) in working_set:")
+        lines.append("      return '...'")
+        lines.append("    else:")
+        lines.append("      working_set.add(id(self))")
+        lines.append("  try:")
+        lines.append(
+            "    return f'{}({})'".format(cls_name_fragment, repr_fragment)
```
hynek, line 1935:
```suggestion
            "    return f'%s(%s)'" % (cls_name_fragment, repr_fragment)
```

src/attr/_make.py
```diff
@@ -1873,64 +1874,136 @@ def _add_eq(cls, attrs=None):
...
+        )
+        lines.append("  finally:")
+        lines.append("    working_set.remove(id(self))")
+        script = "\n".join(lines)
+        return _make_method("__repr__", script, unique_filename, globs=globs)
```
hynek, line 1940: Let's save ourselves the variable and let return breathe.

```suggestion

        return _make_method(
            "__repr__", "\n".join(lines), unique_filename, globs=globs
        )
```

hynek (changes requested):
I haven't spent super much time understanding it, because the tests are passing so it should be fine. #yolo

But I'd like to avoid snail-formatting in new code (maybe should be adding to the contribution guide), so let's not do it. :)

Tinche:
> I haven't spent super much time understanding it, because the tests are passing so it should be fine. #yolo
> 
> But I'd like to avoid snail-formatting in new code (maybe should be adding to the contribution guide), so let's not do it. :)

What's snail-formatting btw?

src/attr/_make.py
Tinche: I usually use a variable so I can just insert a print easily when I need to debug it, but sure :)

hynek:
No, snail formatting is what I call `str.format()` because it's slow as a snail. :D 🐌

Modulo + tuple is the second fastest approach.

hynek:
Looks like our tireless teamwork broke almost all builds!

hynek (approved):

hynek:
Thanks! 🏎

Issue #826, opened 2021-06-15
Performance degradation after creating many classes with the same name


MrSage:
I suspect this is heavily related to https://github.com/python-attrs/attrs/issues/575, but my analysis suggests the issues showed up in `19.2.0 (2019-10-01)` and that issue was created before the release went live. 

I made a simple test harness to demonstrate the issue:

```
def create_class():
    @attr.s
    class MyClass(object):
        access_token = attr.ib()
        expires = attr.ib()
        refresh_token = attr.ib()
        user = attr.ib()
    
    return MyClass(
        access_token="string",
        expires=0,
        refresh_token="string",
        user="string",
    )
```

```
%timeit create_class()
```

I ran this over `attrs` versions: 18.2, 19.1, 19.2, 20.1, 21.1, and 21.2.

I collected the outputs:
```
[
    # Version num, timeit output, whether or not timeit warned for caching
    (18.2, "358 µs ± 4.47 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)", False),
    (19.1, "356 µs ± 1.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)", False),
    (19.2, "22 ms ± 9.5 ms per loop (mean ± std. dev. of 7 runs, 1000 loops each)", "The slowest run took 4.71 times longer than the fastest"),
    (20.1, "22.4 ms ± 9.85 ms per loop (mean ± std. dev. of 7 runs, 1000 loops each)", "The slowest run took 4.74 times longer than the fastest"),
    (21.1, "22.9 ms ± 9.74 ms per loop (mean ± std. dev. of 7 runs, 1000 loops each)", "The slowest run took 5.12 times longer than the fastest"),
    (21.2, "23.4 ms ± 10.5 ms per loop (mean ± std. dev. of 7 runs, 1000 loops each)", ""The slowest run took 4.91 times longer than the fastest"),
]
```

This suggests that something that went along with the `19.2` release is causing a performance leak. We noticed the issue because our CPU metrics started to spike when we upgraded `attrs` from version 18.2 directly to 21.2.

euresti:
Looks like this change caused it (using git bisect):
```
eda9f2de6f7026039c3eee80297f0c95382d51d8 is the first bad commit
commit eda9f2de6f7026039c3eee80297f0c95382d51d8
Author: David Euresti <david@euresti.com>
Date:   Thu Aug 1 03:42:13 2019 -0700

[...]

PR #828, opened 2021-06-23
Rework linecache handling

Tinche:
Working version of a fix for #826

We deduplicate linecache entries, thus avoiding identical class creation time increasing linearly and saving memory in the process.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -329,16 +329,25 @@ def _make_method(name, script, filename, globs=None):
     if globs is None:
         globs = {}
 
-    _compile_and_eval(script, globs, locs, filename)
-
     # In order of debuggers like PDB being able to step through the code,
     # we add a fake linecache entry.
-    linecache.cache[filename] = (
-        len(script),
-        None,
-        script.splitlines(True),
-        filename,
-    )
+    count = 1
+    base_filename = filename
+    while True:
+        linecache_tuple = (
+            len(script),
+            None,
+            script.splitlines(True),
+            filename,
+        )
+        old_val = linecache.cache.setdefault(filename, linecache_tuple)
+        if old_val == linecache_tuple:
+            break
+        else:
+            filename = "{}-{}>".format(base_filename[:-1], count)
+            count += 1
+
+    _compile_and_eval(script, globs, locs, filename)
 
     return locs[name]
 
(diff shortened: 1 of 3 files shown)


hynek:
can we haz benchmark before reviewing? 🐶

Tinche:
Sure, put this into a01.py:
```
import attr


def create_class():
    @attr.s
    class MyClass(object):
        access_token = attr.ib()
        expires = attr.ib()
        refresh_token = attr.ib()
        user = attr.ib()
```

Then, on master:

```
> pyperf timeit -g -s "from a01 import create_class" "create_class()"
.....................
 976 us: 17 ###############################################################################
1.02 ms:  3 ##############
1.07 ms:  0 |
1.12 ms:  0 |
1.16 ms:  0 |
1.21 ms:  0 |
1.25 ms:  0 |
1.30 ms:  0 |
1.35 ms:  0 |
1.39 ms: 16 ##########################################################################
1.44 ms:  4 ###################
1.49 ms:  0 |
1.53 ms:  0 |
1.58 ms:  0 |
1.63 ms:  0 |
1.67 ms:  0 |
1.72 ms:  0 |
1.77 ms:  5 #######################
1.81 ms: 14 #################################################################
1.86 ms:  0 |
1.91 ms:  1 #####

WARNING: the benchmark result may be unstable
* the standard deviation (337 us) is 24% of the mean (1.42 ms)

Try to rerun the benchmark with more runs, values and/or loops.
Run 'python3.9 -m pyperf system tune' command to reduce the system jitter.
Use pyperf stats, pyperf dump and pyperf hist to analyze results.
Use --quiet option to hide these warnings.

Mean +- std dev: 1.42 ms +- 0.34 ms
```

And on the branch:
```
> pyperf timeit -g -s "from a01 import create_class" "create_class()"
.....................
371 us:  4 #####################
374 us: 15 ###############################################################################
376 us: 10 #####################################################
378 us: 11 ##########################################################
381 us:  5 ##########################
383 us:  4 #####################
385 us:  1 #####
387 us:  2 ###########
390 us:  3 ################
392 us:  2 ###########
394 us:  0 |
397 us:  0 |
399 us:  0 |
401 us:  1 #####
404 us:  1 #####
406 us:  0 |
408 us:  0 |
410 us:  0 |
413 us:  0 |
415 us:  0 |
417 us:  1 #####

Mean +- std dev: 381 us +- 8 us
```

19.1.0 is at `286 us +- 9 us`, so we're still a little slower. But it's not too bad!

hynek:
Well, it's O(1), right? That sounds good enough to _me_.

@euresti @MrSage – any feelings?

src/attr/_make.py
```diff
@@ -1632,30 +1640,12 @@ def _generate_unique_filename(cls, func_name):
...
-        if (
-            linecache.cache.setdefault(unique_filename, cache_line)
-            == cache_line
-        ):
-            return unique_filename
-
-        # Looks like this spot is taken. Try again.
-        count += 1
-        extra = "-{0}".format(count)
+    unique_filename = "<attrs generated {0} {1}.{2}>".format(
```
euresti, line 1643: I noticed that `_make_hash` does this:
```
    unique_filename = _generate_unique_filename(cls, "hash")
    type_hash = hash(unique_filename)
    ...
```
Does the fact that `type_hash` is no longer unique matter?  I don't think it does but I figure better to ask.

euresti (approved):
I love this fix.  Ship it!

MrSage:
This definitely gets rid of the surprise factors, I am still curious if the use of `linecache` is needed at all though. But the main concern is taken care of, certainly!

Tinche:
Great, I will add a changelog entry.

src/attr/_make.py
Tinche: Good question. Given two identical classes, equal instances will now hash the same. They will still not compare the same, I believe, so the contract is preserved. Not sure if this affects anything in the real world though. Maybe makes dicts and sets of identical classes a tiny bit less efficient?

hynek:
> This definitely gets rid of the surprise factors, I am still curious if the use of `linecache` is needed at all though. But the main concern is taken care of, certainly!

You mean in general? It’s so you can step through the generated methods with pdb.

Tinche:
Also it'll give you a nice stack trace.

hynek:
> Great, I will add a changelog entry.

🐶😇

hynek (approved):

PR #836, opened 2021-07-30
use an actual class for Factory hints

altendky:
This change relates to making the following code work with mypy in [desert](https://github.com/python-desert/desert).

https://github.com/python-desert/desert/blob/1fc8466ab0c6d2e1139421cf0d79646c90e1379a/src/desert/_make.py#L330-L333
```python
        if isinstance(field.default, attr.Factory):
            if field.default.takes_self:
                return attr.NOTHING
            return field.default.factory
```

Specifically, mypy gives the following errors (and other unrelated ones) without these changes.  Given that `Factory` is hinted as a function, it makes sense that we can't `isinstance()` against it.

```
src/desert/_make.py:330: error: Argument 2 to "isinstance" has incompatible type overloaded function; expected "Union[type, Tuple[Union[type, Tuple[Any, ...]], ...]]"
src/desert/_make.py:331: error: Item "None" of "Optional[Any]" has no attribute "takes_self"
src/desert/_make.py:333: error: Item "None" of "Optional[Any]" has no attribute "factory"
```

Draft for:
- [ ] Discussion and understanding of why a function was used to hint the `Factory` class to begin with.

# Pull Request Check List

### DIFF
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -65,27 +65,36 @@ NOTHING: object
 # Work around mypy issue #4554 in the common case by using an overload.
 if sys.version_info >= (3, 8):
     from typing import Literal
-    @overload
-    def Factory(factory: Callable[[], _T]) -> _T: ...
-    @overload
-    def Factory(
-        factory: Callable[[Any], _T],
-        takes_self: Literal[True],
-    ) -> _T: ...
-    @overload
-    def Factory(
-        factory: Callable[[], _T],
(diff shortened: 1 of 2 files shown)


altendky:
And the answer is right there...  so I probably just need to trick mypy in my super corner case.

https://github.com/python-attrs/attrs/blob/38580632ceac1cd6e477db71e1d190a4130beed4/src/attr/__init__.pyi#L63-L65

altendky:
Sorry for the noise.

PR #843, opened 2021-09-19
Fix bug in resolve_types with subclasses

### LINKED ISSUE #842: `attr.resolve_types` caches badly with inheritence
Reproduction case:

```py

import attr

@attr.define()
class A:
  n: int

@attr.define()
class B(A):
  pass

attr.resolve_types(A)
attr.resolve_types(B)

print(attr.fields(B))
```

I believe this is because `B` is somehow inheriting `A`'s `__attrs_types_resolved__`, even though it gets set later -- which means that it never had its types resolved, but it thinks it has.

This came up when I was using cattrs (I raised an issue there about a trouble I had debugging too :^), which means I do not control the resolution order of types without applying a `attr.resolve_types` decorator to every class (which might not work -- I'm not sure all my class declarations are in order either...)

euresti:
# Pull Request Check List

Fixes #842


euresti:
Does this type of change need a changelog entry?

hynek:
Thanks for stepping up real quick as usual!

Yeah I think it’s worth mentioning!

For every bug report there’s 10 people who didn’t bother.

also pls `src/attr/_funcs.py:390:80: E501 line too long (82 > 79 characters)` :)

euresti:
Ok.  Gotta go make breakfast but I'll get the changelog in after that.

hynek:
Oh god why is codecov angry. If you don’t have time for it, I’ll check once I’m back from Italy and off phone-only.

euresti:
Oh it's because nothing tests the "cached" flow.

euresti:
Sorry for all the lint failures. I don't have python 3.10 on my machine and you can only run pre-commit or tox -e lint if you have python 3.10.

euresti:
Ok I have no idea why codecov is complainign about this:
image

https://app.codecov.io/gh/python-attrs/attrs/compare/843/tree/src/attr/_compat.py

hynek:
It was complaining about the same thing on #840 and then magically stopped…some codecov flakiness? :-/

euresti:
It passed!

hynek:
🤡

hynek (approved):

hynek:
Thanks for the quick turnarounds!

Issue #754, opened 2021-02-03
threading.local global variable in _make.py breaks (cloud)pickling


pwais:
I was trying to use `cloudpickle` and `attrs`-based object and ran into a really weird issue where cloudpickle found a threading.local that obviously can't be pickled.  When I instrumented the pickle and cloudpickle source code in my container, I found the issue traces back to the `_already_repring` global threading.local object in `_make.py`:
https://github.com/python-attrs/attrs/blame/efcbae51cd8053c193055618cc8a6ac7880ef062/src/attr/_make.py#L1744

Here is a minimal example that shows the error:
```
!pip3 install cloudpickle attrs

import cloudpickle
import attr

@attr.s(slots=True, eq=False, weakref_slot=False)
class Foo(object):
    
    
    bar = attr.ib(type=str, default='')

meow = Foo('wow')

cloudpickle.dumps(meow)
```

And here is a notebook where you can see the crash (at the time of writing): https://colab.research.google.com/drive/1WK4EqzWyNQ3PFNSopXk-Y32IQXi8hHL9?usp=sharing

Screenshot of the error:
<img width="839" alt="Screen Shot 2021-02-03 at 2 44 34 PM" src="https://user-images.githubusercontent.com/218145/106819259-61300d80-662e-11eb-82a5-b2d3e84d9d2d.png">

I believe `pickle` for an `attrs`-based object probably works, though, since I've done that with previous versions of `attrs` ...

Can we get rid of this global variable `_already_repring` ?  The addition of it seems to address a rather odd use case ( https://github.com/python-attrs/attrs/commit/5e46afda4900766f14afb7546413de13b50a8054 ).  Perhaps `_already_repring` could be initialized on first invocation of `_make_repr()` (e.g. turn `_make_repr()` into a callable object that creates `_already_repring` on `__call__`) rather than initialized like this at package load time.  (Removing globals is probably a good thing here anyways?)

hynek:
This is a duplicate of #458 and my thoughts on it are in https://github.com/python-attrs/attrs/issues/458#issuecomment-660855504. So far the cloudpickle team hasn't reached out to us.

pwais:
[...]

PR #854, opened 2021-10-29
Explain use of global thread-local variable in repr.

thetorpedodog:
It's not completely obvious to a reader why we need a thread-local
"global" variable to track reference cycles in `__repr__` calls,
and the test does not currently contain one. This change adds a comment
explaining the need and also adds a non-attrs value in the reference
cycle of `test_infinite_recursion`.

---

This was inspired by [a bug that was opened](https://github.com/python-attrs/attrs/issues/754#issue-800763332) where the reporter couldn't see why the global variable was needed in the `__repr__` process. Initially I couldn't either, so I figured it was a good idea to add an explanation of *why* it is there—turns out it is an excellent solution to a real problem. I will gladly accept input about the nature/phrasing/placement of the commentary.

This is essentially a docs-only change (with a minor test change attached) so I don't think the full checklist applies in this case.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1864,6 +1864,14 @@ def _add_eq(cls, attrs=None):
     return cls
 
 
+# Thread-local global to track attrs instances which are already being repr'd.
+# This is needed because there is no other (thread-safe) way to pass info
+# about the instances that are already being repr'd through the call stack
+# in order to ensure we don't perform infinite recursion.
+#
+# For instance, if an instance contains a dict which contains that instance,
+# we need to know that we're already repr'ing the outside instance from within
+# the dict's repr() call.
 _already_repring = threading.local()
 
 if HAS_F_STRINGS:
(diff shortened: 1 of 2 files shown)


hynek:
Can you please fix the line lengths? See https://github.com/python-attrs/attrs/runs/4056466583?check_suite_focus=true#step:5:43

I wonder if we should maybe add a blurb to the repr argument to explain, that setting it True involves threadlocals?

thetorpedodog:
> Can you please fix the line lengths?

79 characters! My worst enemy!

> I wonder if we should maybe add a blurb to the repr argument to explain, that setting it True involves threadlocals?

I actually have an idea that may solve the cloudpickle incompatibility (and doesn’t involve any special handling on either side), which I am going to try once I get into work tomorrow. If that does the trick, I can create a new change with the fix, but if not I’ll add that note to the docs.

thetorpedodog:
> I actually have an idea that may solve the cloudpickle incompatibility (and doesn’t involve any special handling on either side), which I am going to try once I get into work tomorrow. If that does the trick, I can create a new change with the fix, but if not I’ll add that note to the docs.

Update: https://github.com/python-attrs/attrs/pull/857

hynek:
Closing thanks to #857!

PR #857, opened 2021-11-01
Pull threading.local out of the generated repr's globals.

thetorpedodog:
Generated `__repr__` methods for Pythons with f-strings included
a `threading.local` object in that method's `globals()` dictionary.
Because `cloudpickle` attempts to serialize all globals of this method,
it ends up trying to pickle the `threading.local`, which cannot be
pickled.

Instead, we now pull use of the thread-local out into its own function,
which `cloudpickle` will happily serialize. As an added benefit, this
eliminates some duplicated code between f-string and non–f-string
`__repr__`s.

Should fix:

- https://github.com/python-attrs/attrs/issues/458
- https://github.com/cloudpipe/cloudpickle/issues/320

# Pull Request Check List

    I'm not sure what the best direction to go on this is. I run this locally and see that when I cloudpickle an attrs class, it works fine. It would be pretty simple to add a test dep on cloudpickle and then create and pickle an attrs class in a new test case. But would that be too much of a special-case test case for this specific library? That said, it is more likely that changes here would cause problems in cloudpickling, rather than changes in cloudpickle causing problems here, and it would be nice to catch them faster.
- (N/A) New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/main/tests/strategies.py).
- (N/A) Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``).
- (N/A) Updated **documentation** for changed code.
- [X] Changes (and possible deprecations) have news fragments in [`changelog.d`](https://github.com/python-attrs/attrs/blob/main/changelog.d).

Another implementation idea I had would be to simply pull the `threading.local` out into another module, because that would make the global a module (which would be pickled by reference) rather than a local (which would be pickled by value). That, however, would leave essentially a two-line `import threading; repr_context = threading.local()` module, which, ehhhhhhhh

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -4,12 +4,13 @@ import copy
 import inspect
 import linecache
 import sys
-import threading
 import warnings
 
 from operator import itemgetter
 
-from . import _config, setters
+# We need to import _compat itself in addition to the _compat members to avoid
+# having the thread-local in the globals here.
+from . import _compat, _config, setters
 from ._compat import (
     HAS_F_STRINGS,
     PY2,
@@ -1864,8 +1865,6 @@ def _add_eq(cls, attrs=None):
     return cls
 
 
-_already_repring = threading.local()
-
 if HAS_F_STRINGS:
 
     def _make_repr(attrs, ns, cls):
@@ -1883,7 +1882,7 @@ if HAS_F_STRINGS:
             for name, r, _ in attr_names_with_reprs
             if r != repr
         }
-        globs["_already_repring"] = _already_repring
+        globs["_compat"] = _compat
         globs["AttributeError"] = AttributeError
         globs["NOTHING"] = NOTHING
         attribute_fragments = []
@@ -1908,24 +1907,23 @@ if HAS_F_STRINGS:
         else:
             cls_name_fragment = ns + ".{self.__class__.__name__}"
 
-        lines = []
-        lines.append("def __repr__(self):")
-        lines.append("  try:")
-        lines.append("    working_set = _already_repring.working_set")
-        lines.append("  except AttributeError:")
-        lines.append("    working_set = {id(self),}")
-        lines.append("    _already_repring.working_set = working_set")
-        lines.append("  else:")
-        lines.append("    if id(self) in working_set:")
-        lines.append("      return '...'")
-        lines.append("    else:")
-        lines.append("      working_set.add(id(self))")
-        lines.append("  try:")
-        lines.append(
-            "    return f'%s(%s)'" % (cls_name_fragment, repr_fragment)
-        )
-        lines.append("  finally:")
-        lines.append("    working_set.remove(id(self))")
+        lines = [
+            "def __repr__(self):",
+            "  try:",
+            "    already_repring = _compat.repr_context.already_repring",
+            "  except AttributeError:",
+            "    already_repring = {id(self),}",
+            "    _compat.repr_context.already_repring = already_repring",
+            "  else:",
+            "    if id(self) in already_repring:",
+            "      return '...'",
+            "    else:",
+            "      already_repring.add(id(self))",
+            "  try:",
+            "    return f'%s(%s)'" % (cls_name_fragment, repr_fragment),
+            "  finally:",
+            "    already_repring.remove(id(self))",
+        ]
 
         return _make_method(
             "__repr__", "\n".join(lines), unique_filename, globs=globs
@@ -1954,12 +1952,12 @@ else:
             Automatically created by attrs.
             """
             try:
-                working_set = _already_repring.working_set
+                already_repring = _compat.repr_context.already_repring
             except AttributeError:
-                working_set = set()
-                _already_repring.working_set = working_set
+                already_repring = set()
+                _compat.repr_context.already_repring = already_repring
 
-            if id(self) in working_set:
+            if id(self) in already_repring:
                 return "..."
             real_cls = self.__class__
             if ns is None:
@@ -1979,7 +1977,7 @@ else:
             # for the duration of this call, it's safe to depend on id(...)
             # stability, and not need to track the instance and therefore
             # worry about properties like weakref- or hash-ability.
-            working_set.add(id(self))
+            already_repring.add(id(self))
             try:
                 result = [class_name, "("]
                 first = True
@@ -1993,7 +1991,7 @@ else:
                     )
                 return "".join(result) + ")"
             finally:
-                working_set.remove(id(self))
+                already_repring.remove(id(self))
 
         return __repr__
 
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -1,0 +1,26 @@
+"""
+Tests for compatibility against other Python modules.
+"""
+
+import cloudpickle
+
+from hypothesis import given
+
+from .strategies import simple_classes
+
+
+class TestCloudpickleCompat(object):
+    """
+    Tests for compatibility with ``cloudpickle``.
+    """
+
+    @given(simple_classes())
+    def test_repr(self, cls):
+        """
+        attrs instances can be pickled and un-pickled with cloudpickle.
+        """
+        inst = cls()
+        # Exact values aren't a concern so long as neither direction
+        # raises an exception.
+        pkl = cloudpickle.dumps(inst)
+        cloudpickle.loads(pkl)
--- a/tests/test_dunders.py
+++ b/tests/test_dunders.py
@@ -367,6 +367,23 @@ class TestAddRepr(object):
         cycle.cycle = cycle
         assert "Cycle(value=7, cycle=...)" == repr(cycle)
 
+    def test_infinite_recursion_long_cycle(self):
+        """
+        A cyclic graph can pass through other non-attrs objects, and repr will
+        still emit an ellipsis and not raise an exception.
+        """
+
+        @attr.s
+        class LongCycle(object):
+            value = attr.ib(default=14)
+            cycle = attr.ib(default=None)
+
+        cycle = LongCycle()
+        # Ensure that the reference cycle passes through a non-attrs object.
+        # This demonstrates the need for a thread-local "global" ID tracker.
+        cycle.cycle = {"cycle": [cycle]}
+        assert "LongCycle(value=14, cycle={'cycle': [...]})" == repr(cycle)
+
     def test_underscores(self):
         """
         repr does not strip underscores.
--- a/changelog.d/847.change.rst
+++ b/changelog.d/847.change.rst
@@ -1,0 +1,2 @@
+Attrs classes are now fully compatible with cloudpickle.
+`#847 <https://github.com/python-attrs/attrs/issues/847>`_
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,8 @@ INSTALL_REQUIRES = []
 EXTRAS_REQUIRE = {
     "docs": ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"],
     "tests_no_zope": [
+        # For regression test to ensure cloudpickle compat doesn't break.
+        "cloudpickle",
         # 5.0 introduced toml; parallel was broken until 5.0.2
         "coverage[toml]>=5.0.2",
         "hypothesis",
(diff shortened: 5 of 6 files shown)


src/attr/_make.py
```diff
@@ -1864,7 +1865,46 @@ def _add_eq(cls, attrs=None):
...
+    If not, sets up the thread-local context to record that we are repr'ing
+    this instance.
+
+    Yields ``True`` if we're already repr'ing this class;
+    ``False`` if we are not already repr'ing it.
+    """
+    try:
+        working_set = _repr_context.working_set
+    except AttributeError:
+        # Since 'self' remains on the stack (i.e.: strongly referenced)
```
euresti, line 1892: Do you mean `inst`?

thetorpedodog: After much consideration, I have decided that I do.

Tinche:
@thetorpedodog  Could you show us the performance impact of this change on a sample class? I recommend `pyperf`.

thetorpedodog:
```
import textwrap

import attr
import pyperf

@attr.s
class Foo(object):
    bar = attr.ib()
    baz = attr.ib()
    boz = attr.ib()


runner = pyperf.Runner()
runner.timeit(
    name="basic repr",
    stmt="repr(x)",
    setup=textwrap.dedent("""
        import __main__
        x = __main__.Foo(1, "two", 3.0)
    """),
)

runner.timeit(
    name="recursion repr",
    stmt="repr(x)",
    setup=textwrap.dedent("""
        import __main__
        x = __main__.Foo(1, "two", [])
        x.boz.extend((x, x))
    """),
)
```

Benchmarks from Python 3.9.2 on Debian:

Before:

```
(venv) pfish@tiledebian:~/attrs$ python ./microbench.py 
.....................
basic repr: Mean +- std dev: 674 ns +- 10 ns
.....................
recursion repr: Mean +- std dev: 1.02 us +- 0.02 us
```

After:

```
(venv) pfish@tiledebian:~/attrs$ python ./microbench.py 
.....................
basic repr: Mean +- std dev: 673 ns +- 12 ns
.....................
recursion repr: Mean +- std dev: 1.02 us +- 0.02 us
```

Obviously this means it is 1 nsec (± 22 nsec) faster, a clear and meaningful performance improvement that is not at all probably just noise :)

thetorpedodog:
(just did a rebase to linearize history across this and #854)

Tinche:
@thetorpedodog Interesting, because that's not what I'm seeing at all ;)

`main` branch, now:
```
> pyperf timeit -g -s "from attr import make_class, attrib; A = make_class('A', ['a', 'b', 'c', 'd', 'e']); a = A(1, 2, 3, 4, 5)" "repr(a)"
.....................
 916 ns: 1 #########
 923 ns: 1 #########
 930 ns: 2 ##################
 937 ns: 2 ##################
 944 ns: 6 #####################################################
 951 ns: 6 #####################################################
 958 ns: 7 #############################################################
 965 ns: 5 ############################################
 972 ns: 9 ###############################################################################
 979 ns: 6 #####################################################
 986 ns: 7 #############################################################
 993 ns: 3 ##########################
 999 ns: 3 ##########################
1.01 us: 0 |
1.01 us: 1 #########
1.02 us: 0 |
1.03 us: 0 |
1.03 us: 0 |
1.04 us: 0 |
1.05 us: 0 |
1.05 us: 1 #########

Mean +- std dev: 970 ns +- 23 ns
```
Your branch:
```
> pyperf timeit -g -s "from attr import make_class, attrib; A = make_class('A', ['a', 'b', 'c', 'd', 'e']); a = A(1, 2, 3, 4, 5)" "repr(a)"
.....................
2.27 us: 1 ##########
2.28 us: 1 ##########
2.29 us: 0 |
2.30 us: 1 ##########
2.31 us: 1 ##########
2.32 us: 1 ##########
2.33 us: 2 ####################
2.34 us: 3 ##############################
2.35 us: 2 ####################
2.36 us: 5 #################################################
2.37 us: 5 #################################################
2.38 us: 8 ###############################################################################
2.39 us: 5 #################################################
2.41 us: 7 #####################################################################
2.42 us: 4 ########################################
2.43 us: 3 ##############################
2.44 us: 3 ##############################
2.45 us: 4 ########################################
2.46 us: 1 ##########
2.47 us: 2 ####################
2.48 us: 1 ##########

Mean +- std dev: 2.39 us +- 0.05 us
```
(CPython 3.9.7 on Ubuntu.)

That's a very significant slowdown. My gut feeling was there was no chance the proposed code would be as fast as the code that's in there now since not only are functions calls kinda expensive in CPython, you're using a context manager too.

Tinche:
Your script shows the same slowdown btw. I suspect you've forgotten to switch the branch between runs. Sorry :)

hynek:
TBH I wouldn't consider performance of repr a top priority. Isn't it usually used only in interactive use?

My gut feeling is that cloudpickle interaction is a bigger problem because it's widely used in the data science sphere. 🤔

Tinche:
I `repr` on every request, to log the request description. Why should I pay the price for a library I've never used? *shrug*

hynek:
Would it help you if we made it possible to switch off cycle detection for good? I guess that would make it even fast for you? Switching off would have to be opt-in tho.

thetorpedodog:
> I `repr` on every request, to log the request description. Why should I pay the price for a library I've never used? *shrug*

I get the concern here, but if there were code where repr was the hottest path and the thing most limiting performance, that would be a sign of rather deeper architectural problems. My feeling about having a special switch in the decorator for this is that it would be an excessively special case that would be useful only in a vanishingly small number of cases, and at that point the author would probably want to hand-write the Python themselves.

> That said, I don't understand the exact problem here. Cloudpickle doesn't like threadlocals in the globals of `__repr__`, but having `__repr__` call a function that does have a threadlocal in globals is fine for some reason?

It’s, yeah, a bit confusing. I think the deal is that cloudpickle will try to pickle everything in a class instance itself by value, but once it goes outside that class, it will pickle by reference. Or something like that.

The benefit of this is that we can limit the performance impact of this change by restructuring it: simply [pull the threadlocal into another module](https://github.com/thetorpedodog/attrs/tree/pull-it-into-a-module), then reference only that *module* in the `__repr__`’s globals. (I used the `_compat` module, because, uhhhhhh, it’s for compatibility with cloudpickle? Does that make sense? Any better ideas?) That change is sitting on another branch: https://github.com/thetorpedodog/attrs/commit/0d710b451f93b34481a62b50759576ef943e8e0c

For comparison (and this time I **didn’t** screw it up):

Old:

```plain
(a-completely-different-venv) pfish@tiledebian:~/a-completely-different-folder$ python ./microbench.py
.....................
basic repr: Mean +- std dev: 654 ns +- 16 ns
.....................
recursion repr: Mean +- std dev: 988 ns +- 23 ns
```

Thread-local in a different module (0d710b451f93b34481a62b50759576ef943e8e0c):

```plain
(venv) pfish@tiledebian:~/attrs$ python ./microbench.py
.....................
basic repr: Mean +- std dev: 682 ns +- 16 ns
.....................
recursion repr: Mean +- std dev: 1.05 us +- 0.02 us
```

Context-manager (2063781585ecec4f31c4efe6f4fc78097752bb65):

```plain
(venv) pfish@tiledebian:~/attrs$ python ./microbench.py
.....................
basic repr: Mean +- std dev: 1.67 us +- 0.05 us
.....................
recursion repr: Mean +- std dev: 3.77 us +- 0.07 us
```

I think that cost is low enough to merit not worrying about given that it fixes real user issues.

Tinche:
That does look much better. Can we add a unit test for this as well (adding cloudpickle to the dev dependencies)?

pganssle:
Sorry I pressed that button by accident.

thetorpedodog:
Switched this branch over to point at the version that pulls the thread-local into `_compat`. The previous version (context manager) is still available on my [`manage-the-context` branch](https://github.com/thetorpedodog/attrs/tree/manage-the-context).

src/attr/_make.py
```diff
@@ -4,24 +4,11 @@
...
-    PY310,
-    PYPY,
-    isclass,
-    iteritems,
-    metadata_proxy,
-    new_class,
-    ordered_dict,
-    set_closure_cell,
-)
+from . import _compat, _config, setters
```
hynek, line 11: Sorry, but this is too invasive and adds unnecessary attribute lookups. Please leave the old stuff be and just import `_compat` _additionally_ to it.

tests/test_compatibility.py
```diff
@@ -0,0 +1,23 @@
...
+from .strategies import simple_classes
+
+
+class TestCloudpickleCompat(object):
+    """
+    Tests for compatibility with ``cloudpickle``.
+    """
+
+    @given(simple_classes())
+    def test_repr(self, cls):
```
hynek, line 18:
```suggestion
    def test_repr(self, cls):
        """
        attrs instances can be pickled and un-pickled by cloudpickle.
        """
```

hynek (changes requested):
This looks great, thanks for diving into it!

Please address the commens/suggestions and 
add a news fragment: https://github.com/python-attrs/attrs/blob/main/.github/CONTRIBUTING.rst#changelog

Tinche:
LGTM!

src/attr/_make.py
thetorpedodog: Done.

tests/test_compatibility.py
thetorpedodog: Also done.

thetorpedodog (commented):
> add a news fragment

I could have sworn I had one! I must have lost it when editing history.

thetorpedodog:
Squashed everything into one change so that we have one change with a nice cohesive commit message.

Since apparently you can't just compare two arbitrary versions within GitHub, this reproduces the changes relative to https://github.com/python-attrs/attrs/commit/035280e92e8fee2538c120da435f7beb6ce8eb63:

https://github.com/thetorpedodog/attrs/commit/1bca76835b5509ffb039e8b7e92acd182506540a

- Undoes importing `_compat` module only.
- Replaces `lines.append(...)`s with just making the array in one go.
- Since the thread-local has been renamed `repr_context`, I used the `already_repring` name for the attribute on the thread-local.
- Adds changelog entry.
- Updates test compat comment in `setup.py`.
- Adds docstring to `test_repr`.

tests/test_dunders.py
```diff
@@ -364,8 +364,10 @@ class Cycle(object):
             cycle = attr.ib(default=None)
 
         cycle = Cycle()
-        cycle.cycle = cycle
-        assert "Cycle(value=7, cycle=...)" == repr(cycle)
+        # Ensure that the reference cycle passes through a non-attrs object.
+        # This demonstrates the need for a thread-local "global" ID tracker.
+        cycle.cycle = [cycle]
+        assert "Cycle(value=7, cycle=[...])" == repr(cycle)
```
hynek, line 370: ah this is a thing i forgot: is there a reason why you changed this? could we test both?

thetorpedodog: I wanted to add this test to make it obvious why we specifically need a thread-local, so that we can't use some other attrs-specific side-channel in repr (like, detecting if a `__repr__` method has an `already_repring` keyword argument and conditionally setting it), so that this worked:

```plain
repr(some_attrs_instance)
 -> calls some_attrs_instance.__repr__(secret_repring_dict)
     -> returns '...'
```

or alternately if it were less sophisticated:

```plain
repr(some_attrs_instance)
 -> detects that one of its members is `self` and subs in `...`
```

because in either of those cases, this would happen:

```plain
repr(some_attrs_instance)
 -> repr(non_attrs_instance) # no secret_repring_dict, is not `self`
     -> repr(some_attrs_instance) # the same one
         -> uh oh
```

I've kept this test the same and added a long-cycle test in addition.

hynek: perfect!

hynek (approved):
Wonderful, thank you!

hynek:
(JFTR, the news fragments don't need the link to the PR, it's done automatically from the fiile name, so I've edited it slightly: https://github.com/python-attrs/attrs/commit/11c66eff32ddf0c44cca59a9da40ddddbc2336c4  didn't want to pester you with another review round)

thetorpedodog:
> also off by 10: d884af5 😂

something like this was inevitable after I submitted #860

hynek:
Fate doesn't like to be tempted! `¯\_(ツ)_/¯`

PR #858, opened 2021-11-01
Enable use of f-strings in Python 3.6.

thetorpedodog:
CPython 3.6 has f-strings:

```
Python 3.6.13 |Anaconda, Inc.| (default, Jun  4 2021, 14:25:59)
[GCC 7.5.0] on linux
>>> f"hello"
'hello'
```

---

A quick thing I noticed. Feel free to reject if there's an assumption I'm missing (e.g. a feature of f-strings that was introduced after 3.6 that attrs uses).

Checklist: All N/A.

### DIFF
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -8,15 +8,12 @@ import warnings
 
 PY2 = sys.version_info[0] == 2
 PYPY = platform.python_implementation() == "PyPy"
-HAS_F_STRINGS = (
-    sys.version_info[:2] >= (3, 7)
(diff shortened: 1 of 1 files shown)


Tinche:
I guess you're right, must have been a brain fart?

hynek:
Thanks! Confusing when f-strings and dataclasses came to CPythonm respectively has become a national pastime!

andersk:
This seems to have somehow broken `coverage report` on Python 3.6: #895.

Issue #188, opened 2017-05-12
set_run_validators really ought to be a context manager


glyph:
`decimal.Decimal("99.999999999999999999999999999999999999999999")` percent of the time (avoiding floating-point for precision here) when you disable validators you should only be doing it for a little while.  Therefore the primary interface to disable validation should be something like:

```python
with attr.validators.disabled():
    ...
```

to avoid accidentally disabling validation for an entire program lifetime.

matthayes:
Similarly it would be useful to be able to disable automatic validation that currently runs in `__init__`, perhaps as another flag alongside `frozen` and `slots`.  I have a situation where I need to do this and I had to end up writing a separate validation method outside of `attrs`.

matthayes:
I added a PR for my suggestion: #233

hynek:
Fixed in #859!

PR #859, opened 2021-11-02
Add "no_run_validators()" context manager

sscherfke:
# Pull Request Check List

- [xx] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/main/tests/strategies.py).


sscherfke:
My first idea was to name the function `out_validators()`:
```python
with out_validators():
    ...
```
🤪

hynek:
hmmm the problem here is that it's extremely not thread-safe? 😅

But I guess @glyph asked for it already in #188 (2017!) so I guess if we document it properly…?

hynek:
I think I like his `attr.validators.disabled()` naming btw!

sscherfke:
> hmmm the problem here is that it's extremely not thread-safe? 😅

It’s not more thread-(un)safe than the underlying function `set_run_validators()` 🙃 

> I think I like his attr.validators.disabled() naming btw!

I’d have no problem with that, but `set_run_validators()` and `get_run_validators()` are defined in `_config.py` and are made available directly under the `attr` namespace.  So putting `no_run_validators()` to `validators.disable()` would be a bit inconsistent, wouldn’t it?

hynek:
> > hmmm the problem here is that it's extremely not thread-safe? 😅
> It’s not more thread-(un)safe than the underlying function `set_run_validators()` 🙃

100% correct, it's just that a context manager _looks_ like it might be so we should nail it home in the doc strings – ideally with a warning block.

> > I think I like his attr.validators.disabled() naming btw!
>
> I’d have no problem with that, but `set_run_validators()` and `get_run_validators()` are defined in `_config.py` and are made available directly under the `attr` namespace. So putting `no_run_validators()` to `validators.disable()` would be a bit inconsistent, wouldn’t it?

It would. But OTOH I don't expect anyone to use `set_run_validators()` and `get_run_validators()` after `disable()` landed. :D I wouldn't even mind to create better named aliases in the validators namespace and deprecate the old ones (without ever removing them – but not taking them with us into the attrs namespace).

sscherfke:
It has been done.

sscherfke:
Is there anything left to be done?

changelog.d/859.change.rst
```diff
@@ -0,0 +1,4 @@
+Added new context manager ``attr.validators.disabled()`` and functions ``attr.validators.(set|get)_disabled()``.
+They deprecate ``attr.(set|get)_run_validators()``.
+All functions are interoperable and modify the same internal state.
+They are not thread-safe, though.
```
hynek, line 4:
```suggestion
They are not – and never were – thread-safe, though.
```

docs/api.rst
```diff
@@ -590,6 +590,14 @@ Validators
...
         TypeError: ("'x' must be <class 'str'> (got 7 that is a <class 'int'>).", Attribute(name='x', default=NOTHING, validator=<deep_mapping validator for objects mapping <instance_of validator for type <class 'str'>> to <instance_of validator for type <class 'int'>>>, repr=True, cmp=True, hash=None, init=True, metadata=mappingproxy({}), type=None, converter=None, kw_only=False), <class 'str'>, 7)
 
+Validators can be globally disabled if you want to run them only in development and tests but not in production because you fear their performance impact:
+
+.. autofunction:: attr.validators.set_disabled
+
+.. autofunction:: attr.validators.get_disabled
+
+.. autofunction:: attr.validators.disabled
+
```
hynek, line 600: I think we should push the contextmanager as the prime API:

```suggestion
Validators can be both globally and locally disabled:

.. autofunction:: attr.validators.disabled

.. autofunction:: attr.validators.set_disabled

.. autofunction:: attr.validators.get_disabled

```

src/attr/_config.py
```diff
@@ -9,6 +9,9 @@
 def set_run_validators(run):
     """
     Set whether or not validators are run.  By default, they are run.
+
+    .. deprecated:: 21.3.0 will not be moved to new ``attrs`` namespace.
+        Use :func:`attr.validators.set_disabled()` instead.
```
hynek, line 14:
```suggestion
    .. deprecated:: 21.3.0 It will not be removed, but it also will not be
        moved to new ``attrs`` namespace. Use `attr.validators.set_disabled()`
        instead.
```

src/attr/_config.py
```diff
@@ -19,5 +22,8 @@ def set_run_validators(run):
 def get_run_validators():
     """
     Return whether or not validators are run.
+
+    .. deprecated:: 21.3.0 will not be moved to new ``attrs`` namespace.
+        Use :func:`attr.validators.get_disabled()` instead.
```
hynek, line 27:
```suggestion
    .. deprecated:: 21.3.0 It will not be removed, but it also will not be
        moved to new ``attrs`` namespace. Use `attr.validators.get_disabled()`
        instead.
```

src/attr/__init__.pyi
```diff
@@ -473,6 +474,7 @@ def evolve(inst: _T, **changes: Any) -> _T: ...
 
 def set_run_validators(run: bool) -> None: ...
 def get_run_validators() -> bool: ...
+def no_run_validators() -> ContextManager[None]: ...
```
hynek, line 477: this is outdated, also we need hints for the other two new functions

src/attr/validators.py
```diff
@@ -26,9 +31,60 @@
...
+    :type disabled: bool
+
+    .. warning::
+
+        This function is not thread-safe!
+
+    .. versionadded:: 21.3.0
+    """
+    if not isinstance(disabled, bool):
+        raise TypeError("'disabled' must be bool.")
```
hynek, line 54: I don't think there's precedents for doing runtime type checks?

tests/test_validators.py
```diff
@@ -46,6 +46,66 @@ def zope_interface():
     return zope.interface
 
 
+class TestDisableValidators(object):
+    @pytest.fixture(autouse=True)
+    def reset_default(self):
+        """
+        Make sure, validators are always enabled after a test!
```
hynek, line 53:
```suggestion
        Make sure validators are always enabled after a test.
```

tests/test_validators.py
```diff
@@ -46,6 +46,66 @@ def zope_interface():
...
+        _config._run_validators = True
+
+    def test_default(self):
+        """
+        Run validators by default.
+        """
+        assert _config._run_validators is True
+
+    @pytest.mark.parametrize("value, expected", [(True, False), (False, True)])
+    def test_set_validators_diabled(self, value, expected):
```
hynek, line 65:
```suggestion
    def test_set_validators_disabled(self, value, expected):
```

tests/test_validators.py
```diff
@@ -46,6 +46,66 @@ def zope_interface():
...
+        """
+        Sets `_run_validators`.
+        """
+        validator_module.set_disabled(value)
+        assert _config._run_validators is expected
```
hynek, line 70:
```suggestion
        validator_module.set_disabled(value)

        assert _config._run_validators is expected
```

tests/test_validators.py
```diff
@@ -46,6 +46,66 @@ def zope_interface():
...
+
+    @pytest.mark.parametrize("value, expected", [(True, False), (False, True)])
+    def test_get_run_validators(self, value, expected):
+        """
+        Returns `_run_validators`.
+        """
+        _config._run_validators = value
+        assert validator_module.get_disabled() is expected
```
hynek, line 78:
```suggestion
        _config._run_validators = value

        assert validator_module.get_disabled() is expected
```

tests/test_validators.py
```diff
@@ -46,6 +46,66 @@ def zope_interface():
...
+
+    def test_no_run_validators(self):
+        """
+        The `no_run_validators` context manager disables running validators,
+        but only within its context.
+        """
+        assert _config._run_validators is True
+        with validator_module.disabled():
+            assert _config._run_validators is False
+        assert _config._run_validators is True
```
hynek, line 95:
```suggestion
    def test_with_disabled(self):
        """
        The `disabled` context manager disables running validators,
        but only within its context.
        """
        assert _config._run_validators is True
        
        with validator_module.disabled():
            assert _config._run_validators is False
            
        assert _config._run_validators is True
```

tests/test_validators.py
```diff
@@ -46,6 +46,66 @@ def zope_interface():
...
+    def test_no_run_validators_with_errors(self):
+        """
+        Running validators is re-enabled even if an error is raised.
+        """
+        assert _config._run_validators is True
+        with pytest.raises(ValueError):
+            with validator_module.disabled():
+                assert _config._run_validators is False
+                raise ValueError("haha!")
+        assert _config._run_validators is True
```
hynek, line 106:
```suggestion
    def test_with_disabled_with_errors(self):
        """
        Running validators is re-enabled even if an error is raised.
        """
        assert _config._run_validators is True

        with pytest.raises(ValueError):
            with validator_module.disabled():
                assert _config._run_validators is False

                raise ValueError("haha!")

        assert _config._run_validators is True
```

hynek (changes requested):
- a bunch of suggestions
- we need types hints! :)
   - also please typing_example.py
-  add a news fragment please: https://github.com/python-attrs/attrs/blob/main/.github/CONTRIBUTING.rst#changelog

sscherfke:
Done.  Not sure, what kind of example you think of for the typing examples, though.

sscherfke:
"Just a few docs and typing changes.  What could go wrong", he thought.


Gonna fix it tomorrow.

hynek:
> Not sure, what kind of example you think of for the typing examples, though.

Just use the API and see if MyPy complains. :)

sscherfke:
Okay, fixed it.

hynek (approved):
Thanks!

I've done some minor fixes myself. The gh cli is really neat.

PR #861, opened 2021-11-03
WIP: Allow evolve() to recursively evolve attrs classes

### LINKED ISSUE #634: Recursive attr.evolve
In the same manner as `attr.asdict(..., recursive=True)` exists, I have a use case for a similar option to `attr.evolve()`. I want all attr instances evolved, but with no need to deepcopy any non-attr attributes. I ended up implementing the code below. Could this be a consideration for addition into `python-attrs`?

```python
def evolve_recursive(inst, **changes):
    """ Recursive attr.evolve() method, where any attr-based attributes
        will be evolved too.
    """
    cls = inst.__class__
    attrs = attr.fields(cls)
    for a in attrs:
        if not a.init:
            continue
        init_name = attr_name if attr_name[0] != "_" else attr_name[1:]
        if init_name not in changes:
            value = getattr(inst, attr_name)

            # New code
            if attr.has(value.__class__):
                value = evolve_recursive(value)

            changes[init_name] = value
    return cls(**changes)
```

sscherfke:
This is a new atempt to fix #634 (#759 failed: #804, #806)

# Pull Request Check List

### DIFF
--- a/changelog.d/861.change.rst
+++ b/changelog.d/861.change.rst
@@ -1,0 +1,1 @@
+``attrs.evolve()`` now works recursively with nested ``attrs`` classes.
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -623,6 +623,27 @@ In Clojure that function is called `assoc <https://clojuredocs.org/clojure.core/
    >>> i1 == i2
    False
 
+This functions also works for nested ``attrs`` classes.
+Pass a (possibly nested) dictionary with changes for an attribute:
+
+.. doctest::
+
+   >>> @attr.s(frozen=True)
+   ... class Child(object):
+   ...     x = attr.ib()
+   ...     y = attr.ib()
+   >>> @attr.s(frozen=True)
+   ... class Parent(object):
+   ...     child = attr.ib(type=Child)  # Type annotations / passing "type" is important!
+   >>> i1 = Parent(Child(1, 2))
+   >>> i1
+   Parent(child=Child(x=1, y=2))
+   >>> i2 = attr.evolve(i1, child={"y": 3})
+   >>> i2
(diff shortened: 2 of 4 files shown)


sscherfke:
Oh, Pyhton 2! 😿

sscherfke:
What do you think, @hynek?  Time for another try?

This time, recursive evolving works a bit more defensively – only if an attribute explicitly annotates a type.  And we have a few more tests.  I’m not sure though if this is enough. 🤷

hynek:
This probably doesn't work with string type annotations?

My gut feeling is that it's too much magic and too many sharp edges. I wonder if we introduced a new API, how this case could be made water-proof?

sscherfke:
Yes, stringy type annotations do break, of course.

I’ll play around with it in typed settings for a while.  It would be very useful there for monkeypatching nested frozen settings classes in tests.

jacobg:
The following unit test fails. Not sure if your intent was to support settings sparse values.

```python
def test_evolve_recursive_deep():

    @attr.s(auto_attribs=True, frozen=True, slots=True)
    class Baz:
        a: str
        b: str

    @attr.s(auto_attribs=True, frozen=True, slots=True)
    class Bar:
        baz: Baz

    @attr.s(auto_attribs=True, frozen=True, slots=True)
    class Foo:
        bar: Bar

    foo = Foo(bar=Bar(baz=Baz(a='a1', b='b1')))

    assert evolve_recursive(foo, bar={'baz': {'a': 'a2'}}) ==\
           Foo(bar=Bar(baz=Baz(a='a2', b='b1')))

```

hynek:
Closing for staleness; feel free to open a new one.

Issue #726, opened 2020-12-02
Request: one line to explain what `s` and `ib` stand for in the API docs


codekiln:
To better honor the spirit of [PEP 20's _Explicit is better than implicit_](https://www.python.org/dev/peps/pep-0020/) and the library's [_No surprises_ philosophy](https://www.attrs.org/en/stable/overview.html#philosophy), consider adding something like this in or immediately after _the first line_ of the API docs for {[`attr.s`](https://www.attrs.org/en/stable/api.html?highlight=ib#attr.s), [`attr.ib`](https://www.attrs.org/en/stable/api.html?highlight=ib#attr.ib)}: "Note: {`.s`, `.ib`} is not an abbreviation; {`attr.s`, `attr.ib`} is meant to look like {`attrs`, `attrib`}. See the [_On the `attr.s` and `attr.ib` Names_ section in the Overview for more information](https://www.attrs.org/en/stable/overview.html#on-the-attr-s-and-attr-ib-names)."

I'm probably not the only one to see the `Note` included in those API doc entries (way beneath the changelog) and wonder why the library authors chose to leave abbreviation expansion as an exercise left to potential library adopters, having not seen the Overview note.

cipri-tom:
I agree, the first time I looked at attrs it was quite uncomfortable -- I kept wondering "what's an `s` and an `ib` ?" :D

hynek:
Well, there's https://www.attrs.org/en/stable/overview.html#on-the-attr-s-and-attr-ib-names and 
![Screenshot 2020-12-03 at 06 20 34](https://user-images.githubusercontent.com/41240/100967558-c765ba80-352f-11eb-85bf-4fcf589fad2d.png) on the homepage mentions that it's in there?

I could elaborate a bit more, that the original reason for those short terms is basically, that before type annotation, you had to write it out every, single, time. Since that's not the case anymore, [we're moving away from them as the main API](https://www.attrs.org/en/stable/api.html#provisional-apis).

cipri-tom:
hah, I had missed that bullet point 😅. I shall read docs more in-depth before starting to use something.

Well, since they are superseded in the new API, I guess it's OK to leave as it is for now, there's enough explanation for newcomers.

codekiln:
The request above is to make a hyperlink to that explanation in the API docs.

hynek:
I hope all concerns have been adequately fixed by #871.

PR #863, opened 2021-11-11
Switch to NG APIs in docs

Tinche:
Here's what I'm thinking.

Switch `@attr.s` to a bare `@define`. This includes also doing `from attr import define`. I consider `@define` to be as cleanest as we can get, and I love it.

Remove inheriting from `object` pretty much everywhere. Use modern type annotations whereever possible (so `list[int]` instead of `List[int]`). Essentially use all the affordances of modern Python to make the examples as clean as possible.

Avoid `= field()/= attr.ib()` whereever possible, while still leaving it in a few places. Again, the syntax with just the type annotation is extremely clean and will probably help sell this to new folks. Prefer `field` to `attr.ib`.

I haven't gone through all the files yet, this is just the start.


README.rst
```diff
@@ -35,13 +35,12 @@ For that, it gives you a class decorator and a way to declaratively define the a
...
-   >>> from typing import List
-   >>> import attr
+   >>> from attr import define, Factory
 
-   >>> @attr.define
+   >>> @define
    ... class SomeClass:
    ...     a_number: int = 42
-   ...     list_of_numbers: List[int] = attr.Factory(list)
+   ...     list_of_numbers: list[int] = Factory(list)
```
euresti, line 43: I'm nervous about `list[int]` because that only works in 3.9+ and I can imagine lots of people in a lower version.  Though I'm not sure.  I could've sworn there was a chart of which versions people run but I can't find it now.

docs/comparison.rst
```diff
@@ -29,15 +31,19 @@ It is then used as a key function like you may know from `sorted`:
...
+
    >>> S("foo") == S("FOO")
    True
-   >>> @attr.s
-   ... class C(object):
-   ...     x = attr.ib(order=int)
+
+   >>> @define
+   ... class C:
+   ...     x: str = attr.ib(order=int)
```
euresti, line 45: Did you mean to leave this one?

README.rst
hynek: I tried `list` but we can't because RTD is still on old Python and doctest doesn't support future imports. :(

hynek: (I have just moved `main` to 3.8, but it's still not gonna work. I'm afraid we'd need 3.10)

Tinche: Aight, we can move back to `List` for the time being.

docs/comparison.rst
Tinche: Nope, missed it. Thanks!

Tinche:
Did some more converting. A good next step would be to say that type annotations are completely optional, can be replaced by `field` or annotating with `Any`, and we don't do anything with them by default. I would also add that they are very recommended: mypy supports them, and even if a person is not using mypy, they still serve as documentation, can have uses to companion libraries and help with editor autocomplete.

hynek:
> Did some more converting. A good next step would be to say that type annotations are completely optional, can be replaced by `field` or annotating with `Any`, and we don't do anything with them by default. I would also add that they are very recommended: mypy supports them, and even if a person is not using mypy, they still serve as documentation, can have uses to companion libraries and help with editor autocomplete.

I would push back a little on the last part: _unchecked_ type annotations are potentially worse than no type annotations. The number of packages I had to block in mypy.ini is growing every day and even in applications an unchecked type annotation is akin to a comment: it means nothing and can be even confusing.

The first part we could probably put into `types.rst`.

We _need_ to write a chapter on the names tho, but that's something I can do when you're done or even in parallel. Effectively explain the various types and some historic context. Closing #726 in the process.

Tinche:
> I would push back a little on the last part: _unchecked_ type annotations are potentially worse than no type annotations. The number of packages I had to block in mypy.ini is growing every day and even in applications an unchecked type annotation is akin to a comment: it means nothing and can be even confusing.

Interesting. We had been using type annotations for about 3 years without ever using Mypy, and I don't mean only in classes; function arguments too. We found it better than nothing, and would use the editor hover popup as a poor man's Mypy :)) But I appreciate your viewpoint too. We are not a library, but an end consumer of them.

Tinche:
How do I fix ```init.rst:196:more than one target found for 'any' cross-reference 'frozen': could be :py:func:`attr.setters.frozen` or :py:func:`attr.frozen```?

docs/init.rst
```diff
@@ -193,12 +194,12 @@ Since the validators runs *after* the instance is initialized, you can refer to
 This example also shows of some syntactic sugar for using the `attr.validators.and_` validator: if you pass a list, all validators have to pass.
 
 ``attrs`` won't intercept your changes to those attributes but you can always call `attr.validate` on any instance to verify that it's still valid:
+When using `define` or `frozen`, ``attrs`` will run the validators even when setting the attribute.
```
hynek, line 197:
```suggestion
When using `define` or :func:`~attr.frozen`, ``attrs`` will run the validators even when setting the attribute.
```

Tinche: I see, great. Single backticks has a special meaning where it attempts to actually create a link?

hynek: Single backticks together with our default_role means basically "guess, mofo"

See: #571

hynek: (I could've used `attr.frozen`, but then I wouldn't be able to use `~`)

hynek (changes requested):
- Hashing could use some `attr.s` replacements in copy.
- some smaller stuff
- otherwise promising

thanks for doing god's work here Tin!

docs/examples.rst
```diff
@@ -123,25 +123,9 @@ This is useful in times when you want to enhance classes that are not yours (nic
...
-   ... class C(object):
-   ...     @attr.s(repr_ns="C")
-   ...     class D(object):
-   ...         pass
-   >>> C.D()
-   C.D()
-
-``repr_ns`` works on both Python 2 and 3.
-On Python 3 it overrides the implicit detection.
+Slot classes, which are default for the new APIs, don't play well with multiple inheritance so we don't use them in the example.
```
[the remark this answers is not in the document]
Tinche: I bow to your Sphinx skills

docs/init.rst
```diff
@@ -83,13 +84,13 @@ This is when default values come into play:
...
+   >>> from attr import define, field, Factory
+   >>> @define
+   ... class C:
+   ...     a: int = 42
+   ...     b: list = field(factory=list)
+   ...     c: list = Factory(list)  # syntactic sugar for above
+   ...     d: dict = field()
    ...     @d.default
    ...     def _any_name_except_a_name_of_an_attribute(self):
    ...        return {}
```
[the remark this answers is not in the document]
Tinche: I just removed it.

hynek (approved):
Welcome to the future zone!

PR #866, opened 2021-11-15
Title case for to_bool strings

hkclark:
I really like the new `to_bool` converter.  I was thinking it would be good to support some "title case" variations on strings... especially since Python folks might try to use the strings `True` and `False`.  

I did think about modifying the code to do include additional `if` checks inside the `try` block to force `str.lower()` if `isinstance(val, str)` so that we end up handling things like `TRUE`, `TrUe`, etc.... but for now I thought I would proposed the "Keep It Simple" approach since I think "True" is going to be more common than "TRUE" (and things like "TrUe" should be very rare).

I don't think this requires updates to the `versionchanged` info because 21.3.0 hasn't be released yet, but let me know if I have that wrong.

Regards,
HKC

# Pull Request Check List

### DIFF
--- a/src/attr/converters.py
+++ b/src/attr/converters.py
@@ -118,18 +118,18 @@ def to_bool(val):
     Values mapping to :code:`True`:
 
     - :code:`True`
-    - :code:`"true"` / :code:`"t"`
-    - :code:`"yes"` / :code:`"y"`
-    - :code:`"on"`
+    - :code:`"true"` / :code:`"t"` / :code:`"True"` / :code:`"T"`
+    - :code:`"yes"` / :code:`"y"` / :code:`"Yes"` / :code:`"T"`
+    - :code:`"on"` / :code:`"On"`
     - :code:`"1"`
     - :code:`1`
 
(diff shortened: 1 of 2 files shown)


src/attr/converters.py
```diff
@@ -139,8 +139,8 @@ def to_bool(val):
     """
     if isinstance(val, str):
         val = val.lower()
-    truthy = {True, "true", "t", "yes", "y", "on", "1", 1}
-    falsy = {False, "false", "f", "no", "n", "off", "0", 0}
+    truthy = {True, "true", "True", "t", "T", "yes", "Yes", "y", "Y", "on", "On", "1", 1}
```
euresti, line 142: Looking at the code it looks like val should already be a lowercase string.  Am I reading this wrong?

hynek: Yeah. The tests should pass without the changes IMO.

hkclark:
@euresti & @hynek Oh wow... how did I miss that!  Very silly of me and very embarrassing.  I apologize for the bogus PR!  Closing.

PR #872, opened 2021-11-24
Update comparisons to dataclasses

### LINKED ISSUE #761: What distinguishes attrs from PEP 557 dataclasses?
Since we have `@dataclass` in our batteries, a document distinguishing `attrs` from PEP-557 dataclasses would be good; possibly enumerating differences, features available in both, only attrs, only dataclasses, etc.

https://www.python.org/dev/peps/pep-0557/

hynek:
This is to fix #761

There's a bunch of problems to discuss, so I'm just gonna start _somewhere_.

The big questions to me are:

- do we still have to talk about dicts/tuples at all?
- should we add pydantic? (cf. <https://threeofwands.com/why-i-use-attrs-instead-of-pydantic/>)
- **how** are we going to talk about dataclasses at all, since they're something else in each Python release?
   - [x] which definitely is an upside of attrs that should be mentioned.

# Relevant rendered pages

- https://attrs--872.org.readthedocs.build/en/872/overview.html
- https://attrs--872.org.readthedocs.build/en/872/why.html

### DIFF
--- a/docs/why.rst
+++ b/docs/why.rst
@@ -5,51 +5,46 @@ Why not…
 If you'd like third party's account why ``attrs`` is great, have a look at Glyph's `The One Python Library Everyone Needs <https://glyph.twistedmatrix.com/2016/08/attrs.html>`_!
 
 
-…tuples?
---------
+…Data Classes?
+--------------
+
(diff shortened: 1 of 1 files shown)


hynek:
To be clear: I'm asking for bikeshedding here.

PR #877, opened 2021-11-29
Add support for re.Pattern to validators.matches_re()

### LINKED ISSUE #875: matches_re() should accept re.Pattern in addition to str
`attr.validators.matches_re()` does not accept precompiled regular expressions, and instead requires a `str` pattern (and optional flags).

this is not ideal when using existing compiled regular expressions, which already can have flags etc.

it would be a nice addition if the same helper accepts `re.Pattern` (py37+) or `typing.Pattern` for those living in the past.

open questions:

- would `flags=` be allowed if a `re.Pattern` is passed? compiled patterns cannot be ‘reinterpreted’ safely with different flags (e.g. `re.VERBOSE` changes many semantics). that means options are a) reject by raising an exception or b) ignore. my preference is raising.
- in general any reason not to do this?

wbolster:
Fixes #875.

This adds support for pre-compiled regular expressions to
attr.validators.matches_re(). This cannot be combined with flags since
the pre-compiled pattern already has those.

Detailed changes:

- attr.validators.matches_re() now accepts re.Pattern in addition to
  strings; update type annotations accordingly.
- Convert percent-formatting into str.format() for an error message
- Simplify (private) _MatchesReValidator helper class a bit: use the
  actual compiled pattern, and drop the unused .flags attribute.
- Simplify control flow a bit; add pointer about fullmatch emulation.
- Add tests
- Tweak existing test to ensure that .fullmatch() actually works
  correctly by matching a pattern that also matches (but not
  ‘full-matches’) a shorter substring.

# Pull Request Check List

### DIFF
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -14,6 +14,12 @@ from ._make import _AndValidator, and_, attrib, attrs
 from .exceptions import NotCallableError
 
 
+try:
+    Pattern = re.Pattern
+except AttributeError:  # Python <3.7 lacks a Pattern type.
+    Pattern = type(re.compile(""))
+
+
 __all__ = [
     "and_",
     "deep_iterable",
@@ -129,8 +135,7 @@ def instance_of(type):
 
 @attrs(repr=False, frozen=True, slots=True)
 class _MatchesReValidator(object):
-    regex = attrib()
-    flags = attrib()
+    pattern = attrib()
     match_func = attrib()
 
     def __call__(self, inst, attr, value):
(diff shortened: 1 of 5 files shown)


src/attr/validators.py
```diff
@@ -159,7 +164,7 @@ def matches_re(regex, flags=0, func=None):
     A validator that raises `ValueError` if the initializer is called
     with a string that doesn't match *regex*.
 
-    :param str regex: a regex string to match against
+    :param str regex: a regex string or precompiled pattern to match against
```
hynek, line 167: The param type needs to be updated too

wbolster: ack, dropped it

tests/typing_example.py
```diff
@@ -167,7 +167,7 @@ class Validated:
             attr.validators.instance_of(C), attr.validators.instance_of(D)
         ),
     )
-    e: str = attr.ib(validator=attr.validators.matches_re(r"foo"))
+    e: str = attr.ib(validator=attr.validators.matches_re(re.compile(r"foo")))
```
hynek, line 170: I think we should keep both, but I'll do it myself to save ourselves another review round.

hynek (approved):

hynek:
Thanks!

tests/typing_example.py
hynek: ah nevermind, `f` has a str :)

wbolster: yes, couldn't remember the alphabet and i knew you would complain about `e2` 😛

PR #878, opened 2021-11-29
Avoid whitelist/blacklist terminology

wbolster:
The code already uses include() and exclude(), so simply tweak the docstrings.


hynek (approved):

hynek:
Thanks! I remember checking APIs but apparently  not docstrings.

PR #885, opened 2021-12-11
Modernize docs some more

Tinche:
More more doc modernization.

I have come across a problem though.

```
from attr import define, fields, resolve_types


@define
class A:
    a: list["A"]
    b: "B"


@define
class B:
    a: A


print(fields(A).a.type)
fields(A).b.type
resolve_types(A, globals(), locals())
print(fields(A).a.type)
fields(A).b.type
```
This is from the docs but it doesn't actually work. `typing.List["A"]` will produce a ForwardReference by itself, but `list["A"]` won't, so `typing.get_annotations` doesn't process it. What a nasty surprise.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -531,7 +531,7 @@ This will replace the *type* attribute in the respective fields.
 
     >>> @define
     ... class A:
-    ...     a: list['A']
+    ...     a: 'list[A]'
     ...     b: 'B'
     ...
     >>> @define
@@ -539,7 +539,7 @@ This will replace the *type* attribute in the respective fields.
     ...     a: A
     ...
     >>> fields(A).a.type
-    list['A']
+    'list[A]'
     >>> fields(A).b.type
     'B'
     >>> resolve_types(A, globals(), locals())
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -551,10 +551,8 @@ This will replace the *type* attribute in the respective fields.
 
 .. note::
 
-   If you find yourself using string type annotations to handle forward references,
-   wrap the entire type annotation in quotes instead of only the type you need
-   a forward reference to (so ``'list[A]'`` instead of ``list['A']``). This
-   is a limitation of the Python typing system.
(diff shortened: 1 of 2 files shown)


hynek:
So uh what should we do about it?

Tinche:
Well I can do one of the following:
* change the example back to typing.List (ugly)
* change the example to use ForwardRef explicitly (ugly)

I'm just a little stunned. Is this a Python bug? https://www.python.org/dev/peps/pep-0585/ makes no mention of these differences in ForwardRefs. Should we report this?

Tinche:
My gut feeling is when Łukasz wrote PEP 585 he was assuming PEP 563 would be active. This is not really the case just yet, since 563 is essentially being replaced from what I understand, and there's no implementation of its replacement yet. So old Python versions are kinda screwed.

hynek:
Well, why rely on gut feelings, when @ambv can just tell us? ;)

Tinche:
Got a suggestion from @ambv on Twitter, so I went with:
* wrap the entire expression in quotes

hynek:
Given how it surprised you, I think we really should explain it?

hynek:
ping?

Tinche:
Added a note!

docs/examples.rst
```diff
@@ -522,35 +521,41 @@ If you don't mind annotating *all* attributes, you can even drop the `field` and
...
+    list[A]
     >>> fields(A).b.type
     <class 'B'>
 
+.. note::
+
+   If you find yourself using string type annotations to handle forward references,
+   wrap the entire type annotation in quotes instead of only the type you need
+   a forward reference to (so ``'list[A]'`` instead of ``list['A']``). This
+   is a limitation of the Python typing system.
```
hynek, line 557: We use [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) fine sir!

```suggestion
   If you find yourself using string type annotations to handle forward references, wrap the entire type annotation in quotes instead of only the type you need a forward reference to (so ``'list[A]'`` instead of ``list['A']``).
   This is a limitation of the Python typing system.
```

hynek (approved):
thanks!

PR #886, opened 2021-12-13
NG: convert on setattr by default

### LINKED ISSUE #835: Clarification - Converters vs on_setattr / Convert automatically when setting
attrs version: 21.2.0

Using the documentation [on Converters](https://www.attrs.org/en/stable/init.html#converters) it was not immediately clear that converters only work on instantiation of a new attrs-enhanced class. (Or I totally missed something.) 

1. Is this the intended function? That the type conversion only happens on class instantiation and not when setting an attribute? (e.g. ```o.x = "2" ``` in the example in the documentation)
    * If yes: Perhaps the documentation could be clarified to explain this and also provide a clear example of using ```on_setattr``` to also run the converter
    * If no: Perhaps there's a bug because setting an attribute when a converter is specified does not work

And further: from a logical design standpoint, it seems that if you're specifying a converter to be run at class instantiation [which means *all* values on *new* instances are run through that converter....] what circumstances would you *not* want that to run when setting attributes? Part of the purpose of converters is to ensure consistent typing and normalization (from the documentation: _"Finally, sometimes you may want to normalize the values coming in. For that attrs comes with converters."_). Given that, it seems strange that you could break that normalization by simply setting an attribute.

TL;DR: Converters only run on class instantiation. Shouldn't they be run when setting an attribute as well?

hynek:
Not doing that from the get-go was an oversight.

Are we OK with this?

Fixes #835


hynek:
Yuck, looks like our doctests saved us from a performance regression.

This is gonna be more work than it looked like.

hynek:
Looks like we've needlessly added a cached setattr to all classes that were derived from a slotted one (instead of just frozen ones). The setattr wasn't used, but it was a waste of space.

Tinche (approved):

PR #887, opened 2021-12-15
Add attrs namespace

### LINKED ISSUE #487: [RFC] Inconvenient defaults?

### LINKED ISSUE #705: Better defaults for asdict/astuple
- Would it be acceptable to change the default value for `retain_collection_types` to `True`? This would reduce the likelihood of this issue to appear unexpectedly. It is not a full fix though.

- Would it be ok to use `tuple` as default collection type, when `retain_collection_types=False`? I would expect this to completely resolve the issue.

_Originally posted by @tovrstra in https://github.com/python-attrs/attrs/issues/646#issuecomment-682167908_

hynek:
# Summary

ROCK’N’ROLL

Fixes #487, fixes #705,

I don't think it's practical to already update all the docs to the new namespace. Especially since it's provisional for now.

# Pull Request Check List


src/attrs/__init__.pyi
```diff
@@ -0,0 +1,63 @@
...
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+)
+
+# Because we need to type our own stuff, we have to make everything from
+# attr explicitly public too.
+from attr import __author__ as __author__
```
hynek, line 14: @euresti is there a better way to deal with this stuff maybe?

wbolster: would `__all__` help perhaps? some tools (flake8 afaik) look at it and will not warn about ‘unused imports’ for anything appearing there. not sure if similar things are in place for tools (mypy etc) looking at .pyi files.

hynek: help with what exactly? i'd like to elide that whole block :)

the problem is that i need to have type hints for asdict/astuple and with that I can't use tacitly those from attr/__init__.pyi and have to import them here

src/attr/__init__.pyi
```diff
@@ -447,6 +447,7 @@ def make_class(
 # these:
 # https://github.com/python/mypy/issues/4236
 # https://github.com/python/typing/issues/253
+# XXX: remember to fix attrs.asdict/astuple too!
```
wbolster, line 450: is this for now or for later?

hynek: later – its referring to the changes to mypy/typing that once they're done, we need to fix both places :)

src/attrs/__init__.pyi
wbolster: ack, i'm missing some context here. i was referring to the `x as x` repetition

src/attr/__init__.pyi
wbolster: ack, i wondered since both links point to _closed_ issues

Tinche:
If `attrs` is the future, shouldn't we move all the code over there and import it in `attr`?

hynek:
> If `attrs` is the future, shouldn't we move all the code over there and import it in `attr`?

Technically yes, but that would be a breaking change whose blast radius I can't fathom.

This way, even MyPy keeps working out of the box.

src/attr/__init__.pyi
hynek: I hope sOmEoNe tackles it then 😇

src/attrs/__init__.pyi
hynek: according to `attr/__init__.pyi` it’s necessary to make them public and in my experiments it still seems to hold true 🤔

Tinche:
```
>>> from attrs.converters import optional
ModuleNotFoundError: No module named 'attrs.converters'
```

hynek:
ugh crap, I guess we have to mirror all packages into the new namespace?

hynek:
fe3c53e is ugly but should do the job?

wbolster:
> [fe3c53e](https://github.com/python-attrs/attrs/commit/fe3c53eda43346f205145193f5ad4b871f60abfb) is ugly but should do the job?

fwiw, mypy is generally (and rightfully) not happy about `*` imports

hynek:
I played with `reveal_type` and it _seems_ to work? Did they improve or am I missing something?

wbolster:
ah ok, i think the separate `.pyi` file makes it work in this case.

in general, i don't think it works well with (modern) embedded annotations directly in the source code.

hynek:
I'm open to suggestions for improvement, but one more repetition would mean one more place we could forget about. :-/

euresti:
Do you need the pyi file or would mypy pick up the types from the py file?

hynek:
We need overloads, so I'm reasonably sure we need `pyi` files?

wbolster:
imho, as a mostly bystander and casual contributor (aka ignore me), doing these things at the same time would be perfect timing: 🙃

- introducing new package name `attrs`
- becoming py37+ only (only supported pythons from next week)

…but i suspect you may hold different opinions on that, based on [py2 notes in the docs](https://www.attrs.org/en/stable/python-2.html)

hynek:
While that would solve other pain points, overloads are AFAIK something that you need stub files for still. :)

wbolster:
> We need overloads, so I'm reasonably sure we need `pyi` files?

overloads work fine in source code, see e.g.
https://github.com/wbolster/jsonlines/blob/master/jsonlines/jsonlines.py

euresti:
Yeah. We could definitely move the stubs into the py files.  Though attrs has quite a bit of "fake" types that we wouldn't want mypy checking.  It could be an interesting project.

euresti:
Is the expectation that `import attr` is frozen and will not be receiving any new features that `import attrs` might be?  Or is anything available in `attrs` gonna be available in `attr`?

wbolster:
> Though attrs has quite a bit of "fake" types that we wouldn't want mypy checking.

`@typing.type_check_only` may help here: [docs](https://docs.python.org/3/library/typing.html#typing.type_check_only)

hynek:
This all sounds very promising, but this diff is already way too big. Can I somehow convince y'all to help me with inline types for the next release?

attrs 2.7 downloads seem to be down to 2.22% so I might change my mind soon-ish:

| category | percent |  downloads |
|:---------|--------:|-----------:|
| 3.7      |  40.10% | 19,465,561 |
| 3.8      |  21.25% | 10,315,653 |
| 3.6      |  12.75% |  6,189,331 |
| null     |  10.43% |  5,063,048 |
| 3.9      |   9.15% |  4,442,703 |
| 2.7      |   2.22% |  1,079,185 |
| 3.10     |   2.06% |    998,378 |
| 3.5      |   2.01% |    974,174 |
| 3.4      |   0.03% |     13,247 |
| 3.11     |   0.01% |      3,357 |
| 2.6      |   0.00% |          8 |
| 3.3      |   0.00% |          6 |
| 3.2      |   0.00% |          4 |
| Total    |         | 48,544,655 |

Date range: 2021-12-01 - 2021-12-20

Tinche:
I'm going to follow attrs lead in cattrs (big surprise), and have a PR ready. What do I say in the changelog though? Are we recommending new code to import from attrs/cattrs instead? Do we ever plan on removing attr/cattr?

hynek:
Well, you can see what I've said in https://github.com/python-attrs/attrs/pull/887/files#diff-826ab996c410e75f5930990d312d073db10300312fc9fabe2f01fb327a007bde

I'm leaving it provisional for one release, but I think changing the docs to primarily use the new APIs is a strong indicator. ;)

Since we had almost 90 million downloads per month with `attr` in Place, I don't remove it, ever. I think I'll eventually switch the import direction but that's it. 🤔

hynek:
I don't think I can make anyone review a +830/-368 PR so here we go and let's hope for the best. Merry Xmas to me. 🤪

docs/api.rst
```diff
@@ -341,99 +440,99 @@ Validators can be globally disabled if you want to run them only in development
 Validators
 ----------
 
-``attrs`` comes with some common validators in the ``attrs.validators`` module:
+``attrs`` comes with some common validators in the ``attrs.validators`` module.
+All objects from ``attrs.converters`` are also available from ``attr.converters``.
```
sanjioh, line 444: @hynek Hi! I just wanted to let you know that this line about converters does not probably belong here (it looks like a duplicate of a line down below). Thanks for attrs and for its wonderful docs!

hynek: thanks, fixed in e188af7bc4df99eb84772b0ad0f80c4e11f740b4!

PR #890, opened 2021-12-17
Mypy improvements

Tinche:
Not to be merged yet.

The problem is the Mypy plugin doesn't return a namedtuple, but an ordinary tuple. We should fix this in the Mypy plugin.


hynek:
Is this something that should remain open?

Tinche:
Yeah, we need to merge this when Mypy release 0.940 I believe.

hynek:
Maybe we should get it to pass CI until then. 😇

Tinche:
I can't until Mypy releases :p

hynek:
eg ok, your phrasing sounded like it's needed, once mypy is out

Tinche:
It's just more sophisticated type stubs, we don't have to match their release exactly. But we might want to.

hkclark:
@Tinche I noticed recently that using attrs 21.4.0 and the new `attrs` namespace didn't seem to be supported by mypy.  I was initially thinking this PR might add it, but in taking a quick look at the changes, it doesn't seem so (but I'm not expert on mypy type syntax).  I was just curious if there was a plan to get support for the `attrs` namespace or if I was just doing something wrong on my end. :-)  Thanks for your work on this!

hkclark:
> > I noticed recently that using attrs 21.4.0 and the new `attrs` namespace didn't seem to be supported by mypy.
> 
> The new attrs namespace is absolutely supported by mypy. I use it all the time.
> 
> We've achieved it by re-importing symbols from the `attr` namespace and mypy can resolve that.

Hmm, OK, thanks for the info!  I guess let me convert this project at work back to the attrs namespace and see if I can figure out what's going on. :-)

While I have you, I assume it's expected that mypy cannot handle "fancy converters" -- I get `error: Unsupported converter, only named functions and types are currently supported` if I do things like: a) `converter=lambda value: weakref.proxy(value)` or `converter=[...]` or `converter=attrs.validtors.pipe(...)`.

hkclark:
@hynek To follow up on the issues I mentioned in this project I'm working on, when I use the `attrs` namespace for everything I get this from mypy for my converters:

```
error: Module has no attribute "converters"  [attr-defined]
```

hynek:
Unfortunately, mypy support for converters is very bare-bones. There's not much _we_ can do about it. Your error must have some other cause. It's explicitly tested here:

https://github.com/python-attrs/attrs/blob/7804a68ee3dd2c2a0302d482237c51ce975ef17f/tests/typing_example.py#L414

hkclark:
> Unfortunately, mypy support for converters is very bare-bones. There's not much _we_ can do about it. Your error must have some other cause. It's explicitly tested here:
> 
> https://github.com/python-attrs/attrs/blob/7804a68ee3dd2c2a0302d482237c51ce975ef17f/tests/typing_example.py#L414

@hynek OK, thank you.  I'll study it more and if I can't find a cause on my side, I'll try to open a ticket with very bare-bones example code that recreates it.  And maybe I'll open a PR that adds a comment to the docs re converters being limited just so others know about the limitations.  Thank you again!

Tinche:
@hynek this should go in now. I guess I need to rebase though.

hynek:
@Tinche this is breaking even on 3.6…wanna wait until this stuff is fixed in mypy? :-/

hkclark:
> @hkclark turns out you were right and our tests broken: #930 + #931
> 
> Sorry for that!

@hynek No problem at all.  Thanks for getting back to me.  I'm just glad to know I wasn't going totally crazy. :-)  Thank you again for your work on attrs... I really like it and look forward to ongoing new features.

Tinche:
Mergy mergy now?

hynek:
Allmost all of CI redy redy?

hynek:
- why does test_mypy hate us, personally?
- it should be something like "AttrsClassInstance" or "AttrsInstance" for all I care?

hynek (approved):

PR #892, opened 2021-12-28
Do not require cloudpickle for PyPy

mgorny:
The cloudpickle package relies on CPython implementation details,
and does not even import on PyPy:

```
ImportError while importing test module '/tmp/attrs/tests/test_3rd_party.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/pypy3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_3rd_party.py:7: in <module>
    import cloudpickle
.tox/pypy3/lib/pypy3.8/site-packages/cloudpickle/__init__.py:4: in <module>
    from cloudpickle.cloudpickle import *  # noqa
.tox/pypy3/lib/pypy3.8/site-packages/cloudpickle/cloudpickle.py:57: in <module>
    from .compat import pickle
.tox/pypy3/lib/pypy3.8/site-packages/cloudpickle/compat.py:13: in <module>
    from _pickle import Pickler  # noqa: F401
E   ModuleNotFoundError: No module named '_pickle'
```

Disable the dependency for PyPy and make the test handle missing
cloudpickle gracefully.

# Summary

# Pull Request Check List

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ EXTRAS_REQUIRE = {
     "docs": ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"],
     "tests_no_zope": [
         # For regression test to ensure cloudpickle compat doesn't break.
-        "cloudpickle",
+        'cloudpickle; python_implementation == "CPython"',
         # 5.0 introduced toml; parallel was broken until 5.0.2
         "coverage[toml]>=5.0.2",
         "hypothesis",
--- a/tests/test_3rd_party.py
+++ b/tests/test_3rd_party.py
@@ -4,11 +4,14 @@
 Tests for compatibility against other Python modules.
 """
 
-import cloudpickle
+import pytest
 
 from hypothesis import given
 
 from .strategies import simple_classes
+
+
+cloudpickle = pytest.importorskip("cloudpickle")
 
 
 class TestCloudpickleCompat(object):
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "pypy-2.7", "pypy-3.7"]
+        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "pypy-2.7", "pypy-3.7", "pypy-3.8"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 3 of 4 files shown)


mgorny:
I think this is not worthy of a news fragment but I can add one if you disagree.

tests/test_3rd_party.py
```diff
@@ -4,13 +4,16 @@
...
 
-import cloudpickle
+import pytest
 
 from hypothesis import given
 
 from .strategies import simple_classes
 
 
+pytest.importorskip("cloudpickle")
```
hynek, line 14:
```suggestion
cloudpickle = pytest.importorskip("cloudpickle")
```

hynek:
I'm somewhat confused why our PyPy tests passed then? 😳

tests/test_3rd_party.py
mgorny: I'm sorry about the mistake. I should've tested it better.

mgorny:
> I'm somewhat confused why our PyPy tests passed then? flushed

Hmm, maybe it's a change in PyPy3.8's stdlib then.

setup.py
```diff
@@ -50,7 +50,7 @@
     "docs": ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"],
     "tests_no_zope": [
         # For regression test to ensure cloudpickle compat doesn't break.
-        "cloudpickle",
+        'cloudpickle; python_implementation != "PyPy"',
```
thetorpedodog, line 53: Since cloudpickle depends upon the CPython internals, maybe `python_implementation == "CPython"` instead? (I am not familiar enough with packaging to know exactly what the implementation name and syntax should be; but whatever means “install this only on CPython”, rather than just “install this unless it’s PyPy”.)

thetorpedodog (commented):
as a non-maintainer but the person who implemented the cloudpickle fix and test, I have one minor suggestion. seems like a good change overall!

setup.py
mgorny: Sure, done that and squashed the commits.

mgorny:
I have also enabled testing on pypy-3.8 and confirmed that it fails without the change: https://github.com/mgorny/attrs/runs/4656570160

hynek:
So newsfragment is not strictly necessary, but if you'd like, you can add a `Fixed the test suite on PyPy 3.8.` or something like that…the correct version is probably different?

mgorny:
Added.

hynek (approved):
Thanks, let's 🚢  this!

PR #898, opened 2021-12-30
Speedup uses of `object.__setattr__`

davfsa:
# Summary

I was playing around seeing if I could speedup the creation of frozen slot classes for a project of mine since we instantiate a lot of attr classes. Ideally they would be frozen, but we didnt go with it because of the speed implications it had.

To try and speed it up I tried different things:
- Make use of `objects.__setattr__` during the whole of the init and then place the frozen one at the end
  - The idea behind this is to use the default `__setattr__` during the whole `__init__` function and then setting it to the frozen one after, which I thought would bring speedups as well as improve the UX, since they can could set attributes normally in the init. This unfortunately didnt work nicely with slots and slowed it down even further when attempting a workaround
- Playing around a bit with python internals and trying to figure out the ideal way to implement this
  - This is the change this PR adds does

Overall I wasn't really successful, but I did find this small performance gain of "1.1x", which isnt much, but I think it might be worth having.

Bellow you can see the pypref benchmarks on my laptop (to simulate the ones that were in the documentation)

### Before (attrs == 21.4.0)

```bash
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True, frozen=True)" "C(1, 2, 3)"
Mean +- std dev: 499 ns +- 26 ns
```

### After
```bash
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True, frozen=True)" "C(1, 2, 3)"
Mean +- std dev: 450 ns +- 17 ns
```

# Pull Request Check List

### DIFF
--- a/changelog.d/898.change.rst
+++ b/changelog.d/898.change.rst
@@ -1,0 +1,1 @@
+Speedup instantiation of frozen slotted classes.
--- a/docs/how-does-it-work.rst
+++ b/docs/how-does-it-work.rst
@@ -87,16 +87,16 @@ This is (still) slower than a plain assignment:
   $ pyperf timeit --rigorous \
         -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True)" \
         "C(1, 2, 3)"
-  ........................................
-  Median +- std dev: 378 ns +- 12 ns
+  .........................................
+  Mean +- std dev: 228 ns +- 18 ns
 
   $ pyperf timeit --rigorous \
         -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True, frozen=True)" \
         "C(1, 2, 3)"
-  ........................................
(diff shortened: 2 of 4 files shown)


hynek:
Hmmm 10% more performance is nice and he solution is simpler too, which is nice!

Calling @Tinche who seems to be the daddy of the old code – any edge cases we're missing here?

Tinche:
This essentially saves us an attribute lookup at the cost of an additional argument in a couple function calls?

If the tests pass, LGTM.

hynek:
@davfsa can you rebase please? you didn't give us write permissions to your branch

hynek (approved):
Thanks!

davfsa:
Rebased and turned on edits in case anything else is needed :)

PR #909, opened 2022-01-14
Fix hashing for custom `eq` objects

OneRaynyDay:
# Summary

Thanks for making attrs btw! It's an amazing library. In the documentation for `attr.ib` (sorry, still speaking in 2021 attr here), we see:

> hash (Optional[bool]) – Include this attribute in the generated __hash__ method. If None (default), **mirror eq’s value**. This is the correct behavior according the Python spec. Setting this value to anything else than None is discouraged.

Currently this does not happen - if we passed in a `callable` to transform the field's value, we would see `_x_key(self.x)` in the generated eq code, but we don't see the same thing happening for the generated hash function.

Currently, I'd like to compare equality of a non-hashable object. I was surprised to see that the hash function did not respect the `eq` override and get it for free (which is an awesome feature from attr!).

# Pull Request Check List

As the change is only meant to make the code be compliant with the docs, I have not updated any existing documentation. If this is worth adding into the changelog let me know and I'll add it asap. Thank you!

### DIFF
--- a/changelog.d/898.change.rst
+++ b/changelog.d/898.change.rst
@@ -1,2 +1,1 @@
 Speedup instantiation of frozen slotted classes.
-If an eq key is defined, it is also used before hashing the attribute.
--- a/changelog.d/898.change.rst
+++ b/changelog.d/898.change.rst
@@ -1,1 +1,2 @@
 Speedup instantiation of frozen slotted classes.
+If an eq key is defined, it is also used before hashing the attribute.
--- a/changelog.d/909.change.rst
+++ b/changelog.d/909.change.rst
@@ -1,0 +1,1 @@
+If an eq key is defined, it is also used before hashing the attribute.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -325,13 +325,11 @@ def _compile_and_eval(script, globs, locs=None, filename=""):
     eval(bytecode, globs, locs)
 
 
-def _make_method(name, script, filename, globs=None):
+def _make_method(name, script, filename, globs):
     """
     Create the method with the script given and return the method object.
     """
     locs = {}
-    if globs is None:
-        globs = {}
 
     # In order of debuggers like PDB being able to step through the code,
     # we add a fake linecache entry.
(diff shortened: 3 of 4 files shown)


hynek (changes requested):
This looks great, thanks! Always nice to see users caring and fixing subtle edge cases. :)

This is definitely worth a changelog entry, but I wouldn't tie it to the doc entry.

Just something along the lines of "If an eq key is defined, it is also used before hashing the attribute." just in more understandable. 😇

Thanks!

P.S. also looks like coverage is failing now 😇

OneRaynyDay:
Hi @hynek , thanks for getting back to me! I've not used this coverage tool, and I was wondering what I needed to do to bring the coverage back up to 100%. I updated the changelog.d with the update :)

changelog.d/898.change.rst
```diff
@@ -1 +1,2 @@
 Speedup instantiation of frozen slotted classes.
+If an eq key is defined, it is also used before hashing the attribute.
```
hynek, line 2: this is not how our changelog works 😅

please create a new file `changelog.d/909.change.rst` with the contents.

see:  https://github.com/python-attrs/attrs/blob/main/.github/CONTRIBUTING.md#changelog

hynek:
Since I don't have push permissions against your branch, please apply the following patch and try again:

```patch
diff --git src/attr/_make.py src/attr/_make.py
index 962e560..86ced77 100644
--- src/attr/_make.py
+++ src/attr/_make.py
@@ -325,13 +325,11 @@ def _compile_and_eval(script, globs, locs=None, filename=""):
     eval(bytecode, globs, locs)
 
 
-def _make_method(name, script, filename, globs=None):
+def _make_method(name, script, filename, globs):
     """
     Create the method with the script given and return the method object.
     """
     locs = {}
-    if globs is None:
-        globs = {}
 
     # In order of debuggers like PDB being able to step through the code,
     # we add a fake linecache entry.
@@ -1955,7 +1953,7 @@ if HAS_F_STRINGS:
         ]
 
         return _make_method(
-            "__repr__", "\n".join(lines), unique_filename, globs=globs
+            "__repr__", "\n".join(lines), unique_filename, globs
         )
 
 else:
```

hynek (approved):
thanks!

Issue #907, opened 2022-01-14
Compatibility with Python 3.11


frenzymadness:
Hello.

In Fedora, we are again trying to prepare the new Python 3.11 soon and during the rebuild of our packages, we have discovered that some tests of attrs package fail:

We are building attrs 21.4.0 together with Python 3.11.0~a3 ([changelog](https://docs.python.org/3.11/whatsnew/3.11.html))

```
=================================== FAILURES ===================================
_______________________ test_init_subclass_vanilla[True] _______________________

slots = True

    @pytest.mark.parametrize("slots", [True, False])
    def test_init_subclass_vanilla(slots):
        """
        `super().__init_subclass__` can be used if the subclass is not an attrs
        class both with dict and slotted classes.
        """
    
        @attr.s(slots=slots)
        class Base:
            def __init_subclass__(cls, param, **kw):
                super().__init_subclass__(**kw)
                cls.param = param
    
>       class Vanilla(Base, param="foo"):

tests/test_init_subclass.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'tests.test_init_subclass.test_init_subclass_vanilla.<locals>.Vanilla'>
param = 'foo', kw = {}

    def __init_subclass__(cls, param, **kw):
>       super().__init_subclass__(**kw)
E       TypeError: super(type, obj): obj must be an instance or subtype of type

tests/test_init_subclass.py:24: TypeError
_____________ TestClosureCellRewriting.test_closure_cell_rewriting _____________

self = <tests.test_slots.TestClosureCellRewriting object at 0x7fb02ec6da50>

    def test_closure_cell_rewriting(self):
        """
        Slotted classes support proper closure cell rewriting.
    
        This affects features like `__class__` and the no-arg super().
        """
        non_slot_instance = C1(x=1, y="test")
        slot_instance = C1Slots(x=1, y="test")
    
        assert non_slot_instance.my_class() is C1
>       assert slot_instance.my_class() is C1Slots
E       AssertionError: assert <class 'tests.test_slots.C1Slots'> is C1Slots
E        +  where <class 'tests.test_slots.C1Slots'> = <bound method C1Slots.my_class of C1Slots(x=1, y='test')>()
[...]

PR #910, opened 2022-01-16
Fix set_closure_cell on 3.11

Tinche:
The fix for 3.11 compatibility (https://github.com/python-attrs/attrs/issues/907).

Maybe hold off on merging until a discussion on BPO.

### DIFF
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -211,15 +211,22 @@ def make_set_closure_cell():
             )
             set_first_freevar_code = types.CodeType(*args)
 
-        def set_closure_cell(cell, value):
-            # Create a function using the set_first_freevar_code,
-            # whose first closure cell is `cell`. Calling it will
-            # change the value of that cell.
-            setter = types.FunctionType(
-                set_first_freevar_code, {}, "setter", (), (cell,)
-            )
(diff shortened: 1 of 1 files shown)


hynek:
I guess we'll have to finish this up before 3.11? :-/

Tinche:
I don't think the other failing tests were related to this, but other 3.11 incompatibilities. Since the beta is out I can revisit this

src/attr/_compat.py
```diff
@@ -211,15 +211,22 @@ def force_x_to_be_a_cell():  # pragma: no cover
...
-        def set_closure_cell(cell, value):
-            # Create a function using the set_first_freevar_code,
-            # whose first closure cell is `cell`. Calling it will
-            # change the value of that cell.
-            setter = types.FunctionType(
-                set_first_freevar_code, {}, "setter", (), (cell,)
-            )
-            # And call it to set the cell.
-            setter(value)
+        if sys.version_info >= (3, 11):
```
vstinner, line 214: In fact, it works since Python 3.8:

* https://github.com/python/cpython/commit/df8d2cde63c865446468351f8f648e1c7bd45109
* https://github.com/python/cpython/issues/91299#issuecomment-1093947264

PR #915, opened 2022-01-28
Remove incorrect statement in documentation

ngnpope:
# Summary

Both `attr.define` and `attr.s` have `match_args=True` by default.


hynek (approved):
Oops, thanks!

Issue #924, opened 2022-03-01
Passing multiple Validators to the `member_validator` param of the `deep_iterable` validator


vedantpuri:
This could be a dupe, but I noticed when using the `deep_iterable` validator the `member_validator` param expects a callable, whereas in other cases it is possible to pass in a list of validators that would be applied in order (equivalent to `and_`)

Was this behavior intentional for some reason ?

If not, would it be possible to update the `deep_iterable` to take in a list of validators ?

Hope to hear back soon! Thanks

PR #925, opened 2022-03-04
Allow `deep_iterable` member validator to accept a list of validators

vedantpuri:
# Summary

Proposed fix for #924 
Allows users to pass in a list of validators within the member validator of the `deep_iterable` validator. I feel this makes it more consistent with what users would expect. I atleast had a minor use case when using `attrs` and thought it would be worthwhile to support this.

This PR contains:
- Small tweak to the `deep_iterable` validator
- Updates to the doc string of `deep_iterable`
- Tests for the new behaviour (backward compatible)
- Minor doc string fix in a different test

If people think this is useful I would be happy to make a different PR to support the same for the `deep_mapping` validator

# Pull Request Check List

### DIFF
--- a/changelog.d/925.change.rst
+++ b/changelog.d/925.change.rst
@@ -1,0 +1,1 @@
+Allow the `member_validator` of the `deep_iterable` validator to accept a list of validators
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -415,8 +415,8 @@ def deep_iterable(member_validator, iterable_validator=None):
 
     :raises TypeError: if any sub-validators fail
     """
-    if isinstance(member_validator, list):
-        member_validator = _AndValidator(member_validator)
+    if isinstance(member_validator, (list, tuple)):
+        member_validator = and_(*member_validator)
     return _DeepIterable(member_validator, iterable_validator)
 
 
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -408,7 +408,7 @@ def deep_iterable(member_validator, iterable_validator=None):
     A validator that performs deep validation of an iterable.
 
     :param member_validator: Validator(s) to apply to iterable members
-    :param iterable_validator: Validator(s) to apply to iterable itself
+    :param iterable_validator: Validator to apply to iterable itself
         (optional)
 
     .. versionadded:: 19.1.0
@@ -417,8 +417,6 @@ def deep_iterable(member_validator, iterable_validator=None):
     """
     if isinstance(member_validator, list):
         member_validator = _AndValidator(member_validator)
-    if iterable_validator and isinstance(iterable_validator, list):
-        iterable_validator = _AndValidator(iterable_validator)
     return _DeepIterable(member_validator, iterable_validator)
 
 
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -407,8 +407,8 @@ def deep_iterable(member_validator, iterable_validator=None):
     """
     A validator that performs deep validation of an iterable.
 
-    :param member_validator: Validator to apply to iterable members
-    :param iterable_validator: Validator to apply to iterable itself
+    :param member_validator: Validator(s) to apply to iterable members
+    :param iterable_validator: Validator(s) to apply to iterable itself
         (optional)
 
     .. versionadded:: 19.1.0
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -415,6 +415,10 @@ def deep_iterable(member_validator, iterable_validator=None):
 
     :raises TypeError: if any sub-validators fail
     """
+    if isinstance(member_validator, list):
+        member_validator = _AndValidator(member_validator)
+    if iterable_validator and isinstance(iterable_validator, list):
+        iterable_validator = _AndValidator(iterable_validator)
     return _DeepIterable(member_validator, iterable_validator)
 
 
--- a/src/attr/validators.pyi
+++ b/src/attr/validators.pyi
@@ -63,7 +63,7 @@ def matches_re(
     ] = ...,
 ) -> _ValidatorType[AnyStr]: ...
 def deep_iterable(
-    member_validator: _ValidatorArgType[T],
+    member_validator: _ValidatorArgType[_T],
     iterable_validator: Optional[_ValidatorType[_I]] = ...,
 ) -> _ValidatorType[_I]: ...
 def deep_mapping(
--- a/src/attr/validators.pyi
+++ b/src/attr/validators.pyi
@@ -62,7 +62,7 @@ def matches_re(
     ] = ...,
 ) -> _ValidatorType[AnyStr]: ...
 def deep_iterable(
-    member_validator: _ValidatorType[_T],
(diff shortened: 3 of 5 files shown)


vedantpuri:
@hynek / @Tinche Thoughts on this?

hynek:
Sorry for the delay Vedant; I’ll get to it.

vedantpuri:
> Sorry for the delay Vedant; I’ll get to it. 

Sounds good, Thankyou!

tests/test_validators.py
```diff
@@ -583,42 +594,64 @@ def test_fail_invalid_iterable(self):
...
+            member_repr = (
+                "_AndValidator(_validators=[{func}, "
+                "<instance_of validator for type <{type} 'int'>>])"
+            ).format(func=repr(always_pass), type=TYPE)
+        else:
+            member_repr = (
+                "<instance_of validator for type <{type} 'int'>>".format(
+                    type=TYPE
+                )
+            )
```
hynek, line 654: This should be two test methods.

tests/test_validators.py
```diff
@@ -515,21 +515,25 @@ def test_in_all(self):
         """
         assert deep_iterable.__name__ in validator_module.__all__
 
-    def test_success_member_only(self):
+    @pytest.mark.parametrize(
+        "member_validator", [instance_of(int), [always_pass, instance_of(int)]]
+    )
```
hynek, line 520: Your using this a lot; please transform it into a [parametrized fixture](https://hynek.me/til/parametrized-pytest-fixtures/).

hynek (changes requested):
A few minor issues, plus also needs a [news fragment](https://github.com/python-attrs/attrs/blob/main/.github/CONTRIBUTING.md#changelog).

tests/test_validators.py
vedantpuri: This is exactly what I was looking for, Nice blog post!

tests/test_validators.py
vedantpuri: I split them in my latest commit

src/attr/validators.pyi
```diff
@@ -62,7 +62,7 @@ def matches_re(
     ] = ...,
 ) -> _ValidatorType[AnyStr]: ...
 def deep_iterable(
-    member_validator: _ValidatorType[_T],
+    member_validator: Union[_ValidatorType[_T], List[_ValidatorType[_T]]],
```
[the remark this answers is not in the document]
vedantpuri: Done 👍

vedantpuri:
@hynek Thanks for your comments, I added in the Changelog and responded to PR comments!

tests/test_validators.py
hynek: my review took a day longer because i wanted to write it up properly ;)

tests/test_validators.py
```diff
@@ -630,6 +647,28 @@ def test_repr_member_and_iterable(self):
...
+        iterable_validator = instance_of(list)
+        iterable_repr = (
+            "<instance_of validator for type <{type} 'list'>>"
+        ).format(type=TYPE)
+        v = deep_iterable(member_validator, iterable_validator)
+        expected_repr = (
+            "<deep_iterable validator for"
+            " {iterable_repr} iterables of {member_repr}>"
+        ).format(iterable_repr=iterable_repr, member_repr=member_repr)
+        assert expected_repr == repr(v)
```
hynek, line 670:
```suggestion

        assert expected_repr == repr(v)
```

changelog.d/925.change.rst
```diff
@@ -0,0 +1 @@
+Allow the `member_validator` of the `deep_iterable` validator to accept a list of validators
```
hynek, line 1: This isn't markdown :)

```suggestion
``attrs.validators.deep_iterable()``'s *member_validator* argument now also accepts a list of validators and wraps them in an ``attrs.validators.and_()``.
```

src/attr/validators.py
```diff
@@ -407,14 +407,16 @@ def deep_iterable(member_validator, iterable_validator=None):
...
+    :param member_validator: Validator(s) to apply to iterable members
     :param iterable_validator: Validator to apply to iterable itself
         (optional)
 
     .. versionadded:: 19.1.0
 
     :raises TypeError: if any sub-validators fail
     """
+    if isinstance(member_validator, list):
+        member_validator = _AndValidator(member_validator)
```
hynek, line 419: I've just double-checked and if I see it correctly, the validator argument of fields uses `and_()` because that flattens other lists/ands.

hynek (changes requested):

tests/test_validators.py
hynek: Did you forget to push? There's still an `isinstance`?

src/attr/validators.py
vedantpuri: Didn't quite understand this, could you elaborate ?

tests/test_validators.py
vedantpuri: Ahh, there were two tests that used `isinstance`, I thought you wanted only one to be split. I just split the other one too

src/attr/validators.py
hynek: My point is that you're directly instantiating the `_AndValidator` while the equivalent logic otherwhere uses `and_` because that has extra logic:

https://github.com/python-attrs/attrs/blob/b75436630fafde95ba2954052c5c4d7d4f506b28/src/attr/_make.py#L295-L296

Which reminds me that you should also check for tuple too. :)

hynek (changes requested):
The CI is very red too. :)

src/attr/validators.py
vedantpuri: ahh I see. I was trying to mimic how this same logic was implemented for the `optional` validator. I'm happy to change it to `and_` if thats better but then we should probably add in a issue to update the `optional` validator too (I can make the issue if you'd like)
https://github.com/python-attrs/attrs/blob/main/src/attr/validators.py#L278-L293

vedantpuri:
@hynek Responded to your comments, Hopefully CI should pass this time (tried running tox locally with py39 and the issue seems to be resolved)

Might be worth changing the repo workflow to allow the CI to run on PRs ?

src/attr/validators.py
vedantpuri: Haven't updated this yet, will hold off till we have discussed^

hynek: Hm yes, optional should be fixed too, but _not_ as part of this PR. But make this one as in _make please.

vedantpuri: Sounds good, I'll open a separate Issue for that.

Updated this PR with the `and_` validator

hynek:
> Might be worth changing the repo workflow to allow the CI to run on PRs ?

It totally does; the problem is that crypto grifters started abusing free CIs for mining, so now the first PRs for a user on each project are restricted.  :-/

hynek (approved):
Thanks and sorry for the fuss!

AdrianSosic:
Hi @vedantpuri, I noticed the same behavior for `deep_mapping` today and then saw that you offered in the PR description to also fix that one. Since the issue is still open, I just wanted to ask if this has just been deprioritized or if there is maybe another reason that speaks against fixing?

PR #936, opened 2022-03-19
Drop Python 2.7

hynek:
We have paid our dues.

The tooling situation around Python 2 has deteriorated to such a degree that
upholding compatibility is not tenable anymore for a volunteer-run project.

Here's the stats for March so far:

```
| category | percent |  downloads |
|:---------|--------:|-----------:|
| 3.7      |  33.67% | 18,582,627 |
| null     |  19.82% | 10,936,104 |
| 3.8      |  18.75% | 10,344,976 |
| 3.9      |  11.20% |  6,180,529 |
| 3.6      |  10.88% |  6,006,746 |
| 3.10     |   2.95% |  1,625,923 |
| 2.7      |   1.87% |  1,030,024 |
| 3.5      |   0.84% |    463,005 |
| 3.4      |   0.02% |     10,044 |
| 3.11     |   0.01% |      7,899 |
| 3.3      |   0.00% |         25 |
| 2.6      |   0.00% |          8 |
| 3.2      |   0.00% |          2 |
| Total    |         | 55,187,912 |

Date range: 2022-03-01 - 2022-03-18
```

The main impetus being that Python 2.7 is as of macOS 12.3 not usable with virtualenv anymore. I will probably drop 3.5 next since nobody uses it and I can't install it anymore.

There's a lot of code that could probably be simplified (e.g. when just aliases are imported from _compat), but we can keep that for another PR. Same as ridding docs of `(object)`s.


hynek:
I'm gonna need some guidance re `__qualname__`.

We've got partial coverage in

https://github.com/python-attrs/attrs/blob/b093bc7554ddd1e081ad7afb3e5bd22ce62027e2/src/attr/_make.py#L864-L866

(looks like qualname's never None)

and

https://github.com/python-attrs/attrs/blob/2047cfaa51f907463a7f21f76b0e0b727d9240cb/src/attr/_make.py#L1945-L1950

Here:

Line 1950 is never executed so `qualname` appears to be never `None`, contradicting the comment??? Or is the comment so old that it just meant "new Pythons"?

---

Qualified names have been added in 3.3, but maybe there's some weird way to create classes without a qualname?

ssbarnea:
Bit late but drop it! Better late than never.

sscherfke:
🎉 🎉 🎉

hynek:
So I've decided:

1. the second instance is a non-issue, since the new f-string-based repr never cared about it'
2. the first one is a risk we'll take. i mean i've got a tradition of quickly releasing a hot fix release for edge cases to uphold

hynek:
Less emoji, more reviewing gentlepeople! ;)

ssbarnea (approved):

ssbarnea:
I rarely see people that do not have review rights (write) doing format review approvals on github. Even when I explicitly asked them to do it, it got very poor results. Maybe you will be more successful. Another issue is the size of the PR, usually I would have only modify metadata to require newer python and remove the CI jobs for py27, before gradually remove the code in follow-ups, so the reviews are more humanely reviewable.

sscherfke:
I skimmed the changes and haven't noticed any problems.  Also ran the tests on my machine and checked that it does not break Typed Settings.

src/attr/_compat.py
```diff
@@ -24,180 +25,76 @@
...
+        RuntimeWarning,
+        stacklevel=2,
+    )
 
-        return type(name, bases, ns)
 
-    # TYPE is used in exceptions, repr(int) is different on Python 2 and 3.
-    TYPE = "type"
+def isclass(klass):
+    return isinstance(klass, type)
```
wbolster, line 43: this doesn't really warrant a compat helper anymore, but that can be removed later

src/attr/_compat.py
```diff
@@ -24,180 +25,76 @@
...
-            raise TypeError(
-                "'mappingproxy' object does not support item assignment"
-            )
 
-        def update(self, _):
-            # We gently pretend we're a Python 3 mappingproxy.
-            raise AttributeError(
-                "'mappingproxy' object has no attribute 'update'"
-            )
+def iteritems(d):
```
wbolster, line 49: same, i guess a cleanup pass to get rid of compat helpers is useful. can ve done later (and i can help if appreciated)

setup.py
```diff
@@ -56,7 +54,6 @@
         "hypothesis",
         "pympler",
         "pytest>=4.3.0",  # 4.3.0 dropped last use of `convert`
```
wbolster, line 56: i guess this also heavily outdated?

src/attr/_compat.py
```diff
@@ -1,15 +1,16 @@
...
 import threading
 import types
 import warnings
 
+from collections.abc import Mapping, Sequence  # noqa
+
 
-PY2 = sys.version_info[0] == 2
 PYPY = platform.python_implementation() == "PyPy"
 PY36 = sys.version_info[:2] >= (3, 6)
```
wbolster, line 15: 3.5 is also end of life, and tooling is also deteriorating.

‘attrs supports supported python versions’ is not that weird

src/attr/_make.py
```diff
@@ -1634,34 +1614,19 @@ def wrap(cls):
...
-        Check whether *cls* has a frozen ancestor by looking at its
-        __setattr__.
-        """
-        return cls.__setattr__ == _frozen_setattrs
+def _has_frozen_base_class(cls):
+    """
+    Check whether *cls* has a frozen ancestor by looking at its
+    __setattr__.
+    """
+    return cls.__setattr__ == _frozen_setattrs
```
wbolster, line 1611: nit: this should be an identity check?

wbolster (approved):
left a few comments, mostly nits

didn't spot anything (too) weird 🙃

changelog.d/936.breaking.rst
```diff
@@ -0,0 +1,6 @@
+Python 2.7 is not supported anymore.
+
+The tooling around Python 2.7 has become too tedious for a volunteer-run project.
```
Tinche, line 3:
```suggestion
Dealing with Python 2.7 tooling has become too difficult for a volunteer-run project.
```

Tinche (approved):
Cattrs tests pass ;)

setup.py
hynek: well…it's still true ¯\_(ツ)_/¯

src/attr/_compat.py
hynek: i literally wrote about that in the PR ;)

i want to keep it as short as possible, there'll be more subsequent cleanups

src/attr/_compat.py
hynek: yup!

setup.py
hynek: (I don't think we use any newer features anywhere?)

src/attr/_compat.py
hynek: yeah as I wrote 3.5 is going out too. but baby steps :)

src/attr/_make.py
hynek: Seems to work too, yes. I remember that comparing functions and methods was very inconsistent over the versions. 🤔

PR #939, opened 2022-03-21
Drop the deprecated Attribute.cmp

hynek:
This has nothing to do with the cmp argument to attr.s/define which have been
undeprecated.


hynek:
Um I have only now realized that the `cmp` shortcut didn't make it over to the NG APIs.

Nobody complained so do we care?

wbolster (approved):

wbolster:
imho, no need for shortcuts in newer APIs for things that are of mostly historical interest. explicit > implicit etc, and everything `cmp`-based sorting related is a 2nd class citizen in modern python anyway

PR #940, opened 2022-03-21
No longer explicitly subclass ‘object’

wbolster:
This is no longer relevant in Python 3+. Follow-up to !936.


hynek (approved):

hynek:
Thanks Mr GitLab. ;)

wbolster:
ok eh yeah #936

(don't get me started about gitlab ux though 🤬)

PR #941, opened 2022-03-21
Remove various obsolete _compat helpers

wbolster:
See individual commits.


wbolster:
(missed one format string on the first push; should be good now)

hynek (approved):
Thaaaanks!

PR #942, opened 2022-03-21
Clean up some more Python 2-isms

wbolster:
- Reword and simplify a few docstrings.
- Add back a test for 2-arg super() in property() definitions that
  was (accidentally?) dropped in #936.


hynek (approved):

hynek:
Thanks, that super thing was pyupgrade. Fortunately the other super call is also in test_slots that I've blocked in pre-commit.

PR #944, opened 2022-04-01
Remove mutable default arg in function example

MicaelJarniac:
Removed mutable default arg, replaced with `None`.
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

Replaced `dict()` with `{}`.
https://stackoverflow.com/questions/8424942/creating-a-new-dictionary-in-python

# Summary

Documentation changes.

### DIFF
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -171,7 +171,7 @@ Here are some tips for effective use of metadata:
 
   .. doctest::
 
-    >>> from attr import fields, NOTHING
+    >>> from attr import field, NOTHING
     >>> MY_TYPE_METADATA = '__my_type_metadata'
     >>>
     >>> def typed(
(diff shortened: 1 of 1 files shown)


MicaelJarniac:
https://docs.sourcery.ai/suggestions/default-mutable-arg/

https://docs.sourcery.ai/refactorings/dict-literal/

hynek:
I have personally grown more tolerant of using `{}` as a default value, as long as it doesn't get mutated internally simply for the reason that it's nicer to type: one is always dict, the other is `dict | None`.

That said, the current line is cumbersome (I don't think I've ever used `dict()` myself) so the assignment [sc]ould be fixed either way.

hynek (approved):

PR #949, opened 2022-04-06
Add missing cmp_using to stub files

### LINKED ISSUE #948: Missing type annotations for attrs.cmp_using
`attrs.cmp_using` is not included in `attrs/__init__.pyi`, and as a result, attempting to use Mypy to type check any code that uses `attrs.cmp_using` will result in Mypy complaining that `cmp_using` does not exist:

    problem.py:8: error: Module has no attribute "cmp_using"

To reproduce, install Mypy, NumPy, pytest, and attrs, download the attached `problem.py`, and run `mypy problem.py`.  In contrast, when you run `pytest problem.py`, the code runs without issue – the problem is solely with the type annotations (or lack thereof).

`problem.py`:

```python
import attrs
import numpy as np
import numpy.typing as npt


@attrs.define
class Example:
    array: npt.NDArray[np.int64] = attrs.field(eq=attrs.cmp_using(np.array_equal))


def test_attrs() -> None:
    a = Example(array=np.array([1, 2, 3]))
    assert a == a
```

hynek:
Fixes  #948

cc @botant @Julian

### DIFF
--- a/changelog.d/949.change.rst
+++ b/changelog.d/949.change.rst
@@ -1,0 +1,1 @@
+Added missing stub for ``attr(s).cmp_using()``.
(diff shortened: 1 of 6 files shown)


Julian:
LGTM I think!

Issue #945, opened 2022-04-02
Add PEP681 explicit & overridable per-attribute __init__ `alias`.


asford:
[PEP681](https://peps.python.org/pep-0681/) introduces an optional `alias` parameter to field descriptors, which specifies an alternative `__init__` name for the member. This is not implemented in dataclasses, but is [implemented in pydantic](https://pydantic-docs.helpmanual.io/usage/schema/#field-customization). 

It would be useful to implement an explicit & overridable per-attribute `alias` in attrs, mirroring the behavior described in the PEP and `dataclass_transform` specification. This would allow users to directly address the single largest incompatibility noted in the specification,  `attrs` [private attributes handling](https://www.attrs.org/en/stable/init.html?highlight=private#private-attributes), which strips leading underscores from member names, discussed in #795.

Additionally, though we all love `attrs` as a tastemaker with strongly held principles, "private" attribute handling appears to have been (be?) a significant source of confusion and is likely causing at least one active attrs-internal bug. This has also been a source of issues in downstream libraries, noted in references below. 

In the longer-term, this would enable additional *optional* configuration flags to bring `attrs` into a strict-superset of the behavior of `dataclasses`, as roughly noted in #565 and #686. Enabling this work is a design goal here, but outside the scope of this issue.

A (simple?) solution may be to extend `attr.ib` with an optional `alias` parameter, which sets the `__init__` name for the field.
This would, as is the custom, be added to `attrs.Attribute` and would be available during for introspection in `attr.fields` and customization in the `field_transformer` hook.
If `alias` is unspecified it will be set to a `str`, indicating the attribute's argument name in the generated `__init__`.

For example in the current behavior, `alias = name.lstrip("_")` so that:
```python
>>> @attr.define
... class Foo:
...    bar: str
...    _bat: str
>>> attr.fields_dict(Foo)
... { "bar": Attribute(name="bar", alias="bar", ...), "_bat": Attribute(name="_bat", alias="bat", ...) }
```

We would retain the default private attribute renaming behavior in this implementation, which would *allow* users to either adjust the `alias` of the attribute to match their preferences or inform `dataclass_transform` consumers of the attribute's `__init__` name. Ie:

```python
@attr.define
class InformDataclassTransform:
    bar: str
    _bat: str = attr.ib(alias="bat")

@attr.define
class AliasMyWay:
   bar: str
   _bat: str = attr.ib(alias="_bat")
```

### Refs:

Related issues: 
[...]

Issue #572, opened 2019-09-10
Store attribute name used in __init__ parameters


altendky:
It has come up a couple times recently that I wanted to automatically build an instance of an attrs class.  The latest is to just fill it with `True` and then use as a filter for other instances of the class.  For this it would be nice to be able to get the name of a parameter for an attribute in the `__init__()` signature.  I understand that at present the rules for attribute-name -> init-name are pretty simple but it seems reasonable to not force code outside attrs to be aware of it.  I would think either `attr.init_name(cls, name)` or `Attribute.init_name` could resolve this issue.  In my couple cases I would have used the latter though the former is more general and doesn't require a class to be present.  But... the former also presumes a totally constant rule which imposes some amount of limitation on flexibility of the translation.

Anyways, I think `Attribute.init_name` should be easy to add and should cover a vast majority of uses.

I'm happy to make up a PR if this is or might be of interest.

https://github.com/python-attrs/attrs/blob/8824dc26c219abcb43564dd9386fe1a88f938344/src/attr/_make.py#L1681-L1693

hynek:
Since we'll be releasing the field transformation code in the next release, I think this should go in before it gets more complicated. Are you still interested in it?

hynek:
@sscherfke this seems like something that should be of your interest too?

sscherfke:
Hm. I never thought of this before, because usually `attr name == init name` is True. 

I also can’t think of a concrete use case for which you’d need a mapping between field name and init arg-name. 🤔 @altendky could you please provide a concrete example for this?

altendky:
I don't remember the details to make it more concrete than in the OP, sorry.  That said, being able to construct a class instance without explicitly writing out the parameters seems to not be a big request here.  `:]`  Of course, the flip side is that the most interested person here (me) doesn't have a concrete example at present.

If a PR is wanted, I can try to fit that in.  I'm figuring that just `Attribute.init_name` would be the bit to add.  `attr.ib(init=False)` would give `None`?  Or would we need some more specific sentinel to cover some corner case.

altendky:
[...]

Issue #795, opened 2021-04-23
pyright support via dataclass transforms?


asford:
Perhaps this is "already on the radar" around these parts, however it looks like pyrite may begin to support attrs via a generic `typing`-based declaration:

https://github.com/microsoft/pyright/pull/1773
https://github.com/microsoft/pyright/blob/master/specs/dataclass_transforms.md

Would y'all be open to a PR implemented the recommendations in the linked specification?
Perhaps there's a dialog to be had to ensure that the upcoming next-gen decorator semantics can be supported via this extension mechanism? (I don't have enough of the picture in mind to understand whether this is true.)

hynek:
Yes, Eric was kind enough to reach out to me a few weeks ago and we discussed the topic.

Their integration doesn't cover all of our features but it seems good enough – at least good enough to not deliver false positives. NG APIs should work too.

I was hoping for feedback from @euresti (or any other typing gurus around here, really) before going to town on it. But maybe we can indeed squeeze it into 21.1 – at least provisionally.

hynek:
Also pulling in @Tinche who seems to have already played with it!

Tinche:
Seems to work for me. The cost/benefit analysis seems in favor of adding this: it's very easy for us to do and IDE support is very useful, particularly if you're not quite ready to start using mypy.

asford:
I've also verified this on an internal codebase and found it to be generally functional.

Pleasantly, this allows us to implement `attrs` [excellent advice on wrapping the decorator](https://www.attrs.org/en/stable/extending.html#wrapping-the-decorator) in a reasonably straightforward fashion, however as you mention is is a subset of both attrs' current semantics and the mypy-attrs plugin's current semantics.

I do feel that there's some value in having `attrs` begin to support this _provisionally_. It's a huge potential boost to VS Code users and could be a good way to drive more feedback on the `pyright` spec on additional "non dataclass" semantics like `converter`.

euresti:
This is interesting. Similar to an idea I had for supporting this in https://github.com/python/mypy/issues/5406#issuecomment-409233479

I wonder if this could be used in mypy itself.  Might be complicated with all the caching mypy does, but ... maybe?

euresti:
How does it work with the overloads?  Do you only annotate the real implementation?  Can these be put in a .pyi file?

hynek:
@euresti did you get Eric’s e-mail? He sent it to your GitHub address. That would be the best way to get direct answers. :)

[...]

Issue #686, opened 2020-09-03
PEP 557 interop namespace


micimize:
Extracted from https://github.com/python-attrs/attrs/issues/408#issuecomment-514375369

> On the other side of this, what about making a namespace that actually _is_ intentionally similar to PEP 557 `dataclasses` and `fields`, and just has a thin adapter layer between `field` and `attr.ib`, etc? How much incompatibility is there beyond the names? Isn't `attrs` mostly a superset?

It would be good for migrating from large codebases where `dataclasses` are already used, or for pythonistas who want codebases/apis that stick close to the PEP specifications where possible.

wsanchez:
I think that if we did that, we should have an `attrs.dataclass` namespace, rather than using the primary namespace.

The reason for that is that if we start down that path and set and expectations that Attrs is PEP 557 (plus whatever extra awesome) then (a) we might get cornered into design decisions made elsewhere in order to maintain this expectation (b) we might have used a name that conflicts with some future dataclass addition and (c) compatibility may require contortions that would be better contained in a compatibility namespace.

micimize:
agreed – I think it should actually be `attrs.dataclasses`, such that migrating from the standard lib is just adding the prefix:
`sed 's/from dataclasses import/from attrs.dataclasses import'`

asford:
+1 to this idea, given what we've seen in https://github.com/python-attrs/attrs/issues/795, I think there are now more knock-on benefits than ever to adding a "dataclass compatible namespace". 

A few potential benefits would include:

* Direct support for `pyright` et al via the `dataclass_transform` spec, which is now [PEP 681](https://peps.python.org/pep-0681/)
* A smooth upgrade path for codebases that ~~see the light~~ wish to adopt attrs features.
* A smooth "reverse compatibility" path for codebases using attrs, but need to interact with dataclass-ish libraries.

The goal of the namespace would be to provide an attrs mode that is _superset_ of  `stdlib` dataclasses,
where an application can update to attrs by just changing an `import` statement.
This falls into two branches: differences in behavior for existing features and features in dataclasses not supported in attrs.
A non-exhaustive index includes:

Differences for existing features are blocking changes that prevent an application from directly converting from `dataclasses` to `attrs`. This will require optional changes to `attrs` behavior and are "must have" for this proposed namespace:

* `attrs` [private attributes handling](https://www.attrs.org/en/stable/init.html?highlight=private#private-attributes), which strips leading underscores from member names.
[...]

PR #950, opened 2022-04-09
Add Attribute.alias

asford:
# Summary

[PEP681](https://peps.python.org/pep-0681/) introduces an optional `alias` parameter to field descriptors, which specifies an alternative `__init__` name for the member. This is not implemented in dataclasses, but is [implemented in pydantic](https://pydantic-docs.helpmanual.io/usage/schema/#field-customization). 

Add an explicit & overridable per-attribute `alias` in attrs, mirroring the behavior described in the PEP and `dataclass_transform` specification. This allows users to directly address the single largest incompatibility noted in the specification,  `attrs` [private attributes handling](https://www.attrs.org/en/stable/init.html?highlight=private#private-attributes), which strips leading underscores from member names, discussed in #795.

This feature unblocks implementation of a shim layer to support `attrs`/`dataclass` interop, discussed in:
* https://github.com/python-attrs/attrs/issues/565
* https://github.com/python-attrs/attrs/issues/686

closes: https://github.com/python-attrs/attrs/issues/945
closes: https://github.com/python-attrs/attrs/issues/572

Goals:
- [x] This is silent no-op for "normal" users of `attrs` not using `alias`. Libraries building on `attrs` *may* use `alias` to detect both existing private-name init aliases and user-supplied aliases. 
- [x] Add an `alias : Optional[str]` parameter to `attr.ib` et al and property to `Attribute`.
- [x] Lift current private-attribute-naming logic (`name.lstrip("_")`) into `_ClassBuilder` so that `alias` is always populated with the field's `__init__` parameter name, regardless of whether an explicit alias is provided, during `__init__` rendering and `attr.fields` introspection.
- [x] Leave `Attribute.alias` as `None` if not explicitly specified, so that `field_transformer` implementations can (a) create attributes without needing to be aware of `alias` and optionally provide `alias` overrides before `attrs`.
- [x] Update `attr.evolve` to use `Attribute.alias` rather than re-implemented name-mangling.
- [x] Update the attribute and private-attribute-handling documentation to note the new `alias` parameter. 
- [x] Expand dataclass_transform documentation to describe how `alias` can be used to inform type checkers of private-name renaming behavior.
- [x] Politely file a bug against `attrs` noting the inconsistency between documentation and behavior. Dunder-methods are still affected by https://github.com/python-attrs/attrs/issues/619, but `evolve` failures in pathological cases are resolved.
- [x] File a tracking issue against against `cattrs` to access `Attribute.alias` during codegen iff available in `attrs`: https://github.com/python-attrs/cattrs/issues/322

Non-goals:
- Modification of the existing private-attribute-naming logic, though could be tackled pre-release in a follow-up PR.

# Pull Request Check List


docs/extending.rst
```diff
@@ -260,6 +260,27 @@ A more realistic example would be to automatically convert data that you, e.g.,
...
+   ...         for field in fields
+   ...     ]
+   ...
+   >>> @frozen(field_transformer=dataclass_names)
+   ... class Data:
+   ...     public: int
+   ...     _private: float
+   ...
+   >>> Data(public=42, _private="spam")
+   Data(public=42, _privatge='spam')
```
asford, line 282:
```suggestion
   Data(public=42, _private='spam')
```

src/attr/_funcs.py
```diff
@@ -361,7 +361,7 @@ def evolve(inst, **changes):
         if not a.init:
             continue
         attr_name = a.name  # To deal with private attributes.
-        init_name = attr_name if attr_name[0] != "_" else attr_name[1:]
+        init_name = a.alias
```
hynek, line 364: this is actually a nice effect

src/attr/_make.py
```diff
@@ -2247,7 +2261,10 @@ def fmt_setter_with_converter(
         has_on_setattr = a.on_setattr is not None or (
             a.on_setattr is not setters.NO_OP and has_cls_on_setattr
         )
-        arg_name = a.name.lstrip("_")
+        assert (
+            a.alias
+        ), "requires known field alias, should have been set via _ClassBuilder"
```
hynek, line 2266: did you leave this in on purpose?

docs/init.rst
```diff
@@ -42,6 +42,7 @@ Embrace functions and classmethods as a filter between reality and what's best f
 If you look for object serialization, there's a bunch of projects listed on our ``attrs`` extensions `Wiki page`_.
 Some of them even support nested schemas.
 
+.. _private_attributes:
```
hynek, line 45: If I see it correctly, there's no narrative documentation for the new feature?


Would you mind adding it here and into examples.rst somewhere around the handling of private attributes?

hynek (changes requested):
This looks all very good and thorough, just a bunch of questions and nits.

it also needs a newsfragment. make it sounds excited. :)

sorry for taking so long…whenever Attribute changes the diffs make me wann hide under the bed.

hynek:
Anything you need from my end? 😇

asford (commented):
My turn to apologize for the slow turnaround.

I've applied the review comments, and believe this is ready for a final pass.

docs/extending.rst
```diff
@@ -260,6 +260,27 @@ A more realistic example would be to automatically convert data that you, e.g.,
    >>> Data(**from_json)  # ****
    Data(a=3, b='spam', c=datetime.datetime(2020, 5, 4, 13, 37))
 
+Or, perhaps you would prefer alternative default field names.
+Noting, that ``field_transformer`` receives `attrs.Attribute` instances before private-attribute handling.
```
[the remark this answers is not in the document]
asford: Tidied and reworded.

hynek:
w00t! Thanks for coming back! I'll try to be faster this time! :)

asford:
Really appreciate your work on attrs.
Thanks for being stoked on the contribution!

src/attr/_make.py
```diff
@@ -242,6 +246,7 @@ def attrib(
     .. versionchanged:: 21.1.0
        *eq*, *order*, and *cmp* also accept a custom callable
     .. versionchanged:: 21.1.0 *cmp* undeprecated
+    .. versionchanged:: 21.5.0 *alias*
```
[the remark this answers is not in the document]
hynek: you have not :D

hynek (approved):
Nailed it a tiny nit aside!

Thank you so much for coming back, I know how difficult it is to pick up something like this after a long time.

Tinche:
Cool. Cattrs will be no problem, I can just bump the dependency to the latest version of attrs and implement it.

Mypy will be a bigger issue I think.

hynek:
given alias is a dataclass transform feature, shouldn't it work OOB?

Tinche:
Unsure, Mypy has a custom attrs plugin so I don't know if the dataclass_transform code path is in charge of this.

hynek:
aw looks like you're right. pyright works: #1063

mypy doesn't:

Adding 

```
AliasExample(without_alias=42, _with_alias=23)
```

gives me: `tests/typing_example.py:134: error: Unexpected keyword argument "_with_alias" for "AliasExample"; did you mean "with_alias" or "without_alias"?`

@Tinche 🐶🥺?

hynek:
yeah, it's super overdue too.

i'll try to get #1003 in ASAP (it's just an alias, so it should be just busywork) and push it out RSN.

PR #969, opened 2022-06-09
Add support for Python 3.11

hynek:
Extracted and fixes #910

PR #973, opened 2022-06-22
Tin/py311

Tinche:


tests/test_slots.py
```diff
@@ -660,10 +660,12 @@ def test_no_getstate_setstate_for_dict_classes(self):
         As long as getstate_setstate is None, nothing is done to dict
         classes.
         """
-        i = C1(1, 2)
-
-        assert None is getattr(i, "__getstate__", None)
-        assert None is getattr(i, "__setstate__", None)
+        assert getattr(object, "__getstate__", None) == getattr(
```
hynek, line 663: Shouldn't `is` work here and be more correct too? I seem to remember that it didn't work on Python 2, but that shouldn't be an issue anymore.

tests/test_make.py
```diff
@@ -2275,7 +2275,9 @@ class C:
             def __getstate__(self):
                 return ("hi",)
 
-        assert None is getattr(C(), "__setstate__", None)
+        assert getattr(object, "__setstate__", None) is getattr(
+            C, "__setstate__", None
+        )
```
hynek, line 2280: hm this makes no sense to me…this would pass if both don't exist, and if both are the same thing. this doesn't seem like a good test to me :)

maybe have two asserts, depending on version?

tests/test_make.py
```diff
@@ -2291,7 +2293,9 @@ def __setstate__(self, state):
         i.__setstate__(())
 
         assert True is i.called
-        assert None is getattr(C(), "__getstate__", None)
+        assert getattr(object, "__getstate__", None) is getattr(
```
hynek, line 2296: same as above

tests/test_slots.py
```diff
@@ -660,10 +660,12 @@ def test_no_getstate_setstate_for_dict_classes(self):
         As long as getstate_setstate is None, nothing is done to dict
         classes.
         """
-        i = C1(1, 2)
-
-        assert None is getattr(i, "__getstate__", None)
-        assert None is getattr(i, "__setstate__", None)
+        assert getattr(object, "__getstate__", None) is getattr(
```
hynek, line 663: same as above

tests/test_slots.py
```diff
@@ -674,10 +676,12 @@ def test_no_getstate_setstate_if_option_false(self):
         class C:
             x = attr.ib()
 
-        i = C(42)
-
-        assert None is getattr(i, "__getstate__", None)
-        assert None is getattr(i, "__setstate__", None)
+        assert getattr(object, "__getstate__", None) is getattr(
+            C, "__getstate__", None
```
hynek, line 680: same as above

hynek (changes requested):
so um it doesn't look like we're testing here anything? those asserts needs to be fixed.

also can you pls pull in the changes from #969 so CI actually runs against 3.11? 😅

thanks!

tests/test_make.py
Tinche: But... that's what we want. If both don't exist, or both exist and are the same, the functionality is correct. Right?

hynek: uh…i guess i'll take your word for it? it just seems like there's cases where it passes when it shouldn't 🤔

just merge the other shit in then ¯\_(ツ)_/¯

Tinche:
> so um it doesn't look like we're testing here anything? those asserts needs to be fixed.
> 
> also can you pls pull in the changes from #969 so CI actually runs against 3.11? 😅
> 
> thanks!

This is a PR into #969, so we just merge it in? I have no idea why no tests are running here.

hynek:
ah sorry, i missed that!

PR #983, opened 2022-07-14
Use enum for typing NOTHING

KevinMGranger:
# Summary

This will allow those extending attrs to type `NOTHING` as `Literal[NOTHING]`.

This is the [recommended](https://github.com/python/typing/issues/689#issuecomment-561425237) way of doing this, at least until [PEP-661](https://peps.python.org/pep-0661/) lands.

I also just realized that `Attribute` types `default` as `Optional[_T]`, when it should really be `Union[_T, Literal[NOTHING]]`. This would be a prerequisite for that.

# Pull Request Check List


Tinche:
Hm, I don't hate it.

KevinMGranger:
Forgot to mention-- this is the [recommended](https://github.com/python/typing/issues/689#issuecomment-561425237) way of doing this, at least until [PEP-661](https://peps.python.org/pep-0661/) lands.

hynek:
I'm not really firm on these topics, so I'm assigning this to Tin. I hope the test_mypy failures are just caused by this? 😅

hynek:
Don't worry, it's not failing anymore. That was just mypy improving attrs support and subtly changing an error message. I suspect you haven't seen an error, because your local mypy was slightly older.

src/attr/__init__.pyi
```diff
@@ -65,7 +66,12 @@ class AttrsInstance(Protocol):
 
 # _make --
 
-NOTHING: object
+# Typed as enum member so that Literal[NOTHING] will work.
+@enum.Enum
+class _NothingType:
```
Tinche, line 71: The enum name is part of the repr, right? Can we name it something nicer, like `Defaults` or at least `NothingType` (without the underscore)?

Tinche (changes requested):

src/attr/__init__.pyi
KevinMGranger: Will do!

KevinMGranger: Or would you like me to override `__repr__` so it's just `NOTHING`?

Tinche: Good question. @hynek ?

Tinche: Ah I didn't notice it's just the stub. Does it need to be just the stub, though?

KevinMGranger: When I wrote it, it needed to be just the stub because of 2.7 support... [but I just saw it was recently dropped](https://www.attrs.org/en/stable/changelog.html#backwards-incompatible-changes). Excellent!

Now it can be changed to be the actual code. Now I'd recommend changing `__repr__` to be `'NOTHING'`, and making the class `_NothingType` again, to show that it's an implementation detail.

Tinche: Sounds good

hynek: i understanding nothing (lol) of what y'all are talking about, but a repr of `NOTHING` sounds great

KevinMGranger:
`NOTHING` is now the single variant of an Enum, keeping `__repr__` and `__bool__`'s function.

However, since a type stub file overrides any typing from the python module, the enum has to be (mostly) repeated in the stub. If we're fine with it, that's okay. If not, we can work around this by putting it in its own module, and reimporting it in `_make.py` and `__init__.pyi`.

I'm looking into how the enum affects the docs automation right now.

One last question: should I update

Tinche (approved):

Tinche:
@hynek I think this is ready to go. We we want a changelog entry?

hynek:
yeah, because it probably breaks someone too

ideally with a short summary of the upsides please because i keep forgetting myself

KevinMGranger:
The only upside is that it allows the use of `Literal[NOTHING]` in type signatures. That's useful for those extending attrs.

A project I'm working on does some of its own initialization logic, and inspecting `Attribute` properties to see if they're `NOTHING` is necessary. Using it as a [return type annotation](https://github.com/konveyor/pelorus/blob/77db9ecb28dc318b946307204088c29492c1026e/exporters/pelorus/config/loading.py#L187) helps us make sure it's [handled properly](https://github.com/konveyor/pelorus/blob/77db9ecb28dc318b946307204088c29492c1026e/exporters/pelorus/config/loading.py#L230) down the line.

(We have a compatibility shim that just [re-exports `NOTHING`](https://github.com/konveyor/pelorus/blob/77db9ecb28dc318b946307204088c29492c1026e/exporters/pelorus/config/_attrs_compat.py), but [type stubs it as an enum](https://github.com/konveyor/pelorus/blob/77db9ecb28dc318b946307204088c29492c1026e/exporters/pelorus/config/_attrs_compat.pyi))

KevinMGranger:
Ah, I forgot to revisit this one part:

Some of the current typings are technically incorrect. For example `Attribute.default` is typed as `Optional[_T]`. For example:

```python
@frozen
class Foo:
    bar: int = field()
    baz: int = field(default=2)
```

Someone inspecting the `Attribute[int]` for `bar` will have `default` typed as `Optional[int]`, and will be surprised to find a `NOTHING`! Similarly, inspecting `Attribute[int]` for `baz` will have someone handling the `None` case, even though it can't be `None`.

I'll have to see if there are other examples. If you want, this could be handled as part of this PR, or another. Like you said, it might break someone's code (but I think it'll mainly just change up the type checking they do). Doing it in this PR would mean they handle it all in one fell swoop. Thoughts?

Tinche:
Where are you seeing this `Optional[_T]` thing?

KevinMGranger:
In [`attrs.Attribute`'s type stub](https://github.com/python-attrs/attrs/blob/006b31f46c45c3c2c3491e99cf55c6347eb06ec0/src/attr/__init__.pyi#L115)}

hynek:
What are the next steps here?

Tinche:
We add a short changelog entry and merge this.

Tinche:
@hynek I've added a small changelog entry, merge at will.

hynek:
Thanks @KevinMGranger!

KevinMGranger:
I'm seeing really weird behavior. This works perfectly fine within `attrs`, but when importing it outside of `attrs`, mypy gives a weird complaint:

```python
# foo.py
from typing import Literal
from attrs import NOTHING

def foo(val: str | Literal[NOTHING]):
    pass
```
```console
$ mypy foo.py 
foo.py:4: error: Function is missing a return type annotation
foo.py:4: error: Parameter 1 of Literal[...] is invalid
foo.py:4: error: Variable "attr.NOTHING" is not valid as a type
foo.py:4: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
Found 3 errors in 1 file (checked 1 source file)
```
Anyone else seeing this? I'm trying to debug if this is a mypy issue.

KevinMGranger:
pyright doesn't complain, while mypy does. It looks like I misread [GVR's suggestion](https://github.com/python/typing/issues/689#issuecomment-561425237), and missed:

> (Alas, I haven't found a way to alias A = AA.A and be able to write Literal[A].)

There are two (and a half) options here:

1. Revert this change
2. Rename the enum class `Nothing`, use `Nothing` for type annotations, and `NOTHING` for the value. If that's too confusing for users, 1 is the better option.
3. Contribute to mypy so this works (kidding. maybe.)

Tinche:
I think we can leave it and submit a bug report to Mypy. Unfortunate.

Tinche:
Does it help if you annotate the alias with `Final`?

hynek:
Re: MyPy option: how broken is this? Is this gonna make me fix all my projects like the AttrsInstance thing in 22.1 or is this something that's only broken if you're trying to use the new feature it's supposed to implement?

Tinche:
Yeah, I think it's only broken if you try to Literal it, and hopefully Mypy fixes that.

PR #988, opened 2022-07-29
Drop Python 3.5

### LINKED ISSUE #965: Cannot import attrs for Python 3.5
First of all, attrs is great, thanks to all who work on it.

On Python 3.5, the next-gen API (`define`, `field`, etc.) is not defined in `attr/__init__.py`, which causes an ImportError on startup.  Here's a reproducer:

```
# python3
Python 3.5.3 (default, Dec  3 2018, 20:11:08) 
[GCC 7.3.0] on linux
>>> import attrs
  File "/usr/lib/python3.5/site-packages/attrs/__init__.py", line 3, in <module>
    from attr import (
ImportError: cannot import name 'define'
```

Looks like it's happening because of these lines in `attr/__init__`: https://github.com/python-attrs/attrs/blob/main/src/attr/__init__.py#L76-L79

along with  https://github.com/python-attrs/attrs/blob/main/src/attrs/__init__.py#L19 (which tries to import those names unconditionally).

hynek:
Less than 0.5% of our downloads are 3.5 and it allows us to simplify A LOT of code. Just look at all the beautiful (and fast) f-strings (they're courtesy of pyupgrade).

Fixes #965

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "3.11.0-beta - 3.11", "pypy-3.7", "pypy-3.8"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11.0-beta - 3.11", "pypy-3.7", "pypy-3.8"]
 
     steps:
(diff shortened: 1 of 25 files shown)


sumesh-decarate:
> Less than 0.5% of our downloads are 3.5 and it allows us to simplify A LOT of code. Just look at all the beautiful (and fast) f-strings (they're courtesy of pyupgrade).
> 
> Fixes #965

PR #991, opened 2022-08-02
Speedup `_setattr` usage and fix slight performance regressions

davfsa:
# Summary

Fix slight performance regression of #898.

I only noticed this after upgrading to 22.1.0, but the PR i made a while back to speedup the instantiation of frozen classes did so, but at a cost of not scaling correctly the more attributes that need to be set

Didnt want to just revert the commit, so played around a bit and found some optimizations. which are in this pr. They include (partially) going back to some of the old code, but with some optimizations that makes it both scale better and perform better in simple tests, ending up being a win win. Sorry for the slight performance regression this issue caused. It didnt cross to me it wouldnt scale well.

<details>
  <summary>3 arguments performance</summary>

### `attrs-21.4.0`
```
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True, frozen=True)" "C(1, 2, 3)"
Mean +- std dev: 499 ns +- 26 ns
```

### `attrs-22.1.0`
```
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True, frozen=True)" "C(1, 2, 3)"
Mean +- std dev: 450 ns +- 17 ns
```

### `this pr`
```
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['a', 'b', 'c'], slots=True, frozen=True)" "C(1, 2, 3)"
Mean +- std dev: 425 ns +- 16 ns
```
</details>

<details>
  <summary>10 arguments performace</summary>

### `attrs-21.4.0`
```
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'], slots=True, frozen=True)" "C(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)"
Mean +- std dev: 1.01 us +- 0.04 us
```

### `attrs-22.1.0`
```
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'], slots=True, frozen=True)" "C(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)"
Mean +- std dev: 1.18 us +- 0.04 us
```

### `this pr`
```
$ pyperf timeit --rigorous -s "import attr; C = attr.make_class('C', ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'], slots=True, frozen=True)" "C(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)"
Mean +- std dev: 941 ns +- 45 ns
```
</details>

# Pull Request Check List

### DIFF
--- a/changelog.d/991.change.rst
+++ b/changelog.d/991.change.rst
@@ -1,0 +1,1 @@
+Fix slight performance regression in classes with custom ``__setattr__`` and speedup even more.
--- a/docs/how-does-it-work.rst
+++ b/docs/how-does-it-work.rst
@@ -94,9 +94,9 @@ This is (still) slower than a plain assignment:
         -s "import attr; C = attr.make_class('C', ['x', 'y', 'z'], slots=True, frozen=True)" \
         "C(1, 2, 3)"
   .........................................
-  Mean +- std dev: 450 ns +- 26 ns
+  Mean +- std dev: 425 ns +- 16 ns
 
-So on a laptop computer the difference is about 230 nanoseconds (1 second is 1,000,000,000 nanoseconds).
+So on a laptop computer the difference is about 200 nanoseconds (1 second is 1,000,000,000 nanoseconds).
 It's certainly something you'll feel in a hot loop but shouldn't matter in normal code.
 Pick what's more important to you.
 
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -932,7 +932,7 @@ class _ClassBuilder:
             """
             Automatically created by attrs.
             """
-            __bound_setattr = _obj_setattr.__get__(self, Attribute)
+            __bound_setattr = _obj_setattr.__get__(self)
             for name, value in zip(state_attr_names, state):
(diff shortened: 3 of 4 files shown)


davfsa:
Also worth noting that frozen classes don't scale well when it comes to attribute setting. When comparing setting 3 attributes in frozen and unfrozen slotted classes, you get a difference of about 200 nanoseconds. When you expand the test to 10 attributes, you get a difference of almost 500! (943 vs 445). 

Looks like the performance of making so many calls is not ideal, but I can't think of a better way of doing this without really stretching the limits of python.

hynek (approved):
Thanks, I’m gonna have to take your word for it here. 😅

Issue #987, opened 2022-07-28
Type check error with asdict 


dbertouille:
Starting with version attrs 22.1.0, I started running into a type checking issue when running mypy. In particular, this occurs when calling asdict in parent class where that parent class does not use attrs but the child might. 

This is a bit of an odd scenario so I'm not sure if the bug belongs here in attrs, mypy, or should just be an error. However, it did not produce any type checking errors using attrs 21.4.0.

```python
import abc
import attr
from typing import Any, Dict


class BaseModel(abc.ABC):
    def to_dict(self) -> Dict[str, Any]:
        o = {}  # type: Dict[str, Any]
        if attr.has(type(self)):
            return attr.asdict(self)
        return vars(self).copy()


@attr.s(auto_attribs=True)
class Model(BaseModel):
    id: int = 0


model = Model(id=1)

if attr.has(type(model)):
    data = attr.asdict(model)
```

hynek:
Given that your asdict is protected by a `has`, I suspect that belongs into mypy because it's dynamic logic we can't express in attrs. But @Tinche might know more.

The problem is that we introduced an AttrsInstance Protocol. It looks like we forgot to add a newsfragment.

1oglop1:
@hynek I notice that `AttrsInstance` is not imported in `attrs.__init__.pyi` only defined in `attr.__init__.pyi` is that intentional?

hynek:
@Tinche did we just forget it?

Tinche:
Yeah I think so ;/

layday:
Probably worth nothing that `has` is a perfect match for `TypeGuard` from Python 3.10 (or its backport from `typing-extensions`). In `attrs/__init__.pyi`:

```py
def has(cls: type) -> TypeGuard[type[AttrsInstance]]: ...
```

Elsewhere:

```py
[...]
        if attr.has(type(self)):
            # `self` is now an `AttrsInstance`
            return attr.asdict(self)  # Passes!
```

But I suspect that `attrs` would not want to depend on `typing-extensions`.

hynek:
fixed by #996

PR #994, opened 2022-08-06
Re-import AttrsInstance in attrs ns

hynek:
cf. https://github.com/python-attrs/attrs/issues/987#issuecomment-1206077089


Tinche (approved):

Tinche:
I need to fix that Mypy caching bug for the magic attribute, don't I?

hynek:
If you want it fixed… 😇

Tinche:
If I don't fix it, mypy won't recognize the magic attribute and the protocol will be useless. 😭

PR #995, opened 2022-08-07
Eliminate most str.format() and %-formatting

hynek:
Micro-optimization is the root of all success.


hynek:
This btw gives us ~6% performance gain on class creation for _simple_ classes, just by being more idiomatic (and getting rid of the mistake called format).

Measured using:

```python
def create(loops):
    range_it = range(loops)
    d = attrs.define

    t0 = pyperf.perf_counter()

    for _ in range_it:
        @d
        class C:
            a: int
            b: int

    return pyperf.perf_counter() - t0
```

---

I presume that more complex classes give us even bigger gainzzz.

Tinche (approved):

Issue #996, opened 2022-08-08
Use of `AttrsInstance` protocol is incompatible with type checkers other than mypy


layday:
The latest release introduced the `AttrsInstance` protocol with one attribute, `__attrs_attrs__`, which replaced `Any` in the signature of various core functions.  The `__attrs_attrs__` attribute is "injected" by the attrs mypy plug-in; other type checkers have no knowledge of this attribute and adding attributes to the type definition of a dataclass-like class is not supported by `dataclass_transform`.  Therefore, functions such as `fields` and `asdict` generate type errors by default in all type checkers other than mypy.  I am not familiar with mypy plug-ins but I assume one alternative would be to remove the attribute from the protocol definition and inject it dynamically from the plug-in, the way it's done for the attrs class.  Longer term, an extension to `dataclass_transform` could be sought.

Tinche:
Pyright has no plugin interface, so there's not much we can do. Maybe you can open an issue over there?

layday:
It's a topic that's been discussed repeatedly - Pyright is opposed to adding support for plug-ins.  Their resistance to plug-ins is `dataclass_transform`'s raison d'etre.

hynek:
I'm really not sure what to do about all of this. pyright has become a bottle-neck in typing and compared to it, nobody uses mypy. We feel that at all ends.

layday:
I'm not sure I agree.

Pyright has a large audience but the majority of it I would say is Pylance code completion users - Python beginners, hobbyists, data scientists and the like - or even software devs who only use it for completion in their editor.  OTOH, mypy is a mainstay of Python software developers and you won't see Pyright being used to type check any of the popular (and slightly less popular) libraries here on GH, either in CI or in devel.[^1]  I don't know what the situation's like in corporate - things might swing slightly in Pyright's favour - but I don't think mypy's anywhere near the point where no one's using it.

Pyright has been quick to adopt new standards and was (to my knowledge) the only type checker to implement the now rejected short form callable definition PEP.  Its main author has also been involved in writing new typing PEPs and has been a very positive influence in the typing community, from what I've seen.  I think bottleneck is a slightly unfair characterisation.

I don't think framing this in terms of Pyright vs Mypy is particularly insightful either; there are other type checkers out there, either without plug-in support without a plug-in for attrs.

[^1]: I think, primarily, because it's written in JS.

layday:
[...]

PR #997, opened 2022-08-08
Use `TypeGuard` for `has` in Python 3.10 and above

layday:
# Summary

Conditionally defines `has` using `TypeGuard` in Python 3.10+ in the stub file.

xref: https://github.com/python-attrs/attrs/issues/987

# Pull Request Check List


Tinche:
I don't see a direct use-case for this (since attrs classes should be statically known now) but I don't see the harm either.

layday:
The class might not be statically known to be an attrs class (#987, #996) or it can appear in a union with other types.

hynek:
So what exactly does this do?

Tinche:
It essentially makes `has` work at compile time too.

Let's say your function takes a class (not an instance, but an actual class). You will type your argument as `type` (or `type[C]`). You can check if it's an attrs class with `has`, and if that's true you can pass it to a different function that takes `type[AttrsInstance]`. But Mypy will still yell at you, since it doesn't know that `has` means it's an attrs class.

If we annotate `has` with `TypeGuard`, Mypy will know.

Code example:

```
def needs_attrs_class(cls: type[AttrsInstance]) -> None:
    pass

def my_function(cls: type) -> None:
    if has(cls):
        needs_attrs_class(cls)  # Mypy's gonna be unhappy, unless it knows `has` narrows the type of is argument
```

layday:
It tells the type checker that if the condition is true then `cls` is of type `type[AttrsInstance]`. Simplified example:

```py
def is_dict_no_typeguard(value: object) -> bool:
    return isinstance(value, dict)

def is_dict_typeguard(value: object) -> TypeGuard[dict]:
    return isinstance(value, dict)

might_be_a_dict = ...

if is_dict_no_typeguard(might_be_a_dict):
    "type of might_be_a_dict is object"

if is_dict_typeguard(might_be_a_dict):
    "type of might_be_a_dict is dict"


```

hynek:
Sounds cool! Any reason against it? If not: make it happen Tin. ;)

src/attr/__init__.pyi
```diff
@@ -470,7 +470,15 @@ def astuple(
     tuple_factory: Type[Sequence[Any]] = ...,
     retain_collection_types: bool = ...,
 ) -> Tuple[Any, ...]: ...
-def has(cls: type) -> bool: ...
+
+if sys.version_info >= (3, 10):
```
hynek, line 474: Would it be feasible to add a conditional import from typing-extensions?

layday: You can only use one of a small number of constants like `TYPE_CHECKING` and `sys.version_info` to define types conditionally, try/except won't work, so I don't think so.

Tinche: In my very limited testing this seems to work, both in Mypy and Pyright:

Screenshot 2022-08-15 at 13 37 05

layday: Right, but typing_extensions is not a dependency of attrs, so that'll error if typing_extensions is not installed.

layday: Okay, it seems typeshed ships typing_extensions is part of the stdlib, so I guess we can actually do that.

layday: Done in https://github.com/python-attrs/attrs/pull/997/commits/fd2c819ab04bbfe4785ee6abbc9c42a8f566dc22.

hynek:
Would you mind coming up with a way to verify this works in https://github.com/python-attrs/attrs/blob/main/tests/typing_example.py ?

layday:
There's a test in `test_mypy.yml`.  Should I move it over to `typing_example.py`?

hynek:
No, both would be great, since they work somewhat differently. Maybe `if has(c): print(c.__attrs_attrs__)` or something like that? That should pass, right?

Tinche (approved):

hynek:
Just so I understand this correctly: the import from typing_extensions only works, because we have the stubs in pyi files that are only touched by mypy which in turn brings in typing_extensions? IOW: if we'd use inline type hints, this would fall apart?

layday:
That's right. The typeshed has [`typing_extensions`](https://github.com/python/typeshed/blob/master/stdlib/typing_extensions.pyi) in the stdlib.  I assume they did this so that they can use new typing constructs in the typeshed without tripping MyPy.  At runtime, you'd have to do something like:

```py
import typing

if typing.TYPE_CHECKING:
    from typing_extensions import TypeGuard

# Use TypeGuard either in quotes or with `from __future__ import annotations`
```

hynek:
Thanks!

PR #1009, opened 2022-08-16
Change __getstate__ and __setstate__ to use a dict (#1004)

### LINKED ISSUE #1004: Current default implementation of __getstate__ and __setstate__ could be made safer
This is a known "sharp edge" of pickle, but attrs could make this a bit safer by slightly modifying the default implementation of `__getstate__` and `__setstate__`. The problem is that due to returning a tuple in the default implementation of `__getstate__`, removing a member, and then unpickling from a previous version can be __very__ unsafe. This is not a hypothetical situation - it commonly happens when an object is pickled, and then stored in a data-store of some sort, and unpickled some time later by a changed version of the code. 

Here's a simple reproducer that demonstrates it (python 3.10, attrs 22.1.0):

```python
import pickle

import attr

@attr.s(slots=True, hash=False, auto_attribs=True)
class Test:
    count: int
    enable_minor_feature: bool
    should_launch_missiles: bool

t = Test(count=1, should_launch_missiles=False, enable_minor_feature=True)
print(t)  # Test(count=1, enable_minor_feature=True, should_launch_missiles=False)
tp = pickle.dumps(t)

@attr.s(slots=True, hash=False, auto_attribs=True)
class Test:
    count: int
    should_launch_missiles: bool

tl = pickle.loads(tp)
print(tl)  # Test(count=1, should_launch_missiles=True)  <== an attribute is assigned a dangerously wrong value
```

While there's certainly an argument to be made that pickle should not be used for such things (and I'd agree) - the default implementation could, I believe, be made safer. Raise if things don't match, or even ignore unknown attributes in __setstate__,  by potentially returning a dictionary instead of a tuple [here](https://github.com/python-attrs/attrs/blob/c58ffd4e4cba5d5e58356722b985fc362358c48e/src/attr/_make.py#L925) would help.

djipko:
# Summary

We use a dict instead of a tuple, so that we assign only existing attributes on unpickle, if the class
definition has changed in the meantime.

Resolves #1004

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -922,7 +922,7 @@ class _ClassBuilder:
             """
             Automatically created by attrs.
             """
-            return tuple(getattr(self, name) for name in state_attr_names)
+            return {name: getattr(self, name) for name in state_attr_names}
 
         hash_caching_enabled = self._cache_hash
 
@@ -931,8 +931,9 @@ class _ClassBuilder:
             Automatically created by attrs.
             """
             __bound_setattr = _obj_setattr.__get__(self)
-            for name, value in zip(state_attr_names, state):
-                __bound_setattr(name, value)
+            for name in state_attr_names:
+                if name in state:
+                    __bound_setattr(name, state[name])
 
             # The hash code cache is not included when the object is
             # serialized, but it still needs to be initialized to None to
--- a/tests/test_slots.py
+++ b/tests/test_slots.py
@@ -9,6 +9,8 @@ import sys
 import types
 import weakref
 
+from unittest import mock
+
 import pytest
 
 import attr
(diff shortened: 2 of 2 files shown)


hynek:
Looks like we're missing a branch: https://github.com/python-attrs/attrs/runs/7863540189?check_suite_focus=true#step:6:22

hynek:
I think it’s actually an interesting case missing: what are we gonna do, if there’s extra names? By your motivation, we should actually blow up, no?

djipko:
> I think it’s actually an interesting case missing: what are we gonna do, if there’s extra names? By your motivation, we should actually blow up, no?

So the case missing is when what we re-hydrate has less attributes than our current class implementation declares, and we have to leave some of them empty.  So that's an interesting one. I don't think we should necessarily blow up in every case, as there's some chance calling code can handle this (say, check for `None` or `hasattr`). The difference with the reported case and this is that in the former, the calling code stands no chance, it gets a completely wrong value (and why I thought we should fix it). 

I think failing early (on unpickling) is tempting here, but might be too aggressive. It is likely that the calling code that doesn't handle this well and tries to access a non existing attribute will blow up anyway. Another potential solution would be to check if we have an `.init` on the attribute and assign that but even that seems over-reaching as it re-implements parts of initialisation of the class which is clearly skipped when unpickling by design. 

At the end of the day, the point of this is to avoid a clear (if a bit obscure :) ) foot gun, and not re-implement versioning on top of pickle. People who actually need and want that can use protobuf or a number of actual solutions to this problem. So I would go with - just skip it (as it would have before). I will add a test

djipko:
```
$ tox -e coverage-report
coverage-report installed: coverage==6.4.4,tomli==2.0.1
coverage-report run-test-pre: PYTHONHASHSEED='1249099045'
coverage-report run-test: commands[0] | coverage combine
Combined data file .coverage.dubndipanovmacbook1.87338.100213
Combined data file .coverage.dubndipanovmacbook1.87526.556153
coverage-report run-test: commands[1] | coverage report
Name    Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------
---------------------------------------------------
TOTAL    1494      0    723      0   100%

19 files skipped due to complete coverage.
______________________________________________________________________ summary _______________________________________________________________________
  coverage-report: commands succeeded
  congratulations :)
```

Got this locally now so feeling hopeful 🤞 😅

hynek (approved):
Cool, thanks! I'll write a news fragment.

Mirandatz:
I feel like this may be a breaking change. Do objects serialized with the previous version can be deserialized with the new version?

hynek:
They cannot. A fix has been added in #1085 that will be in 23.1, but (de)serializing classes across versions is risky (in general, not just the _attrs_ case) so we'd encourage to look into avoiding that.

PR #1010, opened 2022-08-17
Add a not_ validator

nicktimko:
# Summary

Still a bit of a WIP.

My original motivation for this validator was to enable constructions like `not_(in_(["some", "reserved", "keywords"]))`, and rather than duplicating and swapping `in` for `not in` in the `in_` validator, figured I could make it composable.

# Pull Request Checklist

    - [x] ~Changes to the signature of `@attr.s()` have to be added by hand too.~


hynek:
Yeah sorry, this is a no-go. We can't just blanketly eat exceptions. Sorry for suggesting that! Feel free to submitting a not_in PR.

nicktimko:
I guess I'm not sure how it "blanketly eats" the exceptions. I view validators as effectively returning a boolean: `None` or `raise ValueError` (raising being convenient as it can bubble up through a call stack without a bunch of extra handling). I suppose if you had an unexpected `ValueError` raised within the `not`-wrapped validator, it would be mysterious if it *did* successfully validate after getting inverted by the `not`... However, in that case it's the dual of testing some input *should* fail validation and it's *not* failing, so your test fails without an exception. You'll need to do some extra debugging in either case.

As I mentioned, `not_in` was good enough for my use case, but I've also used `not re.match(...)` and `not isinstance(...)` in other places (though not as attrs validators). I can prep a `not_in` PR.

hynek:
> I view validators as effectively returning a boolean: `None` or `raise ValueError` (raising being convenient as it can bubble up through a call stack without a bunch of extra handling).

This is not a safe assumption to make – validators can raise *any* exception and the fact that our internal ones usually raise ValueErrors (I think there’s a TypeError in there too tho?) is not a general rule.

Since attrs is not a validation library, but tries to help you write classes like you would, I *encourage* people to use any exception that makes sense to them.

***

That said, and after thinking  some more about it, I could live with a `not_`, if it handles ValueError and TypeError and is super clear about that. I’ve just noticed you did the latter part and I have to admit, that I stopped reading after I saw the except ValueError. I’ll re-open and comment inline.

changelog.d/1010.change.rst
```diff
@@ -0,0 +1 @@
+Added ``attr.validators.not_()`` to logically invert a validator.
```
hynek, line 1:
```suggestion
Added ``attrs.validators.not_(wrapped_validator)`` to logically invert *wrapped_validator* by accepting only values where *wrapped_validator* raises a ``ValueError`` or a ``TypeError``.
```

src/attr/validators.py
```diff
@@ -592,3 +593,41 @@ def min_len(length):
...
+
+
+@attrs(repr=False, slots=True, hash=True)
+class _NotValidator:
+    validator = attrib()
+
+    def __call__(self, inst, attr_, value):
+        try:
+            self.validator(inst, attr_, value)
+        except ValueError:
```
hynek, line 605:
```suggestion
        except (ValueError, TypeError):
```

src/attr/validators.py
```diff
@@ -592,3 +593,41 @@ def min_len(length):
...
+
+    def __repr__(self):
+        return "<not_ validator wrapping {what}>".format(
+            what=repr(self.validator)
+        )
+
+
+def not_(validator):
+    """
+    A validator that wraps and logically 'inverts' the validator passed to it,
```
hynek, line 619: This needs to be expanded to TypeErrors. Sorry that it disturbs the nice flow of your sentence. :)

hynek: Also idle though: this could be made configurable if the need ever arises. As in you pass in exactly the exception you want to require.

src/attr/validators.py
```diff
@@ -592,3 +593,41 @@ def min_len(length):
...
+
+    Intended to be used with existing validators to compose logic without
+    needing to create inverted variants, for example, ``not_(in_(...))``.
+
+    :param validator: A validator to be logically inverted.
+
+    :raises ValueError: If the wrapped validator does not raise a
+        `ValueError`, this validator will.
+
+    .. versionadded:: just a PR for now.
```
hynek, line 631:
```suggestion
    .. versionadded:: 22.2.0
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+
+
+class TestNot_:
+    """
+    Tests for `not_`.
+    """
+
+    def test_not_all(self):
+        """
+        Verify that this validator is in ``__all__``.
```
hynek, line 1077:
```suggestion
        The validator is in ``__all__``.
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+        """
+        v = not_(in_("abc"))
+        a = simple_attr("test")
+        with pytest.raises(ValueError) as e:
+            v(None, a, "b")
+        assert (
+            "not_ validator child '{!r}' did not raise error".format(
+                in_("abc")
+            ),
+        ) == e.value.args
```
hynek, line 1131:
```suggestion
        a = simple_attr("test")

        with pytest.raises(ValueError) as e:
            v(None, a, "b")

        assert (
            "not_ validator child '{!r}' did not raise error".format(
                in_("abc")
            ),
        ) == e.value.args
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+
+    def test_composable_with_matches_re_pass(self):
+        """
+        Check something does not match a regex.
+        """
+        v = not_(matches_re("[a-z]{3}"))
+        a = simple_attr("test")
+        v(None, a, "spam")
```
hynek, line 1139:
```suggestion

        v(None, a, "spam")
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+        """
+        v = not_(matches_re("[a-z]{3}"))
+        a = simple_attr("test")
+        with pytest.raises(ValueError) as e:
+            v(None, a, "egg")
+        assert (
+            "not_ validator child '{!r}' did not raise error".format(
+                matches_re("[a-z]{3}")
+            ),
+        ) == e.value.args
```
hynek, line 1153:
```suggestion
        a = simple_attr("test")

        with pytest.raises(ValueError) as e:
            v(None, a, "egg")

        assert (
            "not_ validator child '{!r}' did not raise error".format(
                matches_re("[a-z]{3}")
            ),
        ) == e.value.args
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+
+    def test_repr(self):
+        """
+        Returned validator has a useful `__repr__`.
+        """
+        wrapped = in_([3, 4, 5])
+        v = not_(wrapped)
+        assert (f"<not_ validator wrapping {repr(wrapped)}>") == repr(v)
```
hynek, line 1161:
```suggestion
        wrapped = in_([3, 4, 5])

        v = not_(wrapped)

        assert (f"<not_ validator wrapping {repr(wrapped)}>") == repr(v)
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+
+        v = not_(always_passes)
+        a = simple_attr("test")
+        with pytest.raises(ValueError) as e:
+            v(1, a, 3)
+        assert (
+            "not_ validator child '{!r}' did not raise error".format(
+                always_passes
+            ),
+        ) == e.value.args
```
hynek, line 1109:
```suggestion

        with pytest.raises(ValueError) as e:
            v(1, a, 3)

        assert (
            "not_ validator child '{!r}' did not raise error".format(
                always_passes
            ),
        ) == e.value.args
```

tests/test_validators.py
```diff
@@ -1064,3 +1065,97 @@ def test_repr(self):
...
+        """
+        If the wrapped validator fails, we're happy.
+        """
+
+        def always_fails(inst, attr, value):
+            raise ValueError("always fails")
+
+        v = not_(always_fails)
+        a = simple_attr("test")
+        v(1, a, 3)
```
hynek, line 1091:
```suggestion
        a = simple_attr("test")

        v(1, a, 3)
```

hynek (changes requested):
Thanks, sorry for the detour. It’s almost there, this should be quick to merge.

nicktimko:
Set the default captured exceptions to `(ValueError, TypeError)` as suggested, and allowed whatever exceptions to be provided as the second argument to `not_`.  

In order to validate them, I maybe went a bit overboard and modified the `_InstanceOf` to `_SubclassOf`. Could skip validation entirely I suppose like `instance_of`, which doesn't really care what's passed in, or rather, it doesn't do any 'compose time' validation, it just passes it off to `isinstance(obj, ____)` at 'check time'.

src/attr/validators.py
```diff
@@ -592,3 +593,103 @@ def min_len(length):
...
+
+    def __call__(self, inst, attr_, value):
+        try:
+            self.validator(inst, attr_, value)
+        except self.exc_types:
+            pass  # suppress error to invert validity
+        else:
+            raise ValueError(
+                "not_ validator child '{validator}' did not raise "
+                "a captured error".format(validator=repr(self.validator))
```
hynek, line 660: two things:

1. I know we forgot for a few validators, but all validators should also raise at least the attribute as the second argument. Ideally also what it expects and what it got instead but I'm not sure if that makes any sense here – you could make the validator an arg tho!
2. Should we maybe make the error message configurable? The current one can't be exposed to anyone using the class. E.g. `not_(instance_of(str), msg="x must not be a str.")`

nicktimko: I'll work on this over the weekend. I was just working from another validator's code, but I'll try to replicate a more recent one.

Was the `subclass_of` stuff abhorrent, or OK to leave in (just using it for 'internal-use' validation)?

hynek: I can live with the subclass stuff if it stays internal!

nicktimko:
Should be good for re-review. I have two lingering errors I don't know how to fix, first is with the typing, where mypy is giving me:

```
tests/typing_example.py:214: error: Argument "exc_types" to "not_" has incompatible type 
"Type[ValueError]"; expected "Union[Type[type], Iterable[type]]"
```

The other is Sphinx complaining that my type declaration of `:type exc_types: Exception type or an iterable of Exception types.` isn't a thing. Could just drop that I guess? (edit: I did)

src/attr/validators.py
```diff
@@ -592,3 +594,117 @@ def min_len(length):
...
+                    actual=value.__class__,
+                    value=value,
+                ),
+                attr,
+                self.type,
+                value,
+            )
+
+    def __repr__(self):
+        return "<subclass_of validator for type {type!r}>".format(
```
hynek, line 622: ci is failing because this line is missing coverage. 🤓

nicktimko:
Will fix. Sorry to 🚲🏠, but another edit I'm noticing as I'm looking at what to put in the exception args, various validators seem to have different ideas:

* `instance_of`, `matches_re`, `provides`, `in_`
    * `attr`
    * a parameter of the validator
    * `value` that failed validation
* numbers (`lt`, `le`, `gt`, `ge`)
    * No additional args beyond a formatted message
* `is_callable`
    * Custom Error type
    * `value` as kwarg

Should replicate the top as you suggested and have the docs look like the others, e.g.:

>  ... a human readable error message, the attribute (of type `attrs.Attribute`), the expected (interface|type|...), and the value it got.

The 'expected' part should probably be the wrapped validator that should have failed, but is it worthwhile tacking on the captured errors? As `arg[4]` at the end, or before "the value it got." and keep that as `arg[-1]`

src/attr/validators.py
nicktimko: Can I get privileges to run CI? 😉

hynek: crypto bros ruined that for everybody :( you’ll be in after this is merged

hynek: (takes one merged PR)

nicktimko:
Anything else to do? The only checkmark I'm not sure about is the hypothesis testing: none of the other validators have it, and when I started on it, it felt very much like I was reimplementing the same logic to derive what should be expected from various permutations received.

hynek (approved):
Alright thanks!

One minor nit that I'll fix myself: don't put default values into type stubs but a `...`.

PR #1011, opened 2022-08-17
Use pytest-xdist in tox / CI

hynek:
Locally, this slashes over 70% of runtime.

Let's see what it does to CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Make coverage unique if it exists.
         run: |
           if [ -f .coverage ]; then
-            mv .coverage .coverage.$(uuidgen)
+            mv .coverage .coverage.${{ strategy.job-index }}
           fi
 
       - name: Upload coverage data
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,6 +40,8 @@ jobs:
 
       - run: python -m tox
 
+      - run: ls -al
+
       - name: Upload coverage data
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,14 +49,6 @@ jobs:
 
       - run: python -m tox
 
-      # pytest-cov automatically merges coverage into `.coverage`, so we have to
-      # make it unique before uploading.
-      - name: Make coverage unique if it exists.
-        run: |
-          if [ -f .coverage ]; then
-            mv .coverage .coverage.${{ strategy.job-index }}
-          fi
-
       - name: Upload coverage data
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,12 +51,22 @@ jobs:
 
       - run: ls -al
 
+      - name: Make coverage unique if it exists.
+        run: |
+          COVERAGE_UUID=$(python3 -c "import uuid; print(uuid.uuid4())")
+          echo "::set-output name=COVERAGE_UUID::${COVERAGE_UUID}"
+          if [ -f .coverage ]; then
+            mv .coverage .coverage.${COVERAGE_UUID}
+          fi
+
       - name: Upload coverage data
         uses: actions/upload-artifact@v3
         with:
           name: coverage-data
           path: .coverage.*
           if-no-files-found: ignore
+
+      - run: ls -al
 
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,14 +49,12 @@ jobs:
 
       - run: python -m tox
 
-      - run: ls -al
-
+      # pytest-cov automatically merges coverage into `.coverage`, so we have to
+      # make it unique before uploading.
       - name: Make coverage unique if it exists.
         run: |
-          COVERAGE_UUID=$(python3 -c "import uuid; print(uuid.uuid4())")
-          echo "::set-output name=COVERAGE_UUID::${COVERAGE_UUID}"
           if [ -f .coverage ]; then
-            mv .coverage .coverage.${COVERAGE_UUID}
+            mv .coverage .coverage.$(uuidgen)
           fi
 
       - name: Upload coverage data
@@ -65,8 +63,6 @@ jobs:
           name: coverage-data
           path: .coverage.*
           if-no-files-found: ignore
-
-      - run: ls -al
 
   coverage:
     runs-on: ubuntu-latest
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,10 @@ commands = python -m pytest {posargs:-n auto}
 [testenv:py36]
 extras = tests
 deps = coverage[toml]>=5.3
-setenv = COVERAGE_FILE = .coverage.{envname}
+setenv =
+    COV_CORE_SOURCE={toxinidir}/src
+    COV_CORE_CONFIG={toxinidir}/pyproject.toml
+    COV_CORE_DATAFILE={toxinidir}/.coverage.{envname}
 commands = pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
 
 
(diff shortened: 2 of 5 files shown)


tests/test_pyright.py
```diff
@@ -20,9 +21,11 @@ class PyrightDiagnostic:
 
 
 @pytest.mark.skipif(not _found_pyright, reason="Requires pyright.")
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="Requires Python 3.7+")
```
hynek, line 24: I had to add this because subprocess uses 3.7 and this tests breaks if there's a global pyright.

hynek:
OK looks like pytest-xdist doesn't work with coverage:

- https://stackoverflow.com/questions/72840168/how-to-run-coverage-report-with-parallel-pytest-using-xdist-and-django-coverage
- https://github.com/pytest-dev/pytest-cov/issues/129

Bummer but nothing we can do about I guess.

tox.ini
```diff
@@ -27,24 +27,26 @@ commands =
...
-commands = python -m pytest {posargs}
+commands = python -m pytest {posargs:-n auto}
 
 
 [testenv:py36]
 extras = tests
 deps = coverage[toml]
-commands = coverage run -m pytest {posargs}
+setenv = COVERAGE_FILE = .coverage.{envname}
+commands = python -m pytest --cov --cov-append --dist load {posargs:-n auto}
```
webknjaz, line 37: Try explicitly pre-loading the pytest-cov plugin. This way, it's guaranteed to be loaded before all others. It should help defeat the most common cause of the `Coverage.py warning: No data was collected` warning.
```suggestion
commands = python -m pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
```

tox.ini
```diff
@@ -27,24 +27,26 @@ commands =
...
 # PYTHONWARNINGS=d and --no-compile below make them visible during the Tox run.
-install_command = pip install --no-compile {opts} {packages}
+install_command = python -m pip install --no-compile {opts} {packages}
 setenv =
     PYTHONWARNINGS=d
+    COVERAGE_FILE = .coverage.{envname}
 extras = tests
 deps = coverage[toml]
-commands = coverage run -m pytest {posargs}
+commands = python -m pytest --cov --cov-append --dist load {posargs:-n auto}
```
webknjaz, line 49:
```suggestion
commands = python -m pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
```

tox.ini
hynek: nope: https://github.com/python-attrs/attrs/runs/7911673793?check_suite_focus=true :(

ionelmc:
Try this (**note it's just for py310**, some copypaste/fixing required for others):

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 024d1df..28ce8cb 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,6 @@ filterwarnings = ["once::Warning", "ignore:::pympler[.*]"]

 [tool.coverage.run]
 parallel = true
-concurrency = ["multiprocessing"]
 branch = true
 source_pkgs = ["attr", "attrs"]

diff --git a/tox.ini b/tox.ini
index 00d3f36..18a5f60 100644
--- a/tox.ini
+++ b/tox.ini
@@ -43,7 +43,9 @@ commands = pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
 install_command = python -m pip install --no-compile {opts} {packages}
 setenv =
     PYTHONWARNINGS=d
-    COVERAGE_FILE = .coverage.{envname}
+    COV_CORE_SOURCE={toxinidir}/src
+    COV_CORE_CONFIG={toxinidir}/pyproject.toml
+    COV_CORE_DATAFILE={toxinidir}/.coverage.{envname}
 extras = tests
 deps = coverage[toml]>=5.3
 commands = pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
```

This is in fact documented at https://pytest-cov.readthedocs.io/en/latest/plugins.html, maybe poorly - looking for feedback on how to make this easier to get to. Conftests do act as plugins, maybe it's a bit surprising.

An alternative is to avoid importing your package code in conftest:

```diff
diff --git a/conftest.py b/conftest.py
index 144e5f3..ba1d60b 100644
--- a/conftest.py
+++ b/conftest.py
@@ -1,10 +1,11 @@
 # SPDX-License-Identifier: MIT
+import sys

 import pytest

 from hypothesis import HealthCheck, settings

-from attr._compat import PY310
+PY310 = sys.version_info[:2] >= (3, 10)


 @pytest.fixture(name="slots", params=(True, False))
diff --git a/pyproject.toml b/pyproject.toml
index 024d1df..28ce8cb 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,6 @@ filterwarnings = ["once::Warning", "ignore:::pympler[.*]"]

 [tool.coverage.run]
 parallel = true
-concurrency = ["multiprocessing"]
 branch = true
 source_pkgs = ["attr", "attrs"]

diff --git a/tox.ini b/tox.ini
index 00d3f36..5c2c409 100644
--- a/tox.ini
+++ b/tox.ini
@@ -43,7 +43,6 @@ commands = pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
 install_command = python -m pip install --no-compile {opts} {packages}
 setenv =
     PYTHONWARNINGS=d
-    COVERAGE_FILE = .coverage.{envname}
 extras = tests
 deps = coverage[toml]>=5.3
 commands = pytest -p pytest_cov --cov --cov-append --dist load {posargs:-n auto}
```

Note that if you override COVERAGE_FILE you mess with pytest-cov. You might need to do some renaming before actions/upload-artifact runs to get this fully working.

Another note is that you should not use the multiprocessing option, it does nothing for xdist and only works half of the time for serious multiprocessing code anyway. I wouldn't want to have multiprocessing imported and patched around for nothing if you'd ask my opinion :-)

hynek:
@ionelmc thanks!

Looks like `COV_CORE_DATAFILE={toxinidir}/.coverage.{envname}` doesn’t do anything?

At least in  https://github.com/python-attrs/attrs/runs/7929941929?check_suite_focus=true#step:6:18 and https://github.com/python-attrs/attrs/runs/7929941812?check_suite_focus=true#step:6:18. there’s `.coverage` files.

ionelmc:
It only looks like it doesn't because that file and the files for each xdist worker are combined (pytest-cov does it for you) before pytest exits. 

So I guess now you only need to figure out how to make actions/upload-artifact upload that `.coverage` file with a envname sufix, right?

hynek:
Yeah, that’s not super hard. I suppose there’s no way to prevent pytest-cov from combining tho? How does that work locally over multiple tox runs? That’s what append is for?

Does it make any sense to even set the option?

ionelmc:
Yes that is what the append is for. There's no way to disable combining - there could be but then pytest-cov wouldn't do much of anything at all with combining disabled. So it's pointless to have that option.

hynek:
Great success, thanks @ionelmc!!!

Now can we simplify it a bit?

do we need:

- `--dist load`
- The `COV_CORE_` arguments, _especially_ `COV_CORE_DATAFILE` which seems to be only a tmp file?

I guess I could try them all, but that's a lot of CI runtime, so a pointer would be appreciated. :)

ionelmc:
At minimum COV_CORE_SOURCE is required, as you can see here: https://github.com/pytest-dev/pytest-cov/blob/master/src/pytest-cov.embed#L1

As I understand it `--dist=load` is the default: https://pypi.org/project/pytest-xdist/#running-tests-across-multiple-cpus

hynek:
FYI, they’re all required. In the end I’ve spent all day going thru the permutations anyways.

Thanks for your help @ionelmc (and thank you for sponsoring me!!! 💛).

.github/workflows/main.yml
```diff
@@ -49,6 +49,14 @@ jobs:
 
       - run: python -m tox
 
+      # pytest-cov automatically merges coverage into `.coverage`, so we have to
+      # make it unique before uploading.
+      - name: Make coverage unique if it exists.
+        run: |
+          if [ -f .coverage ]; then
+            mv .coverage .coverage.$(uuidgen)
```
webknjaz, line 57: https://docs.github.com/en/actions/learn-github-actions/contexts#strategy-context suggests you use
```suggestion
            mv .coverage .coverage.${{ strategy.job-index }}
```
here

PR #1016, opened 2022-08-21
Add a preview for the "next release" change notes

webknjaz:
# Summary

This change integrates `sphinxcontrib-towncrier` into the docs build.
It uses Towncrier and its configs to render a draft changelog as RST
and injects that into the `changelog.rst` document right above the
older changelog entries using `towncrier-draft-entries` directive.

I figured, it'd be nicer to be able to have a sneak-peak into the future
release changes. As an alternative to navigating to a folder with a
bunch of standalone uncategorized files.

Usage in the wild: https://pip.pypa.io/en/latest/news/#not-yet-released-2022-08-21.

Preview: https://attrs--1016.org.readthedocs.build/en/1016/changelog.html#changelog.

### DIFF
--- a/changelog.d/1016.doc.rst
+++ b/changelog.d/1016.doc.rst
@@ -1,1 +1,0 @@
-Render the unreleased ``attrs`` version change notes on the changelog page in docs.
--- a/changelog.d/1016.doc.rst
+++ b/changelog.d/1016.doc.rst
@@ -1,0 +1,1 @@
+Render the unreleased ``attrs`` version change notes on the changelog page in docs.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,1 +1,7 @@
 .. include:: ../CHANGELOG.rst
+   :end-before: Changes for the upcoming release can be found
+
+.. towncrier-draft-entries:: |release|:sub:`/UNRELEASED DRAFT/`
+
+.. include:: ../CHANGELOG.rst
+   :start-after: .. towncrier release notes start
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,7 +1,7 @@
 .. include:: ../CHANGELOG.rst
    :end-before: Changes for the upcoming release can be found
 
-.. towncrier-draft-entries:: |release|:sub:`/UNRELEASED DRAFT/`
+.. towncrier-draft-entries:: Not yet released
 
 .. include:: ../CHANGELOG.rst
    :start-after: .. towncrier release notes start
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,12 @@
 # SPDX-License-Identifier: MIT
 
 from importlib import metadata
+from pathlib import Path
+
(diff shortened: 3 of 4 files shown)


changelog.d/1016.doc.rst
```diff
@@ -0,0 +1 @@
+Render the unreleased ``attrs`` version change notes on the changelog page in docs.
```
hynek, line 1: We don’t have a doc category, so this will have to do without a fragment.

docs/changelog.rst
```diff
@@ -1 +1,7 @@
 .. include:: ../CHANGELOG.rst
+   :end-before: Changes for the upcoming release can be found
+
+.. towncrier-draft-entries:: |release|:sub:`/UNRELEASED DRAFT/`
```
hynek, line 4: What exactly is the `sub` supposed to do? I would have guessed substitute the release but it looks like it uses a `<sub>` tag or something? Is that on purpose? If yes: why? :)

hynek (changes requested):
Looks cool, but please address my comments. :)

changelog.d/1016.doc.rst
webknjaz: Oh, I didn't realize that. That's a power of habit :)

Did you consider including `CONTRIBUTING.md` into the Sphinx docs? It's pretty straightforward to add a single Markdown document there via MyST. No need to convert everything into the same format.

P.S. Feel free to try out https://github.com/apps/psf-chronographer if you want to verify the presense of the allowed change note types in PRs, like aiohttp does.

docs/changelog.rst
webknjaz: Would you prefer something like this?
```suggestion
.. towncrier-draft-entries:: |release| | *UNRELEASED DRAFT*
```
Any other thoughts?

hynek: maybe even remove `|release|` too? pip hasn't it either

hynek: fee free to use their `Not yet released`

changelog.d/1016.doc.rst
webknjaz: Personally, I was a little lost when I couldn't locate this in the docs, among the top-level items. I think there was a link on one of the pages where I never land coming from Google.

docs/changelog.rst
webknjaz: Yeah, maybe even `Will be included in the upcoming release`?

hynek: naw too long. what's good enough for pip, is good enough for us (in this case)

webknjaz: I've gone for what pip has.

changelog.d/1016.doc.rst
hynek: there's one on the github readme, the main docs page, and the pypi page :D

maybe i'll just add it to the list so it's less embedded

hynek (approved):
Sweet, thanks!

PR #1018, opened 2022-08-22
Fix pre-commit link syntax in contributing doc

webknjaz:
# Summary

It was using RST but the doc is actually in Markdown.


hynek (approved):
thanks!

PR #1029, opened 2022-09-23
Introduce a gate/check GHA job

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It also allows to reduce the list of required branch protection CI statuses to just one — `check`. This reduces the maintenance burden by a lot and have been battle-tested across a small bunch of projects in its action form and in-house implementations of other people.

I was curious about the spread of use. And I've just discovered that it is now in use in aiohttp (and other aio-libs projects), CherryPy, some of the Ansible repositories, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some PyCQA, PyCA and pytest projects, a few AWS Labs projects (to my surprise). Admittedly, I maintain a few of these but it seems to address some of the pain folks have: https://github.com/jaraco/skeleton/pull/55#issuecomment-1106638475.

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,3 +107,21 @@ jobs:
           python-version: ${{env.PYTHON_LATEST}}
       - run: python -m pip install -e .[dev]
       - run: python -c 'import attr; print(attr.__version__)'
+
+  check:  # This job does nothing and is only used for the branch protection
+    if: always()
+
+    needs:
+      - coverage
+      - install-dev
+      - package
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}
+
+...


.github/workflows/ci.yml
```diff
@@ -107,3 +107,21 @@ jobs:
...
+    needs:
+      - coverage
+      - install-dev
+      - package
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
```
webknjaz, line 123: In case, you'll want to pin it:
```suggestion
        uses: re-actors/alls-green@198badcb65a1a44528f27d5da555c4be9f12eac6  # v1.2.0
```

hynek (approved):
I guess it's nice to not have to click myself through menus to define the necessary checks!

sobolevn:
`master` is now red: https://github.com/python-attrs/attrs/actions/runs/3128734322/jobs/5076983674

webknjaz:
> 📝 coverage → ⬜ skipped [required to succeed]

If it is okay to be skipped, it should be in `allow-skips`. This param can be templated based on the event context.

hynek:
Well, pls fix? 😇

webknjaz:
> Well, pls fix? :innocent:

I just looked into it closer and discovered that in previous commits, the `coverage` job runs. Here's the previous one: https://github.com/python-attrs/attrs/actions/runs/3080585434/jobs/4978137502.

In the last commit, it was skipped: https://github.com/python-attrs/attrs/actions/runs/3128734322/jobs/5076983456. This is because building docs in this job https://github.com/python-attrs/attrs/actions/runs/3128734322/jobs/5076948250 failed. And it most likely failed because of flaky network (there's no visual proof of this but I assume that Sphinx may have failed to fetch bits of the intersphinx objects file).

That said, the status for the branch would show up red _even without this patch_. It just added an explicit additional red cross icon but that's about it.

@hynek I think if you restart that failing py310 job (you can now restart a single job in GHA), the red status would clear out and become green.

Back to changing the checks settings. Do you actually want skipped/unmeasured coverage on pushes to `main`? If so, I can make a PR. But I was under an impression that you would prefer the jobs to be green.

Also, you may want to add `if: always()` to the coverage job so that it'd publish some artifact even in the case of failure. Doing so won't affect the overall workflow status unlike adding `continue-on-failure: true`.

So which way do you want to go? I'll make a PR if you still want it.

webknjaz:
> I assume that Sphinx may have failed to fetch bits of the intersphinx objects file

By the way, a “cure” to this would be pre-caching the intersphinx files in Git — this way Sphinx would be able to link external references even when running offline.

hynek:
On phone so seeing only like 50%, but since restart failed too: could it be that some update of something broke it?

webknjaz:
> could it be that some update of something broke it?

Looks like it: https://github.com/python-attrs/attrs/pull/1030#issuecomment-1259611491.

webknjaz:
Filed https://github.com/python-attrs/attrs/issues/1031 for more transparent tracking.

hynek:
OK finally back at the keyboard and I think this needs more love and I want to run it by you first.

For me the big upside of this is to not have to click around in the UI. This is interesting when e.g. 3.12beta1 comes out and I want to allow it to break.

What would be the best way to achieve that? Just keep it out of coverage?

webknjaz:
> What would be the best way to achieve that? Just keep it out of coverage?

Depending on what you want, you could have `if: always()` on the coverage job. If the failing Python 3.12 job still produces coverage, this might be useful. Or you could have a `continue-on-error: ${{ contains(matrix.python-version, 'beta') && true || false }}` in the `tests` matrix definition. This is something I've been doing in Cheroot: https://github.com/cherrypy/cheroot/blob/e0dd9cb/.github/workflows/ci-cd.yml#L524-L537 (except it now uses a leading tilda syntax to achieve the same pre-release or the first final). The latter gives you more granular control. You could also add an `experimental` factor into the matrix and maintain an explicit list of things that are okay to fail. When a job has `continue-on-error`, GitHub shows a red status (just visually) for that job if it fails, but it sets `success` for the whole job matrix if other jobs succeed.
`continue-on-error` can also be applied to an individual step, in which case the job status won't be red if that step fails, but the step itself will be marked as red in logs. Other implications are about the same.

So to answer the second question — ­no, you don't need to maintain a separate job definition just for that case. I'd advise against it because of extra unnecessary burden bundled with such a solution.

Note that GitHub's `${{ needs }}` context does not contain details of individual jobs if they are matrix definitions — just the overall outcome. This is why it's not implemented as an action input.

PR #1030, opened 2022-09-27
[DNM] Allow skipping `coverage` job on pushes `main`

webknjaz:
# Summary

$sbj. Asked in https://github.com/python-attrs/attrs/pull/1029#issuecomment-1259488083.

But @hynek, I'm pretty sure this _is not what you want_. Which is why I wrote an explanation here: https://github.com/python-attrs/attrs/pull/1029#issuecomment-1259525838. Keeping this as a draft to prevent an accidental merge. Read those comments first, before proceeding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,18 @@ jobs:
       - name: Decide whether the needed jobs succeeded or failed
         uses: re-actors/alls-green@release/v1
         with:
+          allowed-skips: >-
+            ${{
+              (
+                github.event_name == 'push' &&
+                github.ref == format(
+                  'refs/heads/{0}',
+                  github.event.repository.default_branch
+                )
+              )
+              && 'coverage'
+              || ''
+            }}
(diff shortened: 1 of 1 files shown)


hynek:
No it's not...I'm on a phone so I checked only on the second fail and it looks like docs is failing because a symbol can't be found?

webknjaz:
> No it's not...I'm on a phone so I checked only on the second fail and it looks like docs is failing because a symbol can't be found?

Yeah, so I suspect an unpinned Sphinx version or an update in CPython.

webknjaz:
Yep, the last success had `Sphinx==5.1.1` in the env: https://github.com/python-attrs/attrs/actions/runs/3080585434/jobs/4978108196#step:5:90. And the following failure has `Sphinx==5.2.1`: https://github.com/python-attrs/attrs/actions/runs/3128734322/jobs/5092896008#step:5:90.
That must be it.

webknjaz:
Filed https://github.com/python-attrs/attrs/issues/1031 for more transparent tracking.

PR #1032, opened 2022-09-28
Use short syntax to request pre+final py311

webknjaz:
Ref https://github.com/python-attrs/attrs/pull/1029#issuecomment-1261394538


hynek (approved):
TIL!

webknjaz:
> TIL!

Yeah, it's poorly documented (as in not documented at all) but recommended on the tracker. Plus, there's a bunch of debates around `3.11-dev` syntax where folks are trying to convince GH to make just `3.11` work the same way.

PR #1033, opened 2022-09-28
Add unstable Python 3.12 to CI allowing failures

webknjaz:
Ref https://github.com/python-attrs/attrs/pull/1029#issuecomment-1261394538.

This is kept as a draft that is expected to be merged once all of the following are true:
* [x] Python 3.11 final is out
* [x] GitHub updated their available CPython builds to include Python 3.12-dev
* [x] The DNM (do-not-merge) tag is dropped from the PR title

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,12 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "~3.12.0-0"
           - "pypy-3.7"
           - "pypy-3.8"
+
+    continue-on-error: >-
+      ${{ contains(matrix.python-version, '~') && true || false }}
 
     steps:
       - name: Harden Runner
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -13,6 +13,7 @@ from collections.abc import Mapping, Sequence  # noqa
 
 PYPY = platform.python_implementation() == "PyPy"
 PY310 = sys.version_info[:2] >= (3, 10)
+PY_3_12_PLUS = sys.version_info[:2] >= (3, 12)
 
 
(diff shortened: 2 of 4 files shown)


webknjaz:
> * [ ]  Python 3.11 final is out
> * [ ] GitHub updated their available CPython builds to include Python 3.12-dev

So the new job is failing in 6 seconds, which means that it's also okay(ish) to merge this now — it won't waste too much of the CI resources. But there would be a downside of having a red mark in some places on the UI, which is going to be annoying.

hynek:
Yeah not merging until it's passing.

Also lol you've introduced yourself a conflict. ;)

webknjaz:
> Also lol you've introduced yourself a conflict. ;)

Yep, I was expecting that. That's a side effect of having a habit to keep commits and PRs atomic :man_shrugging:

.github/workflows/ci.yml
```diff
@@ -34,9 +34,13 @@ jobs:
           - "3.9"
           - "3.10"
           - "~3.11.0-0"
```
webknjaz, line 36: @hynek note that with this, 3.11 is also allowed to fail. The conditional may need to be improved if you don't want that.

webknjaz:
@hynek looks like 3.12 alpha is available now

webknjaz:
I suppose coverage might've dropped independently.

webknjaz:
> I'm guessing coverage.py isn't there yet

Looks like you're right.

webknjaz:
@hynek it occurred to me that maybe the jobs that are allowed to fail shouldn't influence the coverage check? WDYT?

webknjaz:
@hynek it's green now!

hynek (approved):

PR #1045, opened 2022-10-26
Use the stable version of Python 3.11 in CI

l1storez:
This patch updates the use of Python 3.11 in CI to the newly released stable version, in the test jobs. It also moves the non-test jobs from Python 3.10 to 3.11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ env:
   TOX_TESTENV_PASSENV: FORCE_COLOR
   PIP_DISABLE_PIP_VERSION_CHECK: "1"
   PIP_NO_PYTHON_VERSION_WARNING: "1"
-  PYTHON_LATEST: "3.10"
+  PYTHON_LATEST: "3.11"
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


webknjaz (approved):
@hynek mind allowing the CI to run?

l1storez:
@webknjaz please help

webknjaz:
@l1storez `step-security/harden-runner` action blocks the ability of `actions/setup-python` to download Python 3.11. You'll need to adjust the `allowed-endpoints` value for the failing jobs to allow this, like this is done in the passing jobs.

l1storez:
okay, I will fix it

varunsh-coder:
@l1storez please let me know if you need any help with adjusting the allowed endpoints. Based on the insights [here](https://github.com/python-attrs/attrs/actions/runs/3350788753/jobs/5551865071#step:2:19), looks like calls to `api.github.com:443` are being blocked in the jobs that are failing.

l1storez:
add `api.github.com:443` to jobs, need a test run of workflow to check if changing was right

webknjaz:
@hynek allow the CI plz

l1storez:
smth goes wrong. Checks passed. I'll squash the commits locally and push

hynek (approved):
Thanks!

PR #1067, opened 2022-12-02
Switch changelog to markdown

hynek:
Another step towards rST's well-deserved retirement.

Unfortunately it costs us the towncrier preview until it adds support for Markdown (cc @webknjaz), but progress is always painful. 🤪


docs/conf.py
```diff
@@ -42,7 +42,8 @@
     "sphinx.ext.intersphinx",
     "sphinx.ext.todo",
     "notfound.extension",
-    "sphinxcontrib.towncrier",
+    # Disabled until it speaks Markdown:
```
webknjaz, line 45: @hynek I'm pretty sure it should still work. What problem are you experiencing?

hynek: It included the Markdown rendered as rST including complains about syntax in the console. 

I tried setting a parser like with include but that didn't work. 

I suspect this might not be too hard to fix?

webknjaz: I've seen reports of MyST users using the extension + tried it out myself at some point, and it worked. It only invokes `towncrier --draft` and then feeds the output to the docutils/Sphinx source parser, without specifying any markup — my understanding is that it can be parsed with any parsers that are configured in Sphinx.

I'll try to look into your PR deeper to see what's happening.

docs/changelog.md
```diff
@@ -0,0 +1,4 @@
+```{eval-rst}
```
webknjaz, line 1: Why are you using RST here? Just use a MyST-native include-directive. Maybe this is the reason the extension didn't work for you...

https://github.com/executablebooks/MyST-Parser/pull/469/files#diff-f58f9fe275eb1074a0b4a24a31f69d6fd72afb6a699a512b0bae9e0e1a97f5d5R74-R75

docs/conf.py
webknjaz: So... I haven't checked out this branch locally, but I'm pretty sure it's https://github.com/python-attrs/attrs/pull/1067#pullrequestreview-1202439985.

docs/changelog.md
hynek: nice you should totally stop talking only about rst in the docs – i read it as a given

hynek: I guess with the current design of Towncrier, it's impossible to fix the link of Unreleased, huh?

webknjaz: > nice you should totally stop talking only about rst in the docs – i read it as a given

Yeah, somebody recommended me to stop using rst in the function name years ago and I got to doing that just recently: https://github.com/sphinx-contrib/sphinxcontrib-towncrier/commit/ed76868cfff43091297a9c726815fadb85f3ef2f.

webknjaz: One way could be allowing to specify a different template, plus it should be possible to specify a different config file with the recent versions already. It's a bit ugly to maintain, but might work. Another way could be allowing an arbitrary hook for post-processing whatever `towncrier --draft` spits out.

docs/changelog.md
```diff
@@ -0,0 +1,11 @@
+```{include} ../CHANGELOG.md
+:end-before: Changes for the upcoming release can be found
+```
+
```
webknjaz, line 4: Try this maybe?

````suggestion

```{important} The version marked as __UNRELEASED__ is not yet released and is under active development.
```

````

File: pyproject.toml [hunk 4]
```diff
@@ -50,12 +50,12 @@ toplevel = ["attr", "attrs"]
...
 name = "attrs"
-filename = "CHANGELOG.rst"
-template = "changelog.d/towncrier_template.rst"
-issue_format = "`#{issue} <https://github.com/python-attrs/attrs/issues/{issue}>`_"
 directory = "changelog.d"
-title_format = "{version} ({project_date})"
-underlines = ["-", "^"]
+filename = "CHANGELOG.md"
+template = "changelog.d/towncrier_template.md.jinja"
+title_format = "## [{version}](https://github.com/python-attrs/attrs/tree/{version}) - {project_date}"
```
webknjaz, line 56:
```suggestion
title_format = "## {version} - {project_date}"
```

changelog.d/towncrier_template.md.jinja
```diff
@@ -1,14 +1,7 @@
 {% for section, _ in sections.items() %}
```
webknjaz, line 1:
```suggestion
{%- if versiondata.version != 'Unreleased' -%}
[See the source on GitHub](https://github.com/python-attrs/attrs/tree/{versiondata.version})
{%- endif -%}

{% for section, _ in sections.items() %}
```

webknjaz (commented):
It occurred to me that you could try to improve the situation with the `Unreleased` title link pointing to nowhere by moving the link into the template and keeping the title as plain text.

Another “solution” would be to create a branch literally called `unreleased`. It could even be orphan, only having a README telling everyone that this is now what they are looking for, with a link to the repo main page...

docs/changelog.md
hynek: I've moved the h1 into the template – think I've [nailed it](https://attrs--1067.org.readthedocs.build/en/1067/changelog.html) now. 🎉

PR #1071, opened 2022-12-18
Automate PyPI releases

hynek:
I want to outsource some of the publishing machinery (especially to lower the risk of packaging some local garbage).

This flow has the following ideas:

- it moves package issues into a separate workflow
- on every push it still builds and verifies the package using my [*build-and-inspect-python-package*](https://github.com/hynek/build-and-inspect-python-package) action for continuous packaging monitoring.
- on pushed tags, it pushes the package test PyPI which allows last-second checks
- on GitHub Release, it finally pushes the package to the real PyPI (graciously stolen from @hugovk)

Unfortunately, I'm not sure how to test any of this, except maybe tagging an rc or something?

### DIFF
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -54,6 +54,5 @@ jobs:
         if: github.event.action == 'published' && github.repository_owner == 'python-attrs'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
           packages_dir: ${{ steps.baipp.outputs.dist }}
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -44,7 +44,6 @@ jobs:
         if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && github.repository_owner == 'python-attrs'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: __token__
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
           repository_url: https://test.pypi.org/legacy/
           packages_dir: ${{ steps.baipp.outputs.dist }}
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -1,5 +1,5 @@
 ---
-name: Build and maybe publish to PyPI
+name: Build & push package to (Test-) PyPI
 
 on:
   push:
@@ -12,14 +12,13 @@ on:
       - published
   workflow_dispatch:
 
-
 permissions:
   contents: read
 
-
 jobs:
-  deploy:
+  deploy-test-pypi:
     runs-on: ubuntu-latest
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && github.repository_owner == 'python-attrs'
 
     steps:
       - name: Harden Runner
@@ -29,7 +28,6 @@ jobs:
           allowed-endpoints: >
             files.pythonhosted.org:443
             github.com:443
-            pypi.org:443
             test.pypi.org:443
 
       - uses: actions/checkout@v3
@@ -41,16 +39,36 @@ jobs:
 
       # Upload to Test PyPI on every pushed tag.
       - name: Publish package to TestPyPI
-        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && github.repository_owner == 'python-attrs'
+
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
           repository_url: https://test.pypi.org/legacy/
           packages_dir: ${{ steps.baipp.outputs.dist }}
 
+  deploy-pypi:
+    runs-on: ubuntu-latest
+    if: github.event.action == 'published' && github.repository_owner == 'python-attrs'
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@v2
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            files.pythonhosted.org:443
+            github.com:443
+            pypi.org:443
+
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: hynek/build-and-inspect-python-package@v1
+        id: baipp
+
       # Upload to real PyPI on GitHub Releases.
       - name: Publish package to PyPI
-        if: github.event.action == 'published' && github.repository_owner == 'python-attrs'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.PYPI_API_TOKEN }}
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -1,0 +1,59 @@
+---
+name: Build and maybe publish to PyPI
+
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
+  pull_request:
+    branches: [main]
+  release:
+    types:
+      - published
+  workflow_dispatch:
+
+
+permissions:
+  contents: read
+
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@v2
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            files.pythonhosted.org:443
+            github.com:443
+            pypi.org:443
+            test.pypi.org:443
+
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: hynek/build-and-inspect-python-package@v1
+        id: baipp
+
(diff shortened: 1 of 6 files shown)


hugovk:
Hooray for more automation!

> Unfortunately, I'm not sure how to test any of this, except maybe tagging an rc or something?

It can take a bit of trial and iteration to get working first time.

But we have the TestPyPI bit, you could temporarily test that out on this `auto-publish` branch first to increase confidence.

And an RC sounds good after that (or make a full release, it's been a while since July ;)

hynek:
I would love to publish (this is the last step basically); but would love to verify somehow that it works first. 😅

webknjaz:
> Unfortunately, I'm not sure how to test any of this, except maybe tagging an rc or something?

In @aio-libs, the testing is often done by making a `.a0` release or `.dev0`. These seem more appropriate than release candidates. They are also pre-releases from the PoV of PEP 440, pip and PyPI.

Although, I prefer a different school of thought personally which infolves a more tightly integrated workflow triggered by a `workflow_dispatch` event (essentially, it's an input from on GH), which being the initial trigger, allows to let the automation produce all the release-related pieces (Towncrier commit, tag, GH release with a linked discussion etc.)
With that, I aim to have job(s) for building dists at the first stage, followed by the testing that actually imports whatever's shipped in those dists rather than a Git checkout, and finally an environment-protected job for publishing the same artifacts that were tested rather than rebuilding them with possible differences.

Another approach I use is continuously publishing to PyPI on merges to the default branch. The versions produced by setuptools-scm help me with that.

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,59 @@
...
+      - published
+  workflow_dispatch:
+
+
+permissions:
+  contents: read
+
+
+jobs:
+  deploy:
```
webknjaz, line 21: For extra security, use `environment: pypi` and `environment: testpypi` on the separate release jobs. You can then add protection on the `pypi` env in the repo settings to require a manual approval and a cool-down timer while `testpypi` wouldn't have those. With the env separation you can even use the same secret name for those envs. I ended up defaulting to `PYPI_API_TOKEN`.

webknjaz:
One extra testing possibility is to use devpi — it got recently integrated into the pypi-publish action smoke test. But not sure what you'd test with that — certainly not whether the secrets are correct.

webknjaz (commented):
Here's some things that I tend to do in my automations.

webknjaz:
> **required-checks-pass** *Expected — Waiting for status to be reported* `Required`

You may need to have a job with the same name in this workflow. This is one of the things that you don't have to deal with when you have a unified workflow. Though, if there are cases when both workflows are triggered for the PR, they can create confusion for automerge — if one reports success, GH will count that as a fulfilled branch protection condition even though the other workflow may still be running. This is because these things are dynamic and the GitHub platform does not know what will report the ~status~check with the same name. It's easier with GitHub Apps since you can now scope the expected check to specific bots but if there are several workflows reporting statuses via Checks API, they are all coming from the same GH App so there's no way to declare such expectations on the repo settings level.

webknjaz:
> * it moves package issues into a separate workflow

As a follow up to the last comment, I'd like to point out that reusable workflows are a thing and you can now refer to the workflows in the same repo. So this may help you better structure the job definitions in multiple files while wiring them to the events/triggers in the main one.
Another solution would be integrating some sort of polling into the job that produces the final status for branch protection. I've seen an action facilitating this somewhere. But this does not seem elegant.

hynek:
> > * it moves package issues into a separate workflow
> 
> As a follow up to the last comment, I'd like to point out that reusable workflows are a thing and you can now refer to the workflows in the same repo. So this may help you better structure the job definitions in multiple files while wiring them to the events/triggers in the main one. Another solution would be integrating some sort of polling into the job that produces the final status for branch protection. I've seen an action facilitating this somewhere. But this does not seem elegant.

I'm sorry but I don't understand almost any of this. :D

I've decided to go a slightly less DRY, but more explicit and clear way of splitting it up and protecting the whole jobs. Does this make sense?

re: workflow_dispatch: my impression is, that using the released event, I'm getting basically the same thing?

.github/workflows/pypi-package.yml
hynek: I have done this now – can it still work with the releases trigger? where does the review into play?

hynek:
I guess it's easiest to merge and play with the test env

webknjaz:
> I've decided to go a slightly less DRY, but more explicit and clear way of splitting it up and protecting the whole jobs. Does this make sense?

When you don't have an artifact upload job, you can't download the artifacts for inspection. I suppose it may work for you at the cost of building the dists twice. You can then inspect the dists by downloading them from (Test)PyPI.

> re: workflow_dispatch: my impression is, that using the released event, I'm getting basically the same thing?

That depends on the PoV. It's definitely not 100% the same, but people do use both of these as triggers. The number of things and ways you can automate stuff varies for these two.

The released event happens when you make a GitHub Release. GitHub Release requires a Git tag. So by the time the workflow is triggered, you've created something in Git and published it to GitHub. The workflow may produce broken dists or fail, but the GitHub Release and the tag already exist, and some external automations may start advertising the fact that you've made a release while there's nothing yet on PyPI, and there won't be. You may end up being forced to remove the tag and tag a different commit with the same thing. Another implication of this is having a human-driven manual step for generating a changelog and committing that. This is where things may become fragile — the human factor may play a bad joke.

With the workflow_dispatch, you can define your own inputs and get a form in the UI where you can enter the desired version. You fill that out, click a button and a workflow is triggered. At this point, there's nothing mutated on GitHub or in the Git repo. The artifacts may get tested, changelog may be generated and this will be a part of the GHA workflow run up until the point when there's some verification that all the bits fit together. I like publishing to PyPI and only then pushing a commit+tag for that release — at a point when the automation *knows* the release was successful.
The TestPyPI job runs in parallel and is conditional on the build/test, but is pretty much automatic. So at some point in the workflow, there's something released to TestPyPI and the PyPI job is frozen (by the env approval requirement). This is when you can inspect what got published to TestPyPI (but not to GitHub, no commit, no tag, no release) and click approve on the main release job which could be followed by committing the final stuff back to GitHub.

So this is the difference for me. But for some people it'd be less important, and they won't mind having more manual steps.

.github/workflows/pypi-package.yml
webknjaz: There's nothing special about the relation of triggers and environments assigned to jobs. The environment will be used whenever it's set, regardless of the trigger. The secret access, though, might still be dependent on the event in this case, I'm not sure.

hynek:
Right, I already have plenty local automation so for me this is a meaningful step anyway.

What I'll be doing is leaving out the upload locally and have a last chance to check the readme on Test PyPI etc before starting the release and upload to PyPI proper. I agree tho that it probably gets hairy if I notice a problem, since I can't just delete the version from test PyPI (which is a bummer).

I might go the constantly-uploading route once I'm on hatch-vcs, but it seems like an enormous waste of resources to store every main commit. :|

hynek:
Well, it didn't work because I didn't know that there's a separate hostname for uploads. sigh.

f04dce2

webknjaz:
> I agree tho that it probably gets hairy if I notice a problem, since I can't just delete the version from test PyPI (which is a bummer).

FWIW there's an effort to implement draft uploads in PyPI that seems to be slowly moving towards resolution — it'll allow deleting/reuploading drafts w/o having to burn through versions.

Another ongoing effort is the OIDC integration that allows throwing away the tokens saved as repo secrets and would let us make use of short-lived tokens that would be retrieved automatically in GHA.

> I might go the constantly-uploading route once I'm on hatch-vcs, but it seems like an enormous waste of PSF resources to store every main commit. :|

FWIW C-extension projects that have a ton of wheels hit the project limit in a year or two depending on the size, but it's not such a big problem for pure-python projects.

hynek:
> When you don't have an artifact upload job, you can't download the artifacts for inspection.

JFTR, this seems to be a thing among actions tho?

Because baipp puts the dist to /tmp/baipp and when I `run` an ls there, they are there: https://github.com/python-attrs/attrs/actions/runs/3732374577/jobs/6331733243#step:7:6

But for the action it's not: https://github.com/python-attrs/attrs/actions/runs/3732374577/jobs/6331733243#step:8:15

webknjaz:
> But for the action it's not

@hynek It is indeed unobvious… So I stared at the log for a good minute and then realized what was happening.

TL;DR It turned out to be unrelated to the use of artifacts, but rather the unpacking location.

GitHub allows making 3 types of actions: js-based, composite and docker-based. The two first ones are executed in the job VM runtime directly. But the docker actions are containerized, obviously. They have certain limitations, like being restricted to Linux-based `runs-on`. There are a number of directories that are mounted into the container when the action gets invoked (and there are also bypassed env vars, which created problems for us recently…)
Basically, the workdir, some temporary dirs and GHA framework/helper dirs are mounted but not `/tmp`. You can see the exact `docker run` command here: https://github.com/python-attrs/attrs/actions/runs/3732374577/jobs/6331733243#step:8:12.
The checkout directory also becomes an in-container workdir. This is why, setting a relative path for this action input works — it's in a mounted location.

PR #1072, opened 2022-12-20
Fix minor stub issues

hynek:
I've tried running ruff over attrs and these issues seem legit.

### DIFF
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -48,17 +48,17 @@ _T = TypeVar("_T")
 _C = TypeVar("_C", bound=type)
 
 _EqOrderType = Union[bool, Callable[[Any], Any]]
-_ValidatorType = Callable[[Any, Attribute[_T], _T], Any]
+_ValidatorType = Callable[[Any, "Attribute[_T]", _T], Any]
 _ConverterType = Callable[[Any], Any]
-_FilterType = Callable[[Attribute[_T], _T], bool]
+_FilterType = Callable[["Attribute[_T]", _T], bool]
 _ReprType = Callable[[Any], str]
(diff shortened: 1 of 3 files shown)


hynek:
uh what's with the checks? did i screw something up when adding environments to the deploy stuff? @webknjaz? 😇

https://www.githubstatus.com says all is dandy

Tinche (approved):

webknjaz:
> uh what's with the checks? did i screw something up when adding environments to the deploy stuff? @webknjaz? innocent
> 
> [githubstatus.com](https://www.githubstatus.com) says all is dandy

I haven't seen what was happening when you tagged me, but sometimes GitHub is just drunk. Their eventually consistent event delivery system sometimes fails to propagate events and doesn't re-deliver them after having platform troubles.
It's also worth noting that GitHub doesn't always update the status website. Sometimes there are issues with GHA or unicorns that are short-lasting and that doesn't end up on the status page.
Looking at https://www.githubstatus.com/history, I see that today there were two reported problems with Actions specifically a bit earlier than the PR's timestamp, it seems.

The environments wouldn't affect this because they are applied to jobs in another workflow, and you'd have notifications about the approval requests in your inbox. Besides, workflow definition parsing issues manifest themselves differently — there's a weird error message on the workflow run page with almost zero context on what's broken.

hynek:
Well, it didn't work because I didn't know that there's a separate hostname for uploads. _sigh_.

https://github.com/python-attrs/attrs/commit/f04dce20f1aaa89a6fa888e7a07e133f434ac602

hynek:
They just wouldn't start – it looked like my problems in my package builder where CI wouldn't run on PRs.

PR #1079, opened 2023-01-03
First pass over generics

Tinche:
Support generic classes for `has` and `fields`.


Tinche:
@hynek *chants* dump 3.6, dump 3.6

hynek:
_chants_ coverage, coverage

hynek:
Do you plan on finishing this up or can I close it?

Tinche:
This can be merged, it can just be worked on further, but the further work gets a lot more complex. This first step has value on its own.

hynek:
This cannot be merged because it's [breaking CI](https://github.com/python-attrs/attrs/pull/1079#issuecomment-1379043851).

Tinche:
Ah, oops. Will work on it some more.

Tinche:
Ok, there's something funky with coverage.

Locally, I'm getting this:

Screenshot 2023-02-03 at 16 36 25

But this is obviously wrong, since if I put an Exception below the `if`, one test blows up. So the lines are being hit for sure.
Any ideas?

hynek:
Does it maybe only get hit in a version that doesn't run under coverage?

Tinche:
This was me running coverage _locally_, on Python 3.11. Going to try working on this some more.

Tinche:
Ok, I got it.

Running coverage with this configuration is a little trickier than I thought. I had to run:

`❯ rm .coverage*; coverage run -m pytest; and coverage combine;and coverage html;and open htmlcov/index.html`

But then I get accurage coverage reports (modulo other Python versions, which are not important here).

Anyway, CI is green now.

hynek (approved):

PR #1080, opened 2023-01-04
Fix object index in API docs by splitting API docs in two

### LINKED ISSUE #1077: Some API doc entries in objects.inv have double-attr module prefixes
e.g. `attr.attr.cmp_using`

cursory looking suggests it's autodoc vs manual.

hynek:
So this has turned into a total nightmare thanks to the different behavior of defining functions (and everything else) using `.. autofunction` vs `.. function` and the interaction with `.. module xyz` (can exist only once, but implies currentmodule).

I think instead of merging this crap, I'll split api.rst into `api.rst` and `api-attr.rst`. That should help overview too.

closes #1077


docs/api-attr.rst
```diff
@@ -0,0 +1,242 @@
+API Reference for the ``attr`` Namespace
+========================================
+
+.. module:: attr
+
+
+Core
+----
+
+.. autofunction:: attr.s(these=None, repr_ns=None, repr=None, cmp=None, hash=None, init=None, slots=False, frozen=False, weakref_slot=True, str=False, auto_attribs=False, kw_only=False, cache_hash=False, auto_exc=False, eq=None, order=None, auto_detect=False, collect_by_mro=False, getstate_setstate=None, on_setattr=None, field_transformer=None, match_args=True, unsafe_hash=None)
```
webknjaz, line 10: Does this actually work? I double-checked the docs and it looks like `.. function::` accepts a signature to render but `.. autofunction::` expects just the name.
```suggestion
.. function:: attr.s(these=None, repr_ns=None, repr=None, cmp=None, hash=None, init=None, slots=False, frozen=False, weakref_slot=True, str=False, auto_attribs=False, kw_only=False, cache_hash=False, auto_exc=False, eq=None, order=None, auto_detect=False, collect_by_mro=False, getstate_setstate=None, on_setattr=None, field_transformer=None, match_args=True, unsafe_hash=None)
```

webknjaz: Aha.. So it looks like it works https://attrs--1080.org.readthedocs.build/en/1080/api-attr.html#attr.s, just undocumented in Sphinx.

docs/api.rst
```diff
@@ -1,115 +1,90 @@
 API Reference
```
webknjaz, line 1: @hynek have you considered generating this document via https://pypi.org/project/sphinxcontrib-apidoc/ ?

docs/api.rst
```diff
@@ -1,115 +1,90 @@
...
-  Most of the objects are simply re-imported from ``attr``.
-  Therefore if a class, method, or function claims that it has been added in an older version, it is only available in the ``attr`` namespace.
-
 .. autodata:: attrs.NOTHING
    :no-value:
 
 .. autofunction:: attrs.define
 
-.. function:: attrs.mutable(same_as_define)
+.. function:: mutable(same_as_define)
```
webknjaz, line 28: So I was trying to google something I remembered was discussed on the Sphinx tracker, but couldn't recall the exact feature name. I think I've found it now and it's called `:canonical:`. @hynek I think this might be something you're after.
```suggestion
.. function:: mutable(same_as_define)
   :canonical: attrs.define
```

Check out the docs at https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option-py-function-canonical. FTR I think it's documented quote poorly. I think it was discussed @ https://github.com/sphinx-doc/sphinx/issues/4826 / https://github.com/sphinx-doc/sphinx/pull/9026.

hynek:
> Did you inspect the result with `python -m sphinx.ext.intersphinx`?

I've used sphobjinv and doc2dash – didn't know about sphinx.ext.intersphinx – handy!

---

`attr._make.Factory` and `attr._make.Attribute` aside (which I _think_ stems from the aliasing / adding of methods), that list looks good?

docs/api.rst
hynek: Why should I?

docs/api-attr.rst
hynek: Yeah that always worked. Otherwise you get very ugly signatures for decorators that accept arguments.

docs/api.rst
hynek: That doesn't seem to do anything:

CleanShot 2023-01-05 at 07 44 53@2x

and

```
attrs.mutable                            api.html#attrs.mutable
```

hynek:
@webknjaz does this need anything else or can we merge it?

docs/api.rst
webknjaz: Hm… Maybe I've misinterpreted how it works. Let's postpone using it, then.

docs/api.rst
webknjaz: I tend to do this in an attempt to stop maintaining an API document and just render it templated from docstrings. Specifically for "private api" — it makes many internal objects linkable too. Might be useful here with a custom template, maybe. FWIW, this could be a separate effort, I was just posting ideas to see if it'd interest you.

hynek: hm ok I'm asking because the description is rather inconclusive as to what it gives me and it looks rather unmaintained 🤔

guess this is big enough, I'm open for separate input tho

webknjaz:
> `attr._make.Factory` and `attr._make.Attribute` aside (which I _think_ stems from the aliasing / adding of methods), that list looks good?

I'm wondering if that `:canonical:` feature could help get rid of the unwanted aliases... Anyway, there are no blockers from my side that would justify delaying the merge — I just missed notifications about the previous replies.. Improvements can always be made later.

docs/api.rst
webknjaz: @hynek that project doesn't really need to be maintained actively. Basically, the Sphinx upstream has a `sphinx-apidoc` helper that generates RST documents out of the source code but needs to be run manually — you'd have to rerun it on any changes and commit those documents. `sphinxcontrib-apidoc`, OTOH, wraps that and runs this command automatically on every build allowing you to gitignore the directory with autogenerated docs and keep them in sync with the Python modules. And since the configuration mostly defines what is passed to `sphinx-apidoc` in those automatic invocations, this means that the extension does not need to be updated as often as the Sphinx upstream.

hynek:
_sigh_ GH doesn't start checks

webknjaz:
> _sigh_ GH doesn't start checks

I'm observing that Azure network seems flaky today. This would explain why some events don't get delivered to the orchestrator invoking runners there... GH doesn't do retries on that front.

docs/api.rst
hynek: hmmm I don't think I want such a template-based approach – if I understand it correctly.

I'm trying to editorize my API docs to have a flow of concepts and like to add examples, such that they don't have to be in docstrings which is a nightmare to do :)

webknjaz: Ah, I usually like these things in docstrings. I'm curious if it's possible to combine these approaches — a top-module docstring is often useful for this.

FWIW, the flow thingy is typically separate from the reference per https://diataxis.fr/. Maybe that would be fun to discuss in the Write The Docs Slack instance :)

hynek: I could not disagree more with this dichotomy. I'm a big fan of https://diataxis.fr/ but that doesn't mean that reference documentation should appear in random order. APIs usually build on each other and some are more important / fundamental than others. I absolutely tend to scroll through them to see what a package can do because no package under the sun has exhaustive narrative docs that cover every piece of API. Having to wade thru some weird util functions before the important stuff is positively annoying to me.

webknjaz: I'm not suggesting that the order should be random, though.

PR #1081, opened 2023-01-06
Fix @frozen exceptions to allow __traceback__ to be set.

Julian:
# Summary

This code previously raised an `attrs.FrozenInstanceError`:

```python
from contextlib import contextmanager
from attrs import frozen

@frozen
class Boom(Exception):
    pass


@contextmanager
def do_nothing():
    yield


with do_nothing() as _:
    raise Boom()
```

on CPython (3.11 specifically, but probably other versions). On PyPy, it works, at least on 3.9 here -- which means I think it seems reasonable to merge the two codepaths here as I've done rather than whitelisting different attributes on different interpreters, but lemme know if you disagree.

(In real life a likely way you're going to encounter this, just to help anyone searching, is that `click` will wrap all exceptions during execution of a program, and does so via a `contextlib.contextmanager`)

Refs #712

# Pull Request Check List

### DIFF
--- a/changelog.d/1081.change.md
+++ b/changelog.d/1081.change.md
@@ -1,0 +1,1 @@
+Fix frozen exception classes when raised within e.g. `contextlib.contextmanager`, which mutates their `__traceback__` attributes.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -12,7 +12,7 @@ from operator import itemgetter
 # We need to import _compat itself in addition to the _compat members to avoid
 # having the thread-local in the globals here.
 from . import _compat, _config, setters
-from ._compat import PY310, PYPY, _AnnotationExtractor, set_closure_cell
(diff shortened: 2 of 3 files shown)


hynek:
Ah nice I'm pretty sure I ran junk this myself when switching to 3.11 but decided to just thaw the class and get on with my life. 

Can you finish it up with a news fragment pls?

Julian:
Fixed!

changelog.d/1081.change.rst
```diff
@@ -0,0 +1 @@
+Fix frozen exception classes when raised within e.g. `contextlib.contextmanager`, which mutates their ``__traceback__`` attributes.
```
hynek, line 1: Sir this is a ~Wendy's~ Markdown shop. 🤓

Julian: Ups, fix fixed then.

hynek (approved):
Excellent thank you!

PR #1092, opened 2023-01-28
compat: move check for closure cellvar update wrapper function

torque:
This sanity checking code ended up getting decoupled from the old implementation in a18b3957d55c669de56799a9a72ea31febe93ce4 when a new, much simpler method was added. This check can fail spuriously in certain environments where it doesn't even need to be run because the newer approach is used.

This is noticeable when using python scripts compiled by [nuitka](https://github.com/Nuitka/Nuitka). Since next-gen `attrs` uses slots on everything by default, the use of this fixup function is now pervasive, as well.

Here's the simplest reproduction I came up with (tested on Python 3.10.9 and nuitka 1.4)

```python
# problem.py

# pip install attrs nuitka ordered-set
# nuitka3 --standalone --onefile problem.py
# running `python problem.py` without this patch succeeds
# running nuitka's output `problem.bin` without this patch fails. With this patch it succeeds.

import attrs

@attrs.define
class Parent:
    a: int

    def __attrs_post_init__(self):
        assert self.__class__ is Child, f"{self.__class__} is not {Child} ({id(self.__class__)}, {id(Child)})"

@attrs.define
class Child(Parent):
    b: float

    def __attrs_post_init__(self):
        assert __class__ is Child, f"{__class__} is not {Child} ({id(__class__)}, {id(Child)})"

        super().__attrs_post_init__()


Child(a=1, b=0.0)
```

Because `nuitka` compiles the python code to C, I guess it loses the introspection through the `__code__` attribute, which is why the sanity check causes a problem. Fortunately, that's apparently entirely unneeded with any python version 3.8 or above.

# (Annotated) Pull Request Check List

   >  I am not sure how to reasonably add a test case for this in-tree because my reproduction requires a heavyweight external process (compiling with nuitka).
   > the changed code is an internal function with minimal documentation. There doesn't seem to be anything to add here

### DIFF
--- a/changelog.d/1092.change.md
+++ b/changelog.d/1092.change.md
@@ -1,2 +1,1 @@
-Fix slots class cellvar updating closure in CPython 3.8+ even when `__code__`
-introspection is unavailable.
+Fix slots class cellvar updating closure in CPython 3.8+ even when `__code__` introspection is unavailable.
--- a/changelog.d/1092.change.md
+++ b/changelog.d/1092.change.md
@@ -1,0 +1,2 @@
+Fix slots class cellvar updating closure in CPython 3.8+ even when `__code__`
+introspection is unavailable.
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -81,32 +81,32 @@ def make_set_closure_cell():
 
     # Otherwise gotta do it the hard way.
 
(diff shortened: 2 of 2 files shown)


Tinche (approved):

hynek:
Thanks!

PR #1093, opened 2023-01-28
Unpin tox

hynek:
It works locally for me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
         run: |
           python -VV
           python -m site
-          python -m pip install --upgrade wheel 'tox<4' tox-gh-actions
+          python -m pip install --upgrade wheel tox tox-gh-actions
 
       - run: python -m tox
 
@@ -131,7 +131,7 @@ jobs:
         with:
           python-version: "3.10"
 
-      - run: python -m pip install --upgrade wheel 'tox<4'
+      - run: python -m pip install --upgrade wheel tox
       - run: python -m tox -e docs,changelog
 
   pyright:
@@ -155,7 +155,7 @@ jobs:
         with:
           python-version: ${{env.PYTHON_LATEST}}
 
-      - run: python -m pip install --upgrade wheel 'tox<4'
+      - run: python -m pip install --upgrade wheel tox
       - run: python -m tox -e pyright
 
   package:
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,7 @@ commands =
 [testenv]
 extras = tests
 commands = python -m pytest {posargs:-n auto}
+passenv = SETUPTOOLS_SCM_PRETEND_VERSION 
 
 
 [testenv:py3{7,10,11}]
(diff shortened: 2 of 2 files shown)


hynek:
no it doesn't

webknjaz:
Wait, if it works locally but not in GHA, maybe it's the plugin's fault? I think I've seen a corresponding plugin for tox4 being separate from the one for tox3.

hynek:
no, I ran `tox -e mypy` locally but the broken thing is the test suite. I've isolated & opened a bug: https://github.com/tox-dev/tox/issues/2904

webknjaz:
Ah, I see. But FTR there's https://github.com/tox-dev/tox-gh too.

hynek:
What does that to do with anything? Our problem is that the package pins for Mypy are from PyPI instead of from the local package. I don't see how the plugins play into that.

webknjaz:
It's unrelated to the mypy problem but is related to the tox4 migration, IIUC.

hynek:
> It's unrelated to the mypy problem but is related to the tox4 migration, IIUC.

According to https://pypi.org/project/tox-gh-actions/, tox-gh-actions>=3 works with tox 3 just fine.

I would switch if there's any upsides, but I can't see any rn?

---

Anyways, tox [has fixed the bug](https://github.com/tox-dev/tox/pull/2905) 🎉 and we're waiting for a release now.

webknjaz:
> According to https://pypi.org/project/tox-gh-actions/, tox-gh-actions>=3 works with tox 3 just fine.

Looks like it's `~= 3.0` though.

webknjaz:
FWIW here's some notes regarding the differences https://github.com/tox-dev/tox-gh/issues/7#issuecomment-1013651022

hynek:
what the flying fuck is going on now… it works perfectly fine on my machine!?

webknjaz:
Looks related to the recent Hatch-scm integration. At least one of the logs failed on `int('dev2')`.
Also, I saw a resource warning so there might be an unclosed file descriptor somewhere.

hynek:
but but but… it used to work! otherwise i wouldn’t have merged it! :(

I’m currently in the process of setting up a new computer, once I’m fully functional I’ll look closer *sigh*

webknjaz:
FWIW I checked this branch locally and it does work. Maybe somebody needs to stick https://github.com/marketplace/actions/debugging-with-tmate there and inspect the worker over SSH.

Oh, and I just remembered the most typical setuptools-scm mistake — the contributors usually have a full Git clone locally but CIs tend to do shallow clones. The fix is usually to set the fetch depth to 0 or run git unshallow.

Though, maybe this is not the only problem here, I haven't had time to dig deeper.

hynek:
https://github.com/python-attrs/attrs/pull/1093/commits/0c9b54c887ad117a2c31a91bbbf65341e11a3131 #genius

hynek:
FTR on tox-gh: it seems like it has a very different way of working than tax-gh-actions. I tries using it with structlog and `3.11 = py311` means literally "only run py311" while `tox-gh-actions` essentially says `-f py311`. For attrs it doesn't matter for now, but that's a very significant different:

CleanShot 2023-02-02 at 08 16 50@2x

(using screenshot b/c logs get lost over time)

PR #1094, opened 2023-01-29
Switch to hatch-vcs

hynek:
We're already half-way there, so let's finish this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,8 @@ permissions:
 
 jobs:
   version:
+    # To save ourselves deep checkouts in every job, we determine the in-dev
+    # version once and set SETUPTOOLS_SCM_PRETEND_VERSION to it.
     name: Determine in-dev version
     runs-on: ubuntu-latest
 
@@ -56,6 +58,7 @@ jobs:
       ${{ contains(matrix.python-version, '~') && true || false }}
 
     steps:
+      - run: echo ${{ env.SETUPTOOLS_SCM_PRETEND_VERSION }}
       - name: Harden Runner
         uses: step-security/harden-runner@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,16 +15,26 @@ env:
   PIP_DISABLE_PIP_VERSION_CHECK: "1"
   PIP_NO_PYTHON_VERSION_WARNING: "1"
   PYTHON_LATEST: "3.11"
-  SETUPTOOLS_SCM_PRETEND_VERSION: "1.0" # avoid warnings about shallow checkout
-
 
 permissions:
   contents: read
 
 jobs:
+  version:
+    name: Determine in-dev version
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - run: python -Im pip install --upgrade hatch
+      - run: echo SETUPTOOLS_SCM_PRETEND_VERSION=$(hatch version) >>$GITHUB_ENV
+
   tests:
     name: tox on ${{ matrix.python-version }}
     runs-on: ubuntu-latest
+    needs: version
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ env:
   PIP_DISABLE_PIP_VERSION_CHECK: "1"
   PIP_NO_PYTHON_VERSION_WARNING: "1"
   PYTHON_LATEST: "3.11"
+  # avoid warnings about shallow checkout, but also make sure pytest accepts us.
+  SETUPTOOLS_SCM_PRETEND_VERSION: "99.9.9"
```
webknjaz, line 19: Is this related to hatch? I know this is specific to `setuptools-scm`. Does hatch use it?

webknjaz: Ah, I see, it's a wrapper.

.github/workflows/ci.yml
```diff
@@ -158,23 +160,6 @@ jobs:
       - run: python -m pip install --upgrade wheel 'tox<4'
       - run: python -m tox -e pyright
 
-  package:
```
webknjaz, line 161: @hynek is there a reason you don't want to run tests on the sdist?
Would you be open to trying out my new(ish) action? https://github.com/re-actors/checkout-python-sdist — it's essentially a mechanism to use an sdist instead of a Git checkout. This should bring more reliability + ensure the sdists are downstream-friendly.

hynek: it's on my radar, but i wanted to wait for it to stabilize a bit before embracing it

i'd be open to use it in one of my smaller projects like environ-config i guess

webknjaz: Ah, okay. I'm not planning on changing the public API there: I've started it as a part of the experiment to make an action (and later, a reusable workflow), I was using it through the `@master` version and at some point just decided to tag it properly with probably a year of zero changes. Plus it's quite simple, implementation-wise.

webknjaz (commented):
@hynek so since this wraps `setuptools-scm`, it appears to me that its support for git archives should also work. But said feature requires `.git_archival.txt` and `.gitattributes` to enable everything needed on Git (and GitHub) side. Is this an oversight?

hynek:
> @hynek so since this wraps `setuptools-scm`, it appears to me that its support for git archives should also work. But said feature requires `.git_archival.txt` and `.gitattributes` to enable everything needed on Git (and GitHub) side. Is this an oversight?

I don't know what that means tbh :)

webknjaz:
> I don't know what that means tbh :)

When you run `pip install https://github.com/python-attrs/attrs/archive/refs/tags/22.2.0.tar.gz`, here's what happens:
1. `pip` downloads a source archive from GitHub (not sdist!)
2. `pip` builds a wheel out of it
3. `setuptools` figures out the version for that artifact
	1. Currently, it'll default to `0.0.0` IIRC, since `setuptools-scm` does not see any Git repo on disk to calculate the version from
	2. If the tarball in (1) would be an sdist, it'd just grab the version from its `PKG-INFO` metadata file
	3. If there was a `.git_archival.txt` template with `.gitattributes` for it, it'd get the version from it. The key is having it set up properly
4. `pip` takes care of the deps
5. `pip` unpacks said wheel into `site-packages/`

Back to the git archive thingy. That URL (1) contains an artifact produced by the `git archive` command on GitHub's servers. `.gitattributes` can be set up to make Git apply substitutions at the time of exporting a git version into a tarball. And `.git_archival.txt` is a name recognized by `setuptools-scm` (3.3). It should be checked into Git, in its template form along with the attributes config — this makes sure that the exported tarballs would have this file but with the template rendered (the syntax is normal git format for its log output). You don't really need to know this syntax — copying the files into the project as instructed is enough to make everything work. Basically, this includes some metadata about the commit being exported and the tag distance necessary for `setuptools-scm` to make version calculations later.

In the past, this required using an extra plugin — `setuptools-scm-git-archival` and only worked for tag URLs, but modern Git has extended the syntax to allow bundling more information in the export and this allows installing from arbitrary commit URLs with proper intermediate versions which is why that plugin got absorbed into the `setuptools-scm` itself.

@hynek I hope this sheds some light on the problem.

webknjaz:
FWIW I've made https://github.com/python-attrs/attrs/pull/1095 to address this.

PR #1097, opened 2023-02-02
Switch to tox-gh

hynek:
The relationship seems a bit too antagonist to my taste.


hynek:
tox-gh – unlike tox-gh-actions seems to be very literal about the selection of environments:

CleanShot 2023-02-02 at 08 16 50@2x

This can be solved better.

PR #1106, opened 2023-02-25
Slightly refined documentation while reading it

raratiru:
# Summary

I have fixed some typos and rephrased a few sentences in the docs

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -213,10 +213,11 @@ Traceback (most recent call last):
 ValueError: 'x' has to be smaller than 'y'!
 ```
 
-This example also shows of some syntactic sugar for using the {obj}`attrs.validators.and_` validator: if you pass a list, all validators have to pass.
+In this example, the list of validators passed to the `validator` argument, is syntactic sugar for the {obj}`attrs.validators.and_` validator: All validators in the list have to pass.
 
 *attrs* won't intercept your changes to those attributes but you can always call {func}`attrs.validate` on any instance to verify that it's still valid:
-When using {func}`attrs.define` or [`attrs.frozen`](attrs.frozen), *attrs* will run the validators even when setting the attribute.
+
+When using {func}`attrs.define` or [`attrs.frozen`](attrs.frozen), however, *attrs* will run the validators even when setting the attribute.
 
 ```{doctest}
 >>> i = C(4, 5)
@@ -241,7 +242,7 @@ TypeError: ("'x' must be <type 'int'> (got '42' that is a <type 'str'>).", Attri
 ```
 
 Of course you can mix and match the two approaches at your convenience.
-If you define validators both ways for an attribute, they are both ran:
+If you use both ways to define validators for an attribute, they are both ran:
 
 ```{doctest}
 >>> @define
@@ -263,7 +264,7 @@ Traceback (most recent call last):
 ValueError: value out of bounds
 ```
 
-And finally you can disable validators globally:
+Finally, validators can be globally disabled:
 
 ```{doctest}
 >>> attrs.validators.set_disabled(True)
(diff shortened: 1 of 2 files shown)


docs/init.md
```diff
@@ -213,10 +213,11 @@ Traceback (most recent call last):
 ValueError: 'x' has to be smaller than 'y'!
 ```
 
-This example also shows of some syntactic sugar for using the {obj}`attrs.validators.and_` validator: if you pass a list, all validators have to pass.
+In this example, the list of validators passed to the `validator` argument, is syntactic sugar for the {obj}`attrs.validators.and_` validator: All validators in the list have to pass.
```
chrysle, line 216: I think I liked the former formulation better, since it isn't the list that is _syntactic sugar_, but the usage of the validator.

raratiru: Indeed that was the most difficult. 

Because I understood the example (that all validators in the list should pass), I could not understand "why I read this sentence?". 

It took me some time to grasp the connection between the list of validators and `attrs.validatos.and_`.

What about:

>This example is also showing an alternative use (i.e. syntactic sugar) for the {obj}`attrs.validators.and_` validator: 
All validators in the list have to pass.

chrysle: > What about:
> 
> > This example is also showing an alternative use (i.e. syntactic sugar) for the {obj}`attrs.validators.and_` validator:
> > All validators in the list have to pass.

I'd suggest: 

> This example shows a syntactic speciality for the {obj}`attrs.validators.and_` validator: if you pass a list, all validators in it have to pass.

raratiru: That is very good, I create a commit with it, thank you!

docs/init.md
```diff
@@ -292,8 +293,7 @@ TypeError: ("'x' must be <class 'int'> (got '128' that is a <class 'str'>).", At
 
 ## Converters
 
-Finally, sometimes you may want to normalize the values coming in.
-For that *attrs* comes with converters.
+Sometimes, is necessary to normalize the values coming in, therefore *attrs* comes with converters.
```
hynek, line 296:
```suggestion
Sometimes, it is necessary to normalize the values coming in, therefore *attrs* comes with converters.
```

docs/init.md
```diff
@@ -350,10 +350,10 @@ A converter will override an explicit type annotation or `type` argument.
 
 ## Hooking Yourself Into Initialization
 
-Generally speaking, the moment you think that you need finer control over how your class is instantiated than what *attrs* offers, it's usually best to use a {obj}`classmethod` factory or to apply the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern).
+Generally speaking, the moment you realize the need of finer control -than what *attrs* offers- over how a class is instantiated, it's usually best to use a {obj}`classmethod` factory or to apply the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern).
```
hynek, line 353: We use space-endash-space:

```suggestion
Generally speaking, the moment you realize the need of finer control – than what *attrs* offers – over how a class is instantiated, it's usually best to use a {obj}`classmethod` factory or to apply the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern).
```

docs/init.md
```diff
@@ -213,10 +213,11 @@ Traceback (most recent call last):
 ValueError: 'x' has to be smaller than 'y'!
 ```
 
-This example also shows of some syntactic sugar for using the {obj}`attrs.validators.and_` validator: if you pass a list, all validators have to pass.
+This example shows a syntactic speciality for the {obj}`attrs.validators.and_` validator: If you pass a list, all validators in it have to pass.
```
hynek, line 216: I've seen y'all's discussion, but I think I prefer a third option:

```suggestion
This example demonstrates a convenience shortcut:
Passing a list of validators directly is equivalent to passing them wrapped in the {obj}`attrs.validators.and_` validator and all validators must pass.
```

hynek (changes requested):
Thanks, a few quibbles that should be easy to fix.

hynek (approved):

PR #1111, opened 2023-03-11
Fix for new Mypy

Tinche:
They changed their error messages (and for the better, I think).


hynek (approved):

hynek:
Thanks! I wish the failures were more obvious in CI.

PR #1117, opened 2023-04-03
Raise a deprecation warning when evolve receives insta as a kw arg

### LINKED ISSUE #1109: Make evolve 1st arg positional-only
It's arbitrary that the 1st argument is named `inst` and could conflict with a field.

Since PEP-570 positional-only arguments require Python 3.8 and next major version will be ≥ Python 3.7, we can do it the old-fashion way in the next breaking version (as this is an API change), i.e.
```python
def evolve(*args, **changes):
  (inst,) = args
  ...
```

hynek:
this looking ok @ikonst?

Fixes #1109

### DIFF
--- a/changelog.d/1117.change.md
+++ b/changelog.d/1117.change.md
@@ -1,3 +1,3 @@
-Passing the instance to `attrs.evolve()` (and `attr.evolve()`) using a keyword argument is now deprecated and will be removed in, or after, April 2024.
+It is now possible for `attrs.evolve()` (and `attr.evolve()`) to change fields named `inst` if the instance is passed as a positional argument.
 
-This makes it possible to change a field named `inst`.
+Passing the instance using the `inst` keyword argument is now deprecated and will be removed in, or after, April 2024.
--- a/changelog.d/1117.change.md
+++ b/changelog.d/1117.change.md
@@ -1,0 +1,3 @@
+Passing the instance to `attrs.evolve()` (and `attr.evolve()`) using a keyword argument is now deprecated and will be removed in, or after, April 2024.
+
+This makes it possible to change a field named `inst`.
--- a/src/attr/_funcs.py
+++ b/src/attr/_funcs.py
@@ -2,7 +2,6 @@
 
 
 import copy
-import warnings
 
 from ._compat import PY_3_9_PLUS, get_generic_base
 from ._make import NOTHING, _obj_setattr, fields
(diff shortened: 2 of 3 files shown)


Tinche:
How bout we remove `assoc`? I think we're passed 2018 by now. 😇

hynek:
I think we should change the deprecation, actually. People still use it, because it works differently. I don't remember the use-case but there was one. :-/

src/attr/_funcs.py
```diff
@@ -364,8 +364,26 @@ def evolve(inst, **changes):
...
+    if args:
+        (inst,) = args
+    else:
+        warnings.warn(
+            "Passing the instance per keyword argument is deprecated and "
+            "will stop working in, or after, April 2024.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        inst = changes.pop("inst")
```
ikonst, line 385: Should we translate the `ValueError` (1) and `KeyError` (2) into a clearer `TypeError` as Python would emit?

hynek: good point, check out now!

hynek:
> How bout we remove `assoc`? I think we're passed 2018 by now. 😇

lol it was even you who found out: https://github.com/ericvsmith/dataclasses/issues/42#issuecomment-330309215

src/attr/_funcs.py
```diff
@@ -365,8 +366,40 @@ def evolve(inst, **changes):
...
+        except ValueError:
+            raise TypeError(
+                f"evolve() takes 1 positional argument, but {len(args)} "
+                "were given"
+            )
+    else:
+        try:
+            inst = changes.pop("inst")
+        except KeyError:
+            raise TypeError(
```
ikonst, line 390: Consider raise from None

src/attr/_funcs.py
```diff
@@ -365,8 +366,40 @@ def evolve(inst, **changes):
...
+       it will become an error. Always pass the instance as a positional
+       argument.
     """
+    # Try to get instance by positional argument first.
+    # Use changes otherwise and warn it'll break.
+    if args:
+        try:
+            (inst,) = args
+        except ValueError:
+            raise TypeError(
```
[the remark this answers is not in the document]
hynek: right, I felt like that's super new but… oof 3.3

changelog.d/1117.change.md
```diff
@@ -1,3 +1,3 @@
-Passing the instance to `attrs.evolve()` (and `attr.evolve()`) using a keyword argument is now deprecated and will be removed in, or after, April 2024.
+It is now possible for `attrs.evolve()` (and `attr.evolve()`) to change fields named `inst` if the instance is passed as a positional argument.
```
ikonst, line 1: ➕

hynek:
Thanks for your help @ikonst – we've smuggled a small feature in!

PR #1120, opened 2023-04-05
Deprecate zope-interface support

hynek:
It's a C-based dependency for a dozen of lines of code.

If there's interest, we can publish it as a separate package.


chrysle:
Why, that's quite a lot being deprecated in or after april 2024! How do you keep track?

hynek:
I don’t. :D

Things get usually removed when you stumble over deprecated code while trying to do something else. I guess there’s not that much of a downside to remove deprecated code later.

pganssle (approved):
Seems good to me. My main involvement with this was that we *don't* use it but it was annoying to package `zope` just to run the `attrs` tests. The current situation, where `zope` is an optional test dependency, works perfectly for our purposes, but I think that droping mainline support for `zope` and moving it to a third-party package if there's interest is a good, and probably more sustainable approach in the long term.

itamarst:
Tahoe-LAFS is using this, but I'm not sure it's worth your time to package this separately—we could just copy/paste the relevant code into the codebase with appropriate license/attribution? Or I guess since we're using `mypy-zope` just rely on that.

PR #1130, opened 2023-04-26
Fix intersphinx

hynek:
They've deprecated the old syntax.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -159,9 +159,7 @@ texinfo_documents = [
 
 epub_description = "Python Clases Without Boilerplate"
 
-intersphinx_mapping = {
-    "https://docs.python.org/3": None,
-}
+intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
 
 # Allow non-local URIs so we can have images in CHANGELOG etc.
 suppress_warnings = ["image.nonlocal_uri"]


docs/conf.py
```diff
@@ -159,9 +159,7 @@
 
 epub_description = "Python Clases Without Boilerplate"
 
-intersphinx_mapping = {
-    "https://docs.python.org/3": None,
-}
+intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
```
webknjaz, line 162: Pro tip: the second value in the tuple is a fallback which can be a cached file in the repo. It tends to be useful when working on stuff offline.

hynek: I don't know what that means. 🙈

chrysle: > I don't know what that means. 🙈

See [here](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping). This for example:

```python
intersphinx_mapping = {"python": ("https://docs.python.org/3", "python-inv.txt")}
``` 

Will read the inventory file `python-inv.txt` from the base directory if no internet connection is available. The inventory file would otherwise be downloaded from https://docs.python.org/3/objects.inv (as far as I understood).

I guess that's useful?

webknjaz: Yep. It's not very well know. We have a script to refresh the object inventory files in the Ansible repo, I think. Also, while googling, I saw something similar/reusable from the twisted ecosystem. Helps beat flaky network, especially if you have nitpicky/strict modes enabled.

chrysle: > We have a script to refresh the object inventory files in the Ansible repo, I think. Also, while googling, I saw something similar/reusable from the twisted ecosystem.

Do you have links?

chrysle: @webknjaz Should I try implementing the fallback and refreshing mechanism (I would appreciate a link to a script though)?

webknjaz: Let's wait for @hynek to state his preference. My comment was mostly educational.

But if you want to look into this for yourself, here's some pointers:

- Ansible's download snippet is here https://github.com/ansible/ansible/blob/dbd3bf2/hacking/build_library/build_ansible/command_plugins/update_intersphinx.py#L48-L93
- https://pypi.org/project/sphobjinv/ seems to be able to download these files (though, regular curl/wget can do the same)
- https://pypi.org/project/pydoctor/ is a complex thing from the Twisted ecosystem. It's probably interesting to inspect internally

Of course, if something is to be implemented, it shouldn't include copy-pasting stuff but rather reusing things that are just enough.

hynek: Yeah no pls don't make this any more complex. :)

webknjaz: Yeah, that's the reaction I anticipated :)

PR #1140, opened 2023-06-04
Simplify CI and use .python-version

hynek:


webknjaz:
Looking at the gitignore removal, I remembered that it could be useful to make a `changelog.d/README.md` — it is rendered on github in non-root directories too which is quite useful to contributors browsing for change note examples.

.python-version
```diff
@@ -0,0 +1 @@
+3.11
```
webknjaz, line 1: @hynek this is going to cause friction for the contributors using pyenv (like me) since this file is managed by the `pyenv local` command. I strongly recommend gitignoring it as it tends to get in the way.
It is meant to be controlled by the end-users and for me, it often contains a virtualenv name, maybe with a few fallbacks.
Also, by using an unversioned `3.11` w/o a patch component, it's bound to introduce surprise “command not found” whenever a contributor installs a slightly newer version (as in `pyenv install 3.11.1` followed by `pyenv install 3.11.4` months later, while working on something else).
This is why, in my opinion, it's not worth intruding into contributors' setups.

.github/workflows/ci.yml
```diff
@@ -14,10 +14,10 @@ env:
   TOX_TESTENV_PASSENV: FORCE_COLOR
   PIP_DISABLE_PIP_VERSION_CHECK: "1"
   PIP_NO_PYTHON_VERSION_WARNING: "1"
-  # Keep in sync with .pre-commit-config.yaml/default_language_version/python.
-  PYTHON_LATEST: "3.11"
```
webknjaz, line 18: @hynek FYI, there's a new context in GHA — `${{ vars }}`. You can control it from a centralized location like repo settings, it could be accessed as `${{ vars.PYTHON_LATEST }}` then.

.github/CONTRIBUTING.md
```diff
@@ -41,10 +41,17 @@ The official tag is `python-attrs` and helping out in support frees us up to imp
 
 You can (and should) run our test suite using [*tox*].
 However, you’ll probably want a more traditional environment as well.
-We highly recommend to develop using the latest Python release because we try to take advantage of modern features whenever possible.
-Also, running [*pre-commit*] later on will require the latest Python version.
 
-First [fork](https://github.com/python-attrs/attrs/fork) the repository on GitHub.
+First, create a [virtual environment](https://virtualenv.pypa.io/) so you don't break your system-wide Python installation.
+We recommend using the Python version from the `.python-version` file in project's root directory.
```
webknjaz, line 46: @hynek to prevent accidental commits from pyenv users doing `pyenv local`, we'd need to instruct all the contributors to run `git update-index --assume-unchanged .python-version` or a similar exclusion. That is, if you still want to keep this after my other comments.

.python-version
chrysle: I actually wonder which use such a file has, holding a single python version. Since it doesn't get updated automatically?

hynek: It’s the other way round: it’s the only standard we have to set a default version that’s respected by several tools including setup-python. It’s nice to have a canonical truth _somewhere_.

I’ve seen GHA offers a way to use a different name for that file but I’m not sure if it doesn’t create more problem than it solves…

.github/workflows/ci.yml
webknjaz: Then, the contributing guide should hint the contributors how to ignore this invasive config in Git, I think.

.python-version
webknjaz: > I actually wonder which use such a file has, holding a single python version. Since it doesn't get updated automatically?

@chrysle, if you set up your shell to use pyenv and a matching Python version or venv exists (could be several values, actually), it'll automatically make use of those versions listed in `.python-version` (or activate a venv). If there's no such version, it'll fall back to the pyenv global value. As it stands now, it effectively breaks the `pyenv-virtualenv` use case (with project-specific envs being picked up by `pyenv local`). Well, kinda. The `pyenv local` command updates the contents of `.python-version` and then, Git _sees_ the file as updated. This would likely be misleading for some people that don't know how to get Git to pretend there's no changes.

Issue #934, opened 2022-03-17
Brainstorming `fields()`


Tinche:
Hello. Due to... circumstances... (mostly https://github.com/python/mypy/issues/5144) looks like if we want proper `fields()` support in Mypy, we'll have to implement it ourselves in the attrs plugin. I volunteer to do this, and have a version ready in a branch.

Now, if we're already implementing special logic for this in the Mypy plugin we might as well discuss potential improvements.

Let me start with an example, to set the tone.

Right now, fields() returns a namedtuple of `attr.Attribute`. This is quite sufficient for introspection (so, enough for cattrs) but it's a little weak for implementing simple ORMs/ODMs. Which could be why we haven't seen many of those based on attrs.

I'm working on open sourcing our internal Mongo ODM, where classes are defined using attrs. Two interesting use cases come to mind. 

# Example 1: loading projected data.

```python
from attrs import fields as f, define

@define
class Model:
    _id: ObjectId
    username: str

res = await mongo_client.find(Model, {"_id": ...}, projection=[f(Model)._id, f(Model).username])
```
This works in runtime, and will work statically once the plugin implements `fields`. One improvement that comes to mind here would be having the `__attrs_attrs__` field available under a different name. For example, we could do something like SQLAlchemy and have it under `Model.c`. The query would then be:

```python
res = await mongo_client.find(Model, {"_id": ...}, projection=[Model.c._id, Model.c.username])
```
This can work today in runtime, but not statically. Maybe we could think of a way to make it declarative and then support it in Mypy?

Might be nice enough to use for the query condition as well.

```python
res = await mongo_client.find(Model, {Model.c._id: ...}, projection=[Model.c._id, Model.c.username])
```

I guess a shitty thing about the `.c` approach is that you cannot have an attribute named `c`, and it's essentially arbitrary. So maybe it should be configurable? A good thing about it is that it's easier on the eyes and no overhead of a function call.

# Example 2: using a different Attribute class

I don't have this particular problem in Mongo since queries in Mongo are just bson documents, but if you tried replicating some of SQLAlchemy you'd run into it. You might want to add functionality to `attr.Attribute` to be able to write things like:
```
condition = f(User).username == "Tin"
```
This would just produce a False, instead of something else you might want to use later.

If we could supply our own attribute class to `fields` we might be able to deal with this issue. So `fields` might look like:

```
def fields(cls, *, attr_cls=None): ...  # Return a `NamedTuple[attr_cls, ...]` if provided instead
```
[...]

PR #1144, opened 2023-06-08
Initial custom attributes work

Tinche:
This PR enables other libraries to define their own Attribute classes (replacing attrs.Attribute), and hence to define their own `fields()` function.

_Why does this need to be in attrs?_

Because the next step will be to add support for this to the attrs Mypy plugin, and we need to design this feature with that in mind.

Having Mypy support for this will actually enable statically checking things that were previously _impossible_ in Python, so if we do it right, it might be a big deal.

For example, as far as I know no ORM in Python today can statically check queries. We may change this.

Since this is a feature for library authors (not end users directly), we may consider designing it so it is easily supportable in Mypy rather than easy to use (as funny as that sounds).

My personal motivation is that my codebases are pretty heavily typed, but I keep running into untypeable things to do with databases. I would love to plug the hole, even partially.

This is a continuation of https://github.com/python-attrs/attrs/issues/934.

# Use cases

## cattrs customization

Right now, here's how a cattrs structuring function can be customized:

```python
from cattrs.gen import override, make_dict_structure_fn

make_dict_structure_fn(MyClass, converter, a=override(omit=True))
```

Here's how it could be:

```python
from cattrs.gen import make_dict_structure_fn, fields as f

make_dict_structure_fn(MyClass, converter, f(MyClass).a.omit())
```

_Why is it better?_

A more fluent API, better errors, statically checked.

## cattrs validation

cattrs has no validation right now, here's how it could work.

```python
from cattrs.v import fields as f

validator = f(MyClass).a.gt(5) & f(MyClass).b.len(min=1)

validator(MyClass(1, [])
```

## PocketMongo query builder

The current approach to creating a Mongo query condition:

```python
{"user_id": user_id}
```

How it could be:

```python
from pocketmongo import q, find_one

find_one(q(Model).user_id.equal(uid))
# Mypy error if `user_id` is misspelled (field doesn't exist)
# Mypy error if `uid` is not the same type as `user_id`
```

Why is it better:

A more fluent API. Statically checked.

This might end up being a model for attrs-based ORMs going forward.

# Problems

* Pyright probably won't support this, but that's just something we have to live with. Mypy supporting it will add plenty of value.

* I'm not sure about handling nested attributes yet. If class A has an attribute `b`, typed as class `B`, and class `B` has a field called `c`. `f(A).b.c` should probably work, but it maybe requires special casing? That's TBD, but sounds like mostly a Mypy issue.

### DIFF
--- a/src/attrs/__init__.py
+++ b/src/attrs/__init__.py
@@ -43,6 +43,7 @@ __all__ = [
     "AttrsInstance",
     "cmp_using",
     "converters",
+    "custom_fields",
     "define",
     "evolve",
     "exceptions",
--- a/src/attrs/custom_fields.py
+++ b/src/attrs/custom_fields.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+import typing
+
+from typing_extensions import Protocol
+
+from attr._make import _make_attr_tuple_class
+from attrs import Attribute, AttrsInstance, fields
+from attrs import resolve_types as _resolve_types
+
+
+__all__ = ["custom_fields"]
+
+T = typing.TypeVar("T")
+
+
+class AttributeModel(Protocol[T]):
+    """Custom attributes must conform to this."""
+
+    @classmethod
+    def _from_attrs_attribute(
+        cls: type[AttributeModel],
+        cl: type[AttrsInstance],
+        attribute: Attribute[T],
+    ) -> AttributeModel[T]:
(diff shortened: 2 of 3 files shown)


Mybeterlife-continue2025mylife (approved):
https://github.com/skills/introduction-to-github/tree/main

2f5fab60071d9d37bcd766bf0674377171ba129f

Mybeterlife-continue2025mylife (commented):
my+succxx1

PR #1146, opened 2023-06-13
Use a distinct file name for default python versions

hynek:
Does this solve your woes @webknjaz?


.github/workflows/ci.yml
```diff
@@ -48,9 +46,7 @@ jobs:
           allow-prereleases: true
           cache: pip
 
-      - run: python -Im pip install --upgrade tox
-
-      - name: Determine Python version for tox
+      - name: Prepare & run tox
         run: |
```
webknjaz, line 50: _Educational comment:_

FYI if you put multiple commands into a shell step, only the last one's return code is taken into account. Hence, this allows `pip install`, or `tr`, or a pipe failure to go unnoticed since the step could be green anyway (if there's an older tox present and the tests succeed).
This might seem harmless but it might also cause some debugging pain. Which is why I generally recommend having a single shell command per step.
A workaround could be joining the commands with `&&` (which is only a partial solution) or adding `set -eEuo pipefail` as the first one to each such step.

```suggestion
         run: |
          set -eEuo pipefail
```

webknjaz (commented):
> Does this solve your woes

Yes, thanks!

Though, I noticed another problem that you might want to look into. See the inline comment.

.github/workflows/ci.yml
hynek: Wait, I used to set `-e` and someone told me that's set automatically in GHA. 😡

Since each step takes a while, I'm trying to lower the number of them. :|

hynek: oh, look it does: https://github.com/python-attrs/attrs/actions/runs/5252364725/jobs/9488338346?pr=1146#step:4:26

They set `ehB` so we should be fine?

hynek: if i run into stupid stuff, i'll just revert it :)

webknjaz: > They set ehB

They only set `-e`, the rest is inherited:

> ```console
>  shell: /usr/bin/bash -e {0}
> ```

This can be changed through `defaults:`.

PR #1154, opened 2023-06-26
Build only one universal wheel when running tox

hynek:
Sadly, hatch-autorun and tox seem to be fighting over env variables so we create the pth file by hand.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,16 +111,6 @@ text = """
 """
 
 
-# Make coverage play nicely with pytest-xdist.
-[tool.hatch.build.targets.wheel.hooks.autorun]
-dependencies = ["hatch-autorun"]
-code = """
-import coverage
-coverage.process_startup()
-"""
-enable-by-default = false
-
-
 [tool.pytest.ini_options]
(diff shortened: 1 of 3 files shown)


hynek:
OK, this is interesting. @gaborbernat it looks like `SETUPTOOLS_SCM_PRETEND_VERSION` is **not** passed through when using the shared build environment despite being in `[testenv]`'s `pass_env`.

I couldn't find a way to affect env variables for it (which is why I gave up on hatch-autorun that needs one) – any advice? Would you consider this a bug in tox?

gaborbernat:
> OK, this is interesting. @gaborbernat it looks like `SETUPTOOLS_SCM_PRETEND_VERSION` is **not** passed through when using the shared build environment despite being in `[testenv]`'s `pass_env`.
> 
> I couldn't find a way to affect env variables for it (which is why I gave up on hatch-autorun that needs one) – any advice? Would you consider this a bug in tox?

https://tox.wiki/en/latest/upgrading.html#packaging-configuration-and-inheritance ?

gaborbernat:
Note the new config subcommand `tox c -e py311 -k pass_env set_env` that would have allowed you to inspect these settings.

Issue #1162, opened 2023-07-08
Using `attr` library instead of `attrs` in documentation


Why-not-now:
In the documentation, there is a mix of usage of the libraries `attr` and `attrs`. One example of this is in [here](https://www.attrs.org/en/stable/examples.html#immutability), where the docs referenced `attr.evolve` and used `attrs.evolve` in the example code. This confused me on the first read through of the documentation. Is updating the documentation to be using the library `attrs` in the works?

hynek:
Generally speaking we try to use `attrs.` wherever possible. You actually found bug in the docs, because that `attr.evolve` used to be a link to the reference but somehow didn't make it through conversion so thanks for that! The rule is kinda that if only one syntax is used, use `attrs.`. But there's a lot of docs…

Why-not-now:
So is there work being made for the `attr` module to be replaced by `attrs` module in the docs? Also glad to help out and all the best to you guys.

Tinche:
Shoot us a PR and we'll gladly merge it in 🙏

Why-not-now:
> Shoot us a PR and we'll gladly merge it in 🙏

There is one paragraph I still do not have a good understanding on:
> You can still have power over the attributes if you pass a dictionary of name: [field()](https://www.attrs.org/en/latest/api.html#attrs.field) mappings and can pass arguments to @attr.s:

PR #1163, opened 2023-07-16
Update examples.md

Why-not-now:
# Summary

Changing order type annotations to fit code
Added attrs.define implementation to slots category

This is to fix the docs problems mentioned in #1162. There is still one line that I think is confusing, line 670:
> You can still have power over the attributes if you pass a dictionary of name: {func}`~attrs.field` mappings and can pass arguments to `@attrs.define`:

# Pull Request Check List

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -498,7 +498,7 @@ If you're the author of a third-party library with *attrs* integration, please s
 
 ## Types
 
-*attrs* also allows you to associate a type with an attribute using either the *type* argument to {func}`attr.ib` and {func}`attr.field` or using {pep}`526`-annotations:
+*attrs* also allows you to associate a type with an attribute using either the *type* argument to using {pep}`526`-annotations or {func}`attrs.ib` and {func}`attrs.field`:
 
 ```{doctest}
 >>> @define
@@ -588,6 +588,11 @@ However it's useful for writing your own validators or serialization frameworks.
 Defining `__slots__` by hand is tedious, in *attrs* it's just a matter of using {func}`attrs.define` or passing `slots=True` to {func}`attr.s`:
 
 ```{doctest}
+>>> @define
+... class Coordinates:
+...     x: int
+...     y: int
+
 >>> import attr
 
 >>> @attr.s(slots=True)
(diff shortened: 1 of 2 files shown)


tests/test_annotations.py
```diff
@@ -587,7 +587,7 @@ def test_self_reference(self, slots):
         @attr.s(slots=slots, auto_attribs=True)
         class A:
             a: "A"
-            b: typing.Optional["A"]  # noqa: will resolve below
```
hynek, line 590: this breaks the CI.

generally speaking, thanks to `yesqa`, you don't have to manually remove `noqa`s

tests/test_annotations.py
```diff
@@ -601,7 +601,7 @@ def test_forward_reference(self, slots):
 
         @attr.s(slots=slots, auto_attribs=True)
         class A:
-            a: typing.List["B"]  # noqa: will resolve below
```
hynek, line 604: same as above

docs/examples.md
```diff
@@ -498,7 +498,7 @@ If you're the author of a third-party library with *attrs* integration, please s
 
 ## Types
 
-*attrs* also allows you to associate a type with an attribute using either the *type* argument to {func}`attr.ib` and {func}`attr.field` or using {pep}`526`-annotations:
+*attrs* also allows you to associate a type with an attribute using either the *type* argument to using {pep}`526`-annotations or {func}`attrs.ib` and {func}`attrs.field`:
```
hynek, line 501:
```suggestion
*attrs* also allows you to associate a type with an attribute using either the *type* argument to using {pep}`526`-annotations or {func}`attrs.field`/{func}`attr.ib`:
```

We can put field first, but we shouldn't erase ib completely, since it helps people to navigate older attrs code bases. Also attrs.ib doesn't exist. :)

docs/examples.md
```diff
@@ -662,7 +667,7 @@ True
 <class 'int'>
 ```
 
-You can still have power over the attributes if you pass a dictionary of name: {func}`~attrs.field` mappings and can pass arguments to `@attr.s`:
+You can still have power over the attributes if you pass a dictionary of name: {func}`~attrs.field` mappings and can pass arguments to `@attrs.define`:
```
hynek, line 670: I tried to remember really hard what I wanted to say with this and I think it's this:

```suggestion
You can still have power over the attributes if you pass a dictionary of name: {func}`~attrs.field` mappings and can pass the same arguments as you can to `@attrs.define`:
```

hynek (changes requested):
hi thanks and sorry for the delay!

if you apply my suggestions, I think this is good to go

tests/test_annotations.py
hynek: (I have just noticed that this appeared to be a rogue yesqa action – still revert pls I've removed yesqa :))

hynek (approved):

PR #1172, opened 2023-08-04
Allow subclassing of AttrsInstance and Protocol at once

### LINKED ISSUE #1133: AttrsInstance is only a Protocol in mypy
`__init__.pyi` defines AttrsInstance as follows:
```
class AttrsInstance(AttrsInstance_, Protocol):
    pass
```

So the following typechecks correctly:
```
import attr
from typing import Protocol

class Foo(attr.AttrsInstance, Protocol):
    def attribute(self) -> int:
        ...


def bar(foo: Foo) -> int:
    return foo.attribute()

@attr.define
class CFoo:
    a: int

    def attribute(self) -> int:
        return self.a

bar(CFoo(15))
```

But if you run it you get:

`TypeError: Protocols can only inherit from other protocols, got <class 'attr.AttrsInstance'>`

Because in `__init__.py` it's defined as:
```
class AttrsInstance:
    pass
```

instead of 
```
class AttrsInstance(Protocol):
    pass
```

hynek:
Fixes #1133

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
           python -Im coverage html --skip-covered --skip-empty
 
           # Report and write to summary.
-          python -Im coverage report | sed 's/^/    /' >> $GITHUB_STEP_SUMMARY
+          python -Im coverage report --format=markdown >> $GITHUB_STEP_SUMMARY
 
           # Report again and fail if under 100%.
           python -Im coverage report --fail-under=100
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ tmp*
 attrs.docset
 attrs.tgz
 Justfile
+t.py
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
           python -Im coverage html --skip-covered --skip-empty
 
           # Report and write to summary.
-          python -Im coverage report | sed 's/^/    /' >> $GITHUB_STEP_SUMMARY
+          python -Im coverage report --format=markdown >> $GITHUB_STEP_SUMMARY
```
euresti, line 96: This isn't related to the change right?

.gitignore
```diff
@@ -19,3 +19,4 @@ tmp*
 attrs.docset
 attrs.tgz
 Justfile
+t.py
```
euresti, line 22: What's this for?

euresti (approved):
Looks good. Thanks!

.github/workflows/ci.yml
hynek: yes, sorry that snook in

.gitignore
hynek: The place I tested your types. ;) Hatch is a bit weird about what files it packages, so I've stopped using global gitignores / .git/info/exclude for projects that I release myself.

PR #1179, opened 2023-08-20
CI: Only attempt publish for upstream

hugovk:
# Summary

When updating my fork with upstream, the CI runs `pypi-package.yml`. 

It's good to run the "Build & verify package" job for forks.

But there's no point running the [Test] PyPI upload jobs, because they'll fail because they don't have the credentials set up -- and we wouldn't want to upload from forks anyway.

image

```
Error: Trusted publishing exchange failure: 
Token request failed: the server refused the request for the following reasons:

* `invalid-publisher`: valid token, but no corresponding publisher (All lookup strategies exhausted)

This generally indicates a trusted publisher configuration error, but could
also indicate an internal error on GitHub or PyPI's part.
```

https://github.com/hugovk/attrs/actions/runs/5916632350/job/16044128992

So let's only attempt to publish for the upstream repo:

image

https://github.com/hugovk/attrs/actions/runs/5916665345

I went for symmetry, but we could probably just do this for `release-test-pypi`, because it's unlikely a fork will trigger `github.event.action == 'published'` for `release-pypi`?

# Pull Request Check List

- [n/a] Added **tests** for changed code.
- [n/a] New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/main/tests/strategies.py).
- [n/a] Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``).
- [n/a] Updated **documentation** for changed code.
- [n/a] Documentation in `.rst` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

### DIFF
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -33,7 +33,7 @@ jobs:
   release-test-pypi:
     name: Publish in-dev package to test.pypi.org
     environment: release-test-pypi
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    if: github.repository_owner == 'python-attrs' && github.event_name == 'push' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     needs: build-package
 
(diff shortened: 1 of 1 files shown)


hynek:
Thanks, symmetry is fine!

hynek (approved):

PR #1187, opened 2023-09-21
Pass args from `__init__` to `__attrs_pre_init__` if requested

### LINKED ISSUE #1102: Pass all args to `__attrs_pre_init__()`
I am subclassing a class that I have no control over, which is written in standard Python.  The base class A has a number of arguments that I want to set.  Infact there are 3 classes.  C (attrs class) is a sublcass of B (attrs class) which is a subclass of A (non-attrs class).

When I instantiate C, I want to pass arguments to B and A.

I implemented class based factory functions (vai `@classmethod`), which works well for a single level inheritance, but I'm not sure how to use the classmethod of B from the classmethod of A.  I figured I'm stuck with duplicating of the B initialisation within the C classmethod and the instantiating class C by passing all the necessary arguments.

All this involves defining a custom `__init__()` method (https://www.attrs.org/en/stable/init.html#custom-init), but it seems to me that could be avoided by using `__attrs_pre_init__()`, but unfortunately it seems no arguments are passed to that (i.e. no `*args` or `**kwargs`).

Is there a reason why arguments are not passed to `attrs_pre_init__()`, and only passed to `__init__()`?

I think it makes sense to pass all arguments to `__attrs_pre_init__()` so the user can have the flexibility to do what is required (or nothing at all).

```python

    def __attrs_pre_init__(self, *args, **kwargs):
        ...
        super().__init__(*args, **kwargs)
        ...
```

I can see why it is less relevant for `__attrs_post_init__()`, but I'm sure someone can think of a use-case where the arguments could be useful (e.g. for arguments that aren't stored as instanace attributes).

Tyrubias:
# Summary

Detect if `__attrs_pre_init__` has arguments besides `self` using `inspect.signature`. If so, pass `__attrs_pre_init__` the same arguments that `__init__` (or `__attrs_init__`) expects. This closes #1102.

This is my first contribution to this project, so I would greatly appreciate any feedback.

# Pull Request Check List

### DIFF
--- a/changelog.d/1187.change.md
+++ b/changelog.d/1187.change.md
@@ -1,2 +1,2 @@
-`__attrs_pre_init__` can now optionally accept the same arguments as `__init__`.
-This allows you to, for example, pass arguments to `super().__init__`.
+If *attrs* detects that `__attrs_pre_init__` accepts more than just `self`, it will call it with the same arguments as `__init__` was called.
+This allows you to, for example, pass arguments to `super().__init__()`.
--- a/changelog.d/1187.change.md
+++ b/changelog.d/1187.change.md
@@ -1,0 +1,2 @@
+`__attrs_pre_init__` can now optionally accept the same arguments as `__init__`.
+This allows you to, for example, pass arguments to `super().__init__`.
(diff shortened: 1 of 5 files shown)


Tinche (approved):

hynek:
Thanks!

PR #1190, opened 2023-10-16
Bump mypy, fix tests

Tinche:
We're running some tests without `strict_optional = False` now, which I think is acceptable (I think that option sucks anyway). It seems to be triggering an issue in typeshed that's too much of a bother to work around.

I've also refactored the dependency tree to be able to use the Mypy version from pyproject.toml in tox, should be better this way, right?


hynek (approved):
Same procedure as every Mypy release 🫡

PR #1192, opened 2023-10-17
docs: explain how we compare

### LINKED ISSUE #1191: Docs for comparison (ordering) operators do not define their semantics
When defining an attrs struct with `order=True`, attrs generates comparison operators automatically. With these, we can ask if one struct instance is (e.g.) less than another. Unfortunately I can't see anywhere the semantics of these operations are explicitly spelled out. It'd be great if the attrs docs had a sentence or two to be explicit on this! 

I'm _guessing_ that the semantics match those of [stdlib dataclasses](https://docs.python.org/3/library/dataclasses.html?highlight=dataclass#dataclasses.dataclass)

> order: If true (the default is False), [`__lt__()`](https://docs.python.org/3/reference/datamodel.html#object.__lt__), [`__le__()`](https://docs.python.org/3/reference/datamodel.html#object.__le__), [`__gt__()`](https://docs.python.org/3/reference/datamodel.html#object.__gt__), and [`__ge__()`](https://docs.python.org/3/reference/datamodel.html#object.__ge__) methods will be generated. These compare the class as if it were a tuple of its fields, in order.

(a lexicographic ordering). As far as I can see from the source, that's how attrs does it too:

https://github.com/python-attrs/attrs/blob/fb9725218640a359fd59c3d48c80f46a1be51d0e/src/attr/_make.py#L1762-L1786

In today's docs, I couldn't see anything about ordering in https://www.attrs.org/en/stable/comparison.html other than

> attrs will also create a full set of ordering methods that are based on the defined fields

and in https://www.attrs.org/en/stable/api-attr.html#attr.s I could only see

> order (Optional[[bool](https://docs.python.org/3/library/functions.html#bool)]) – If True, add `__lt__`, `__le__`, `__gt__`, and `__ge__` methods that behave like _eq_ above and allow instances to be ordered. If `None` (default) mirror value of _eq_.

(For interest, my motivation comes from https://github.com/matrix-org/synapse/pull/16507/files#r1361962916. In that change we converted a bunch of TypedDicts `a, b` to attrs structs `x, y`. We used to assert that `a.items() <= b.items()` (a subset test), but we changed this to `x <= y` (a lexicographic ordering test). These semantics are not the same and so our test wasn't asserting what we thought it was!)

hynek:
fixes #1191

is this clear enough @DMRobertson?

you can see the rendered docs here: https://attrs--1192.org.readthedocs.build/en/1192/comparison.html

### DIFF
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -1,9 +1,15 @@
 # Comparison
 
-By default, two instances of *attrs* classes are equal if all their fields are equal.
+By default, two instances of *attrs* classes are equal if they have the same type and all their fields are equal.
 For that, *attrs* writes `__eq__` and `__ne__` methods for you.
 
-Additionally, if you pass `order=True`[^default], *attrs* will also create a full set of ordering methods that are based on the defined fields: `__le__`, `__lt__`, `__ge__`, and `__gt__`.
+Additionally, if you pass `order=True`, *attrs* will also create a complete set of ordering methods: `__le__`, `__lt__`, `__ge__`, and `__gt__`.
+
+Both for equality and order, *attrs* will:
+
+- Check if the types of the instances you're comparing are equal,
+- if so, create a tuple of all field values for each instance,
+- and finally perform the desired comparison operation on those tuples.
 
 [^default]: That's the default if you use the {func}`attr.s` decorator, but not with {func}`~attrs.define`.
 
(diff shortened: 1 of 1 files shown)


DMRobertson:
Looks like a nice improvement to me, thank you!

PR #1200, opened 2023-11-07
Add support for cached_properties to slotted attrs classes.

diabolo-dan:
# Summary

# Pull Request Check List

- [ ] Do **not** open pull requests from your `main` branch – **use a separate branch**!


hynek:
Since `cached_property` is 3.8+, 3.7 is failing.

In my other projects I've dropped 3.7 long ago, but attrs has still the most downloads on 3.7 so I would like to let it live on a little longer. Would it be a lot of effort to make your gate this feature to 3.8?

Tinche (changes requested):
Generally LGTM but some questions.

Do cached properties make sense on frozen classes? Philosophically and implementation-wise?

Second, since `__getattr__` doesn't chain, will this scenario work:

```
@define
class A:
    @cached_property
    def a(self) -> int:
        return 1

@define
class B(A):
    @cached_property
    def b(self) -> int:
        return 2

B().a
```

We might need to crawl the MRO for the final `__getattr__`?

diabolo-dan:
> Since `cached_property` is 3.8+, 3.7 is failing.
> 
> In my other projects I've dropped 3.7 long ago, but attrs has still the most downloads on 3.7 so I would like to let it live on a little longer. Would it be a lot of effort to make your gate this feature to 3.8?

Ah, I'd forgotten that cached_proeperty was a "new" addition

I've added a commit that restricts the checks to 3.8+, and checked it locally, so it should be working now.

diabolo-dan:
> Generally LGTM but some questions.
> 
> Do cached properties make sense on frozen classes? Philosophically and implementation-wise?
> 

IMHO, they work well together philosophically. A cached property is persistent for the duration of an instance, so it's already "frozen", it's just a question of delaying the computation (potentially indefinitely, if it's never needed).  That makes it roughly comparable to a `lazy val` in scala, which is compatible with immutability from an interface perspective. 

And, I don't see any issues implementation wise -- my implementation already works with frozen classes.


> Second, since `__getattr__` doesn't chain, will this scenario work:
> [...] 
> We might need to crawl the MRO for the final `__getattr__`?

Yeah, good point.  I've played around a bit, and thought I had a working solution, but I think it breaks down on multiple inheritance. I'll try to work something out.

diabolo-dan:
I've pushed a fix to crawl the mro for `__getattr__` (via super).  I've also switched to using `_make_method` as that seems to be the more consistent/preferred approach, though I needed a wrapper to get the closure working properly.

To get the `original_getattr` closure picking up the new class, I put it on the class dictionary.  I suppose if we want to avoid this, we could instead add an explicit update for the closure cells for that function.

hynek:
Thanks for working on this – I'm really excited for this to land! I'll leave the actual reviewing to Tin, but 2 quick notes/questions:

1. You're missing (branch) coverage which made the CI red:

```
Name                Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------
src/attr/_make.py     954      0    455      1    99%   923->929
---------------------------------------------------------------
TOTAL                1638      0    718      1    99%
```

2. I think I know the answer, but just to double-check: this has no performance impact if I don't use `cached_property`s, correct?

diabolo-dan:
>     You're missing (branch) coverage which made the CI red:

Hmmm, It looks like coverage data is only being uploaded for 3.7,3.10, and 3.11.

The branch that coverage is complaining about only gets hit in 3.8/3.9 because `__annotation__` are ~always set in 3.10+, and the code is flagged behind 3.8

I've pushed a commit to maintain comparable behaviour which I believe should resolve the coverage problem in any case.

>    I think I know the answer, but just to double-check: this has no performance impact if I don't use cached_propertys, correct?

There is some extra code at class creation to check for cached_properties:
```
if PY_3_8_PLUS:
    cached_properties = {
        name: cached_property.func
        for name, cached_property in cd.items()
        if isinstance(cached_property, functools.cached_property)
    }
[...]
if cached_properties:
```

Some basic numbers for reference (timing for 1000 attrs.frozen applications):

upstream/main:
```
With 0 methods, took [0.534622973005753, 0.5135590879945084, 0.5537665199954063, 0.543172550998861, 0.5421680030121934]s
With 1 methods, took [0.5519589369941968, 0.552411498996662, 0.5460843400214799, 0.5372608639881946, 0.5253033999761101]s
With 10 methods, took [0.5476001699862536, 0.5778162179922219, 0.5552783849998377, 0.5771896800142713, 0.5804753249976784]s
With 100 methods, took [0.6648160929908045, 0.6790099880017806, 0.655211570992833, 1.1942919419961981, 0.843360381986713]s
With 1000 methods, took [1.327303853991907, 1.2969695710053202, 1.3404755429946817, 1.2886629370041192, 1.3148314789868891]s
With 10000 methods, took [8.27522504600347, 9.543007285014028, 9.15497858900926, 8.78463426901726, 9.318197846994735]s
```

vs

New changes
```
With 0 methods, took [0.5188219360134099, 0.5184214509790763, 0.5280631629866548, 0.5325713030179031, 0.5457004369818605]s
With 1 methods, took [0.5420163899834733, 0.545073927991325, 0.5370112269883975, 0.5418282540049404, 0.5468027850147337]s
With 10 methods, took [0.5930820080102421, 0.5426902859762777, 0.552412015007576, 0.5950264070124831, 0.6470643689972349]s
With 100 methods, took [0.9766806430125143, 0.6893237390031572, 0.6373719449911732, 0.6113516429904848, 0.6429122459958307]s
With 1000 methods, took [1.4384041439916473, 1.735367571003735, 1.6085309459886048, 1.7646593420067802, 1.3692002349998802]s
With 10000 methods, took [9.481361449987162, 9.846626015001675, 9.705778863019077, 10.955383850989165, 11.169956739002373]s
```

So it does start to become noticeable at high method counts (or presumably other non-attribute fields), but for that to be meaningful it means dynamically creating multiple classes with

hynek:
> > ```
> > You're missing (branch) coverage which made the CI red:
> > ```
> 
> Hmmm, It looks like coverage data is only being uploaded for 3.7,3.10, and 3.11.
> 
> The branch that coverage is complaining about only gets hit in 3.8/3.9 because `__annotation__` are ~always set in 3.10+, and the code is flagged behind 3.8
> 
> I've pushed a commit to maintain comparable behaviour which I believe should resolve the coverage problem in any case.

You can absolutely change coverage to be measured on 3.9 too btw – no need to do any crazy hacks around that (disclaimer: I haven't looked at what you've done – just stating).

> So it does start to become noticeable at high method counts (or presumably other non-attribute fields), but for that to be meaningful it means dynamically creating many classes with many methods, which doesn't seem like a typical use case.
> 
> As far as object creation or usage, changes to the class only occur when there are `cached_property`s, so it should be completely unimpacted.

Yeah this is absolutely OK, thank you for humoring me.

src/attr/_make.py
```diff
@@ -597,6 +599,48 @@ def _transform_attrs(
...
+        "    __class__ = _cls",
+        "    def __getattr__(self, item):",
+        "         func = cached_properties.get(item)",
+        "         if func is not None:",
+        "              result = func(self)",
+        "              _setter = _cached_setattr_get(self)",
+        "              _setter(item, result)",
+        "              return result",
+        "         if '__attrs_original_getattr__' in vars(__class__):",
+        "              return __class__.__attrs_original_getattr__(self, item)",
```
Tinche, line 619: Do you have to do this check here? Can this be checked at class construction time and passed as a parameter to `_make_cached_property_getattr`? Might make the `__class__` global unnecessary too.

src/attr/_make.py
```diff
@@ -597,6 +599,48 @@ def _transform_attrs(
...
+        "         if hasattr(super(), '__getattr__'):",
```
Tinche, line 620: Same here.

src/attr/_make.py
```diff
@@ -597,6 +599,48 @@ def _transform_attrs(
...
+def _make_cached_property_getattr(
+    cached_properties,
+    cls,
+):
+    lines = [
+        # Wrapped to get `__class__` into closure cell for super()
+        # (It will be replaced with the newly constructed class after construction).
+        "def wrapper(_cls, cached_properties, _cached_setattr_get):",
+        "    __class__ = _cls",
+        "    def __getattr__(self, item):",
```
Tinche, line 611: You can make `cached_properties` be a local instead of a global inside of `__getattr__` by making it an optional parameter with `cached_properties` as the default value.

It's an old trick for speeding up variable access.

Tinche (changes requested):

src/attr/_make.py
diabolo-dan: I'm presuming this applies to the other non-local references inside `__getattr__`, so I've added them all.  I've also removed them from `wrapper`, as that seems unnecessary if they're then being localised.

src/attr/_make.py
diabolo-dan: I've moved the check out to construction, and passed in the `original_getattr` to avoid the class lookup.

`__class__` is still needed in the closure for the `super()` call to work.

I've removed  `__attrs_original_getattr__` on the class, as a separate MR, because it needed some other changes  for it goes through the `__class__` closure reference updating, which has it's own trade-off.

diabolo-dan:
> You can absolutely change coverage to be measured on 3.9 too btw – no need to do any crazy hacks around that (disclaimer: I haven't looked at what you've done – just stating).
> 

Fair enough.  I don't think it's a hack, just a slightly different approach, but the code change is in it's own commit, so if the prior option is preferred, it should be easy to revert (in which case I'll look into the github ci settings).

src/attr/_make.py
diabolo-dan: Sorry, should have been clearer, I was thinking of something along the lines of:
```
@attrs.frozen
class A:
    a: int
    @cached_property
    def b(self) -> int:
        return self.a + 1

class Mixin:
    __slots__ = ()
    def __getattr__(self, item: str) -> int:
        if item ==  'c':
            return 1
        return super().__getattr__(item)

@attrs.frozen
class C(A, Mixin):
    ...

C(1).c
```

the `__getattr__` on `A`  needs to call `super` so that it passes to the `__getattr__` on `Mixin`, but can't know about it's existence at construction time.

Tinche: Ah I see. I suspected multiple inheritance might be the thing complicating our life here ;)

Tinche (approved):

PR #1201, opened 2023-11-10
Remove work-arounds for setting cell_contents

eltoder:
# Summary

Simple assignment to cell_contents works starting from Python 3.7[1]. This is also the minimum supported version of Python according to pyproject.toml. It follows that we don't need complex work-arounds that were needed in the past.

Also tested on pypy 3.8.

[1] https://bugs.python.org/issue30486

### DIFF
--- a/src/attr/_compat.py
+++ b/src/attr/_compat.py
@@ -4,8 +4,6 @@ import inspect
 import platform
 import sys
 import threading
-import types
-import warnings
 
 from collections.abc import Mapping, Sequence  # noqa: F401
 from typing import _GenericAlias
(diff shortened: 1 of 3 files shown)


hynek (approved):
omg we're getting rid of some of the gnarliest code!? thank you so much! I wonder why we thought the other code is 3.8+

eltoder:
@hynek Possibly someone noticed the change mentioned on this page: https://docs.python.org/3.8/whatsnew/changelog.html and assumed it was new in 3.8. However, this is a cumulative changelog, and the change is under "Python 3.7.0 alpha 1" :-)

PR #1202, opened 2023-11-10
Fix mypy and pyright tests in CI

eltoder:
# Summary

Some error messages changed in mypy 1.7.0 and pyright 1.1.335.

Also, test_mypy sometimes fails with xdist (-n auto). A failure I can
reliably reproduce locally is due to changes in type variables numbering.
There may be other issues.


hynek (approved):
thanks!

PR #1234, opened 2024-02-02
Tweak stubs for better VS Code import

Tinche:
I moved the definitions of `define`, `mutable`, `frozen` and `field` to `attrs`, and made `attr` import them.

Screenshot 2024-02-03 at 00 24 34

Then I modernized the stubs since I couldn't help myself. We can use modern typing stuff in stubs.

This was motivated by https://github.com/microsoft/pylance-release/issues/5388.


Tinche:
After this release goes out, let's move the others too (`Factory`, `has`, `fields`, `AttrsInstance`, etc).

hynek:
Are you positive the new syntax doesn’t break something somewhere? I’ve got this vague memory of trying it myself and having to abort. We also kinda stopped running Mypy on 3.7. 🤔

changelog.d/1234.change.md
```diff
@@ -0,0 +1,2 @@
+Change how our type stubs are organized to optimize VS Code quick-fix import suggestions.
+Modernize type stubs.
```
hynek, line 2:
```suggestion
Our type stubs now use modern type notation and are organized such that VS Code's quick-fix prefers the `attrs` namespace.
```

sOmEoNe forgot how to write changelogs eh ಠ_ಠ

Tinche:
My main work codebase is 3.8 and Mypy 1.8.0, no problems here. Best I can do :( Can you run it over your stuff?

PEP 604 (new union syntax) is from 2019 so typecheckers have had some time to catch up.

> sOmEoNe forgot how to write changelogs eh ಠ_ಠ

Are we writing snippets for devs to scan through or novels for when you're on the toilet? 🤔

hynek:
> My main work codebase is 3.8 and Mypy 1.8.0, no problems here. 

Thoughts and prayers!

> Best I can do :( Can you run it over your stuff?

I’m all 3.12. 😇 

> PEP 604 (new union syntax) is from 2019 so typecheckers have had some time to catch up.

Right, we’re benefiting once again from our external typing stubs, aren’t we?

> > sOmEoNe forgot how to write changelogs eh ಠ_ಠ
> 
> Are we writing snippets for devs to scan through or novels for when you're on the toilet? 🤔

Yes!

Tinche:
> Right, we’re benefiting once again from our external typing stubs, aren’t we?

Yeah. We would also benefit if we were using `from __future__ import annotations`, although our annotations wouldn't be introspectable at runtime. (But stubs aren't introspectable either, so...)

hynek:
There's a few cases where that doesn't work as well tho, because sometimes types aren't used just as annotations. Stuff like `T = X | list[Y]` -- so that's an upside of pyi I guess.

hynek (approved):

PR #1248, opened 2024-02-27
Fix ruff config namespace

econchick:
`[tool.lint....]` isn't anything 😄 Should be `[tool.ruff.lint]` according to ruff's [docs](https://docs.astral.sh/ruff/configuration/). Hope this doesn't break anything 🤞🏻


hynek (approved):
Yay thanks!

PR #1250, opened 2024-02-29
Fixes tox command for watching docs in CONTRIBUTING.md

mikejturner:
# Summary

The tox command in CONTRIBUTING.md referred to a tox environment `docs-serve` that no longer exists in the tox.ini.

The tox command in CONTRIBUTING.md was modified to refer to `docs-watch` instead.


hynek (approved):
Good catch, thanks!

PR #1260, opened 2024-03-14
Fix broken link from attr API reference back to attrs

koenigdavidmj:
# Summary

`attr.fields_dict` was a link back to itself. The intention, like all the other nearby aliases, was to refer back to the version in `attrs` for the full story.

# Pull Request Check List

Omitted for trivial documentation change.


hynek (approved):
Thanks!

PR #1266, opened 2024-03-16
Use tox-uv in CI

hynek:
Let's see if it makes a difference.


hynek:
Not much of a difference in total, but some differences individually.

PR #1267, opened 2024-03-17
converters: allow wrapping and passing self and fields

### LINKED ISSUE #709: Allow three-argument converters (like validators/on_setattr)
I'd like to move the discussion from [converter decorator PR](https://github.com/python-attrs/attrs/pull/404) to this issue.

I think converters are semantically closer to `on_setattr` and `validator` than `default`. E.g. `attr.ib(converter=...)` allows you to pass a list of callables and pipes them automatically - exactly like `attr.ib(validator=...)` and `attr.ib(on_setattr=...)` and there is `attr.converters` module, like `attr.setters` and `attr.validators`. 

If we allow passing half-initialized self to converter, why don't allow full-form converters, `converter(self, attr, value)` to make them the same as `on_setattr`, but for initialization? 

To support one, two and three-argument converters there should be either inspect-magic (in py2 - getargspec, in py3 - signature) or mandatory `Converter` wrapper for two and three-argument converters.

hynek:
This allows to wrap converters into classes and pass them the instance that is being initialized and the field for which the value is converted.

Things open:

- [x] narrative docs in `init.md`
- [ ] I'm not sure how to approach the typing side, since every attempt failed. I suspect due to a lack of Mypy plugin support. Can we somehow cheat our way here?

fixes #404, fixes #709 (and probably more)

It will enable #240 if anyone still cares about it.


Tinche:
Here are the changes for the next step.

We need to make `Converter` generic over the input and output types:

```python
In = typing.TypeVar("In")
Out = typing.TypeVar("Out")


class Converter(typing.Generic[In, Out]):
    """
    Stores a converter callable.

    Allows for the wrapped converter to take additional arguments. The
    arguments are passed in the order they are documented.

    :param Callable converter: A callable that converts a value.
    :param bool takes_self: Pass the partially initialized instance that is
        being initialized as a positional argument. (default: `False`)
    :param bool takes_field: Pass the field definition (an `Attribute`) into
        the converter as a positional argument. (default: `False`)

    .. versionadded:: 24.1.0
    """

    __slots__ = (
        "converter",
        "takes_self",
        "takes_field",
        "_first_param_type",
        "_global_name",
    )

    @typing.overload
    def __init__(
        self,
        converter: typing.Callable[[In, AttrsInstance, typing.Any], Out],
        *,
        takes_self: typing.Literal[True] = ...,
        takes_field: typing.Literal[True] = ...,
    ) -> None: ...

    @typing.overload
    def __init__(
        self,
        converter: typing.Callable[[In, typing.Any], Out],
        *,
        takes_field: typing.Literal[True] = ...,
    ) -> None: ...

    @typing.overload
    def __init__(
        self,
        converter: typing.Callable[[In, AttrsInstance], Out],
        *,
        takes_self: typing.Literal[True] = ...,
    ) -> None: ...

    @typing.overload
    def __init__(self, converter: typing.Callable[[In], Out]) -> None: ...

    def __init__(self, converter, *, takes_self=False, takes_field=False):
        self.converter = converter
        self.takes_self = takes_self
        self.takes_field = takes_field

        self._first_param_type = _AnnotationExtractor(
            converter
        ).get_first_param_type()
```

Then, we change the type stubs for `field` to take:

```
converter: _ConverterType | Converter[Any, _T] | None = ...,
```

I don't think this matters much since Mypy patches it.

Optional additional step: we finally make `attrs.Attribute` generic at runtime. That will let us have better typing here (and is more correct in general).

Then, we implement support for this in the Mypy plugin. This should be pretty straightforward since `Converter` is nicely generic.

hynek:
@Tinche thanks for fixing Mypy tests – is there anything else to do around them or is this the best we can do for now?

Tinche (approved):

Yedfree (approved):

Issue #1193, opened 2023-10-18
Use field *repr* callable to dynamically define repr based on instance value


RNKuhns:
@hynek thanks for the work on `attrs`!

Ideally, I'd only like a field to appear in a class' *repr* if that field is set to a non-default value. An example of this approach is scikit-learn's approach to printing class representations.

I'm trying to understand if this is already possible currently in `attrs`. I know that `attrs` accepts a callable for the *repr* parameter of an `attrs.field`.  But reading through the documentation, it is not clear what context (if any) is available to the *repr*. If the class' default values and the actual value of the parameter are as context to the callable, it would be relatively straightforward to craft a returned string that accomplished this. 

Can you provide some info on what's available to the callable so I can understand if this is possible?

hynek:
Unfortunately, what you want it is currently not possible. The repr function only receives the value that's being repr-ed and the return value is added to the repr unconditionally:

https://github.com/python-attrs/attrs/blob/c849440ac93d5f33578e9e0d1b1d8f26fbd590ae/src/attr/_make.py#L1874-L1878

RNKuhns:
> Unfortunately, what you want it is currently not possible. The repr function only receives the value that's being repr-ed and the return value is added to the repr unconditionally:
> 
> 
> 
> https://github.com/python-attrs/attrs/blob/c849440ac93d5f33578e9e0d1b1d8f26fbd590ae/src/attr/_make.py#L1874-L1878

@hynek Is this something you'd be interesting in having added. If so, I can look into the codebase and see if it falls into something I could make a PR for.

hynek:
If someone came up with a good PR, I would merge it.

But I would _strongly_ suggest to discuss your implementation approach with us first, before you put work into this that doesn't have a chance to be merged!

PR #1276, opened 2024-04-11
Provide option in attrs.define to allow users to exclude parameters set to default value from `repr`

RNKuhns:
# Summary

This PR provides functionality in line with #1193.

It adds a new parameter to attrs.define to allow users to toggle on/off the ability to create classes that dynamically generate their repr to include only parameters set to values other than their default. The added parameter is set to a default value that maintains the existing functionality (always have static repr).

```python
@attrs.define(kw_only=True, only_non_default_attr_in_repr=True)
class SomeClass:

    something: int | None = attrs.field(default=None, repr=True)
    something_else: bool = attrs.field(default=False, repr=True)
    another: float | None = attrs.field(default=11.0, init=False, repr=False)


some_class = SomeClass(something=7)
SomeClass(something=7)


# If we wanted to exclude the something param from the repr in the field def we still can
@attrs.define(kw_only=True, only_non_default_attr_in_repr=True)
class SomeClass:

    something: int | None = attrs.field(default=None, repr=False)
    something_else: bool = attrs.field(default=False, repr=True)
    another: float | None = attrs.field(default=11.0, init=False, repr=False)


some_class = SomeClass(something=7)
SomeClass()


# The default is equivalent of only_non_default_attr_in_repr=False, so existing behavior is default
@attrs.define(kw_only=True)
class SomeClass:

    something: int | None = attrs.field(default=None)
    something_else: bool = attrs.field(default=False)
    another: float | None = attrs.field(default=11.0, init=False, repr=False)


some_class = SomeClass(something=7)
SomeClass(something=7, something_else=False)


# The use of repr in a field works just like it did before so we can exclude individual attributes
# from the repr (or pass custom callable)
@attrs.define(kw_only=True)
class SomeClass:

    something: int | None = attrs.field(default=None, repr=False)
    something_else: bool = attrs.field(default=False)
    another: float | None = attrs.field(default=11.0, init=False)


some_class = SomeClass(something=7)
SomeClass(something_else=False, another=11.0)
```

@hynek I'll look into creating some test cases for this so the CI passes. I ran the test suite locally and didn't have any other failures, so should hopefully just have to add tests for coverage of this functionality. But I'd appreciate some feedback on whether this approach makes sense to you before I finish up with that.

Note there are some nuisance edits to the `_make.py` that ``ruff`` made when I saved the file in my setup, but these look to be minor reformatting.

# Pull Request Check List

- [X] Do **not** open pull requests from your `main` branch – **use a separate branch**!
    - [ X] New functions/classes have to be added to `docs/api.rst` by hand.
    - [ X] Changes to the signature of `@attr.s()` have to be added by hand too.
    - [ X] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).
- [ X] Documentation in `.rst` and `.md` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1014,9 +1014,14 @@ class _ClassBuilder:
                         cell.cell_contents = cls
         return cls
 
-    def add_repr(self, ns):
+    def add_repr(self, ns, only_non_default_attr_in_repr=False):
         self._cls_dict["__repr__"] = self._add_method_dunders(
-            _make_repr(self._attrs, ns, self._cls)
+            _make_repr(
+                self._attrs,
+                ns,
+                self._cls,
+                only_non_default_attr_in_repr=only_non_default_attr_in_repr,
+            )
         )
         return self
 
@@ -1332,6 +1337,7 @@ def attrs(
     field_transformer=None,
     match_args=True,
     unsafe_hash=None,
+    only_non_default_attr_in_repr=False,
 ):
     r"""
     A class decorator that adds :term:`dunder methods` according to the
@@ -1562,6 +1568,12 @@ def attrs(
         non-keyword-only ``__init__`` parameter names on Python 3.10 and later.
         Ignored on older Python versions.
 
+    :param bool only_non_default_attr_in_repr:
+        If `False` (default), then the usual ``attrs`` repr is created. If `True`
+        then only parameters set to their non-default values will be printed.
+        This means when this is set to `True` the repr output is dynamic based
+        on the state of the class.
+
     .. versionadded:: 16.0.0 *slots*
     .. versionadded:: 16.1.0 *frozen*
     .. versionadded:: 16.3.0 *str*
@@ -1655,7 +1667,10 @@ def attrs(
         if _determine_whether_to_implement(
             cls, repr, auto_detect, ("__repr__",)
         ):
-            builder.add_repr(repr_ns)
+            builder.add_repr(
+                repr_ns,
+                only_non_default_attr_in_repr=only_non_default_attr_in_repr,
+            )
         if str is True:
             builder.add_str()
 
@@ -1961,7 +1976,7 @@ def _add_eq(cls, attrs=None):
     return cls
 
 
-def _make_repr(attrs, ns, cls):
+def _make_repr(attrs, ns, cls, only_non_default_attr_in_repr=False):
     unique_filename = _generate_unique_filename(cls, "repr")
     # Figure out which attributes to include, and which function to use to
     # format them. The a.repr value can be either bool or a custom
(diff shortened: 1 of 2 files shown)


RNKuhns:
Note that I'll also add examples to the docs before this is finalized/merged if the concept moves forward.

RNKuhns:
@hynek I know you are probably busy, but I just wanted to check in to see if you had a chance to take a look and provide feedback.

hynek:
yeah sorry I'm swamped right now, as you can tell in my own PR #1267 that hasn't moved in a while. I don't have the headspace for bigger changes right now, but it won't get lost as long as you leave it open. I hope to be able to clean up the trackers before leaving for PyCon US (2024 – just in case ;))

RNKuhns:
> yeah sorry I'm swamped right now, as you can tell in my own PR #1267 that hasn't moved in a while. I don't have the headspace for bigger changes right now, but it won't get lost as long as you leave it open. I hope to be able to clean up the trackers before leaving for PyCon US (2024 – just in case ;))

No problem at all! I'll leave it open and add some of the finishing touches (tests cases, docs) as I have time on the next week or two.

RNKuhns:
@hynek I've updated this to handle default factories and include a description of the functionality in the `attrs by examples` section of the documentation.

Let me know when you have a chance to take a look.

hynek:
Hey sorry once again, for all the delays. :( Before starting investigating, do you have any idea why this PR slows down class creation (**not** instantiation) by almost 30%? That's quite a lot and sadly a rather critical benchmark in the whole space of creating classes. :|

src/attr/_make.py
```diff
@@ -2062,7 +2081,27 @@ def _make_repr(attrs, ns, cls):
...
+        "     attr_frags = []",
+        "     for a in getattr(self, '__attrs_attrs__', []):",
+        "         value = getattr(self, a.name, NOTHING)",
+        "         if hasattr(a.default, 'factory') and callable(a.default.factory):",
+        "             if a.default.takes_self:",
+        "                 default_ = a.default.factory(self)",
+        "             else:",
+        "                 default_ = a.default.factory()",
+        "         else:",
+        "             default_ = a.default",
```
mara004, line 2095: * I think you could generate the defaults only once on first repr and cache them for subsequent runs.
* IIRC, attrs applies converters on defaults, so shouldn't this do, too?
* Calling factories unconditionally is a bit problematic for things like `factory=lambda: next(generator)` because it would advance the caller's generator without an actual instantiation. Overall, it might be better to keep track of the defaults produced on construction and compare against that?

FWIW, I had written a [caller-side wrapper to hide defaults](https://gist.github.com/mara004/5f02da5f3a09f9206c0755b5bcf6f105). (The way it works around the factory problem is by letting the caller mark factories as dynamic if they don't want them called in repr.)

Disclaimer: I'm not an attrs maintainer, so my views may not matter here.

docs/examples.md
```diff
@@ -381,6 +381,80 @@ C(x=1, y=2, z=[])
...
+>>> C(x=2, y=3)
+C(x=2, y=3)
+```
+
+Other attrs repr features, including turning the repr off for a field or
+providing a custom callable for the repr of a field, will work as usual, when the
+field's value is not the default.
+
+But the field is still excluded from the repr when it is set to the default value
+because `only_non_default_attr_in_repr` overrides `repr=True` or the repr being a
```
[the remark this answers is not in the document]
mara004: ah, no, because then this wouldn't be available if a custom repr is given. so perhaps add a new argument `always_repr` or something?

docs/examples.md
```diff
@@ -381,6 +381,80 @@ C(x=1, y=2, z=[])
...
+```{doctest}
+>>> @define(only_non_default_attr_in_repr=True)
+... class C:
+...     x: int = field(default=1, converter=lambda value: value + 0.5)
+...     z: int = field(default=12, converter=int)
+>>> C(x=0.5, z="12")
+C()
+>>> C(x=1)
+C(x=1.5)
+```
```
mara004, line 456: I don't think this is correct (see the other comment). The real default for field `x` would be `1.5` here:
```pycon
>>> import attrs
>>>
>>> @attrs.define()
... class Test:
...     x = attrs.field(default=1, converter=lambda v: v + 0.5)
... 
>>> test = Test()
>>> test.x
1.5
```

RNKuhns:
@hynek I do plan to circle back on this. I've just got a couple things going on at the moment.

RNKuhns:
> Hey sorry once again, for all the delays. :( Before starting investigating, do you have any idea why this PR slows down class creation (**not** instantiation) by almost 30%? That's quite a lot and sadly a rather critical benchmark in the whole space of creating classes. :|

Not a problem -- I'll take a look at this and work through what is driving those performance degradations.

hynek:
hey this is now extremely stale, so I'm closing it. feel free to open a new shot anytime!

Issue #1288, opened 2024-06-01
`__getattr__` in child class gets called when mixed with `cached_property`


dalejung:
```python
from functools import cached_property
from attrs import define

@define
class Bob:
    @cached_property
    def howdy(self):
        return 3

class Sup(Bob):
    def __getattr__(self, name):
        raise AttributeError(name)

b = Sup()

# reaches __getattr__ where I raise AttributeError
b.howdy

```

This previously did not error. I would expect `__getattr__` to not be called since the attribute exists.

hynek:
I believe his is fixed in #1253. I wanted to push a release before PyCon/travel but sadly #1267 blocked me. :(

dalejung:
I just checked against latest master and it still exists. It looks like 23.2.0 is when the error started, 23.1.0 works.

hynek:
@dlax any insights? 🤔 I can’t test anything for weeks to come.

dlax:
The fix from #1253 does not cover this inheritance case, indeed. I'll see if this can be fixed.

dlax:
#1289 fixes a slightly different version of the original problem, namely when the child class is also a (slotted) attr class, thus from the original example:
```python
@define
class Sup(Bob):
    ...
```

I'm not sure if the original problem may be resolved. Or at least, not in the same manner as we have no mean to inspect the plain subclass.

dlax:
https://github.com/python-attrs/attrs/pull/1289 is another attempt, which resolves both situations mentioned above.

diabolo-dan:
Just noticed this issue, and thought I could give my insight for what it's worth.


As noted the problem listed is unfixable with the approach, because `__getattr__` on the attrs class is never getting called.  This is to some extent an inherent problem with `__getattr__` from multiple sources, and there's a fairly reasonable argument that `__getattr__` in the derived class should be calling `super().__getattr__(name)`, as a general practice, not just in this case, because you don't necessarily know how a class will be further sub-classed.

```
from functools import cached_property
import attrs



@attrs.define
class Bob:
    @cached_property
    def howdy(self):
        return 3

class Sup(Bob):
    def __getattr__(self, name):
        try:
            return super().__getattr__(name)
        except AttributeError:
            raise AttributeError(name)

b = Sup()
print(b.howdy)
```

Runs with no errors (and prints 3).

That said, obviously it's surprising to encounter this error, and it would be better if it could be avoided.

[...]

PR #1291, opened 2024-06-03
Define __getattribute__() instead of __getattr__() on slotted classes with cached properties

dlax:
# Summary

Fix https://github.com/python-attrs/attrs/issues/1288

Method `__getattribute__()` is [documented][1] as the "way to to actually get total control over attribute access" so we change the implementation of slotted classes with cached properties by defining a `__getattribute__()` method instead of `__getattr__()` previously.

[1]: https://docs.python.org/3/reference/datamodel.html#customizing-attribute-access

Just changing that preserves the current behaviour, according to the test suite, but also makes sub-classing work better, e.g. when the subclass is not an attr-class and also defines a custom `__getattr__()` as evidenced in added test from #1288.

In tests, we replace most custom `__getattr__()` definitions by equivalent `__getattribute__()` ones, except in regression tests where `__getattr__()` is explicitly involved. Also, in test_slots_with_multiple_cached_property_subclasses_works(), we replace the `if hasattr(super(), "__getattr__"):` by a `try:`/`except AttributeError:` as using `hasattr(..., "__getattribute__")` would be meaningless since `__getattribute__()` is always defined.

# Pull Request Check List

### DIFF
--- a/changelog.d/1291.change.md
+++ b/changelog.d/1291.change.md
@@ -1,0 +1,2 @@
+Rework handling of `functools.cached_property` in slotted classes by defining a `__getattribute__()` method instead of `__getattr__()` previously.
+While this is closer to the guidance for [customizing attribute access](https://docs.python.org/3/reference/datamodel.html#customizing-attribute-access), this fixes inheritance resolution of cached properties in slotted classes when subclasses are not attr-decorated but define a custom `__getattr__()` method.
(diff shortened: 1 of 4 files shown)


hynek:
Closing due to https://github.com/python-attrs/attrs/issues/1288#issuecomment-2227820600 – thank you very much for the work though! Sometimes it takes an experiment to find out.

PR #1299, opened 2024-07-11
Add Codspeed

hynek:
Let's see if we can get a better handle on our performance development.


hynek:
Let's hope merging will create the missing baseline

Yedfree (approved):

PR #1304, opened 2024-07-17
Improve sponsor handling

hynek:
this allows to serve the logos from fixed tags

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,6 +11,10 @@ build:
     post_checkout:
       - git fetch --tags
 
+    # Replace versions in sponsor URLs
+    pre_build:
+      - cog -rP docs/index.md
+
 python:
   install:
     - method: pip
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,6 +4,28 @@ Release **{sub-ref}`release`**  ([What's new?](changelog.md))
 
 ```{include} ../README.md
 :start-after: 'teaser-begin -->'
+:end-before: '<!-- sponsor-break-begin'
+```
+
+<!-- [[[cog
+# This is mainly called from RTD's pre_build job!
+
+import pathlib, tomllib, importlib.metadata
+
+if "dev" in (version := importlib.metadata.version("attrs")):
+    version = "latest"
+
+for sponsor in tomllib.loads(pathlib.Path("pyproject.toml").read_text())["tool"]["sponcon"]["sponsors"]:
+      print(f'<a href="{sponsor["url"]}"><img title="{sponsor["title"]}" src="/en/{version}/_static/sponsors/{sponsor["img"]}" width="200" height="60" /></a>')
+]]] -->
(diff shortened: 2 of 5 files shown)


.readthedocs.yaml
```diff
@@ -11,6 +11,10 @@ build:
     post_checkout:
       - git fetch --tags
 
+    # Replace versions in sponsor URLs
+    pre_build:
+      - cog -rP docs/index.md
```
webknjaz, line 16: @hynek would you be open to sticking `tox` into the RTD config and making it as close as possible to the CI and local runs? I did it in a few places and liked how it simplifies the setup and makes it more predictable.

hynek: I would, it would just complicate the RTD setup :|

docs/index.md
```diff
@@ -4,6 +4,28 @@ Release **{sub-ref}`release`**  ([What's new?](changelog.md))
 
 ```{include} ../README.md
 :start-after: 'teaser-begin -->'
+:end-before: '<!-- sponsor-break-begin'
+```
+
+<!-- [[[cog
```
webknjaz, line 10: I think, this could be turned into an in-tree Sphinx extension and could be better integrated with tox and other things that call Sphinx. Just an idea for the future.

hynek: I did some research and it wasn't trivial, so I took the safe route for now. The problem is that I need cog for the README anyways :|

.readthedocs.yaml
hynek: fixed!

.readthedocs.yaml
```diff
@@ -11,6 +11,11 @@ build:
     post_checkout:
       - git fetch --tags
 
+    # Replace versions in sponsor URLs
+    pre_build:
+      - python -Im pip install tox-uv
+      - python -Im tox run -e docs-sponsors
```
webknjaz, line 17: Here's a few examples w/ tox/nox/scripts I got, with more involved integrations, FYI:

- https://github.com/ansible/awx/blob/587f0ec/.readthedocs.yaml
- https://github.com/ansible/docsite/blob/3ae5a22/.readthedocs.yaml
- https://github.com/ansible/jinja-docsite/blob/00290a6/.readthedocs.yaml
- https://github.com/ansible/ansible-documentation/blob/780b4d9/.readthedocs.yaml

Not something to implement in this PR, but maybe at some point. One of the things to do in CIs would be running tox twice. First, with `--notest` and then, with `--skip-pkg-install`. This makes sure that two stages are represented as separate inspectable entries in the log.

docs/index.md
webknjaz: Sure. What I meant is that a sphinx extension could still call cog or whatever. It'd just be less copy-pasted things. But that's not for now. I know how to write an extension, but I don't suggest that this is something to be done right in this PR.

webknjaz (approved):

PR #1305, opened 2024-07-17
Don't run CodSpeed on merge_group events

hynek:
It's not supported.


Tinche (approved):

webknjaz:
@hynek sorry, I was AFK but looks reasonable. *Technically*, if this was like super desired, the workflow could be migrated to the `push` event since Merge Queues also push to a temporary branch, so two events are emitted and some services don't really care enough about separating them.

PR #1307, opened 2024-07-19
Switch to build-and-inspect-python-package

hynek:
Use the occasion to factor out PyPy jobs which simplifies everything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,6 +234,7 @@ jobs:
 
     needs:
       - coverage
+      - tests-pypy
       - docs
       - install-dev
       - pyright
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -51,47 +63,85 @@ jobs:
...
+  tests-pypy:
+    name: Tests on ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    needs: build-package
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - pypy-3.8
```
webknjaz, line 106: Isn't 3.8 unsupported by PyPy already? Pytest was hitting terrible crashes related to a bug in their GC quite often. So we ended up upgrading to 3.9 that's the lowest supported today and has a fix for that obscure bug.

hynek: If that’s true, I’m more than happy to drop it. I’ve just kept what was before

hynek: “We currently support python 3.10, 3.9, and 2.7.” yay

PR #1308, opened 2024-07-19
Switch docstrings to Napoleon style

hynek:
This makes them much more readable while editing.

c.f. https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

This has been a pet peeve of mine lately, since the default style is super crampy and noisy. This takes more space, but is a joy to read.

Rendered:

- https://attrs--1308.org.readthedocs.build/en/1308/api.html
- https://attrs--1308.org.readthedocs.build/en/1308/api-attr.html

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -164,16 +164,23 @@ You will find the built documentation in `docs/_build/html`.
 
 - If you add a new feature, demonstrate its awesomeness on the [examples page](https://github.com/python-attrs/attrs/blob/main/docs/examples.md)!
 
-- For docstrings, we follow [PEP 257](https://peps.python.org/pep-0257/) and use the `"""`-on-separate-lines style:
+- For docstrings, we follow [PEP 257](https://peps.python.org/pep-0257/), use the `"""`-on-separate-lines style, and [Napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html)-style API documentation:
 
   ```python
-  def func(x):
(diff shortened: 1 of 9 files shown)


webknjaz:
Interesting… My preference is the opposite, typically. By the way, some linters like darglint often take a noticeable performance hit when the alternative docstring arg styles are in use. FYI.

.github/CONTRIBUTING.md
```diff
@@ -164,16 +164,23 @@ You will find the built documentation in `docs/_build/html`.
...
+      Args:
+          x: A very important argument.
 
-      :rtype: str
+          y:
+            Another very important argument, but its description is so long
+            that it doesn't fit on one line. So, we start the whole block on a
+            fresh new line to keep the block together.
+
+      Returns:
```
webknjaz, line 182: It'd be useful to show that raises/yields exist too.

sscherfke: Instead of `Returns` you can also use `Return`, which makes imho a bit more sense given the fact that docstrings are written in imperative (i.e. "Do something, return something"). 

But this is just a minor nitpick and afaik it is both rendered the same way.

webknjaz: Only the summary line needs to be in imperative mood, though. The rest is behavior explanation.
Besides, the rendered HTML will still have “Returns” so why not match it.

hynek: Yeah, to me it’s different too. The main body says what will happen, the rest is rather “descriptive”.

PR #1310, opened 2024-07-19
Speed up __eq__

Tinche:
Create the `__eq__` as a chain of ANDed comparisons, instead of a tuple.

On my computer, a class with two attributes now compares almost twice as fast: 141 ns vs 222 ns.

This will introduce a very slight incompatibility which I think is worth it for us (and dataclasses already introduced): attributes that compare equal by identity but not by value (= NaNs) will behave differently.

### DIFF
--- a/changelog.d/1310.change.md
+++ b/changelog.d/1310.change.md
@@ -1,0 +1,1 @@
+Speed up the generated `__eq__` methods significantly.
--- a/changelog.d/1310.change.md
+++ b/changelog.d/1310.change.md
@@ -1,1 +1,0 @@
-Speed up the generated `__eq__` methods significantly.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -1871,7 +1871,7 @@ def _make_eq(cls, attrs):
                 # of the evaluated function.
                 globs[cmp_name] = a.eq_key
                 lines.append(
-                    f"        {cmp_name}(self.{a.name}) == {cmp_name}(self.{a.name})"
+                    f"        {cmp_name}(self.{a.name}) == {cmp_name}(other.{a.name})"
                 )
             else:
                 lines.append(f"        self.{a.name} == other.{a.name}")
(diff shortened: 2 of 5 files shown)


src/attr/_make.py
```diff
@@ -1864,20 +1864,20 @@ def _make_eq(cls, attrs):
...
         for a in attrs:
             if a.eq_key:
                 cmp_name = f"_{a.name}_key"
                 # Add the key function to the global namespace
                 # of the evaluated function.
                 globs[cmp_name] = a.eq_key
-                lines.append(f"        {cmp_name}(self.{a.name}),")
-                others.append(f"        {cmp_name}(other.{a.name}),")
+                lines.append(
+                    f"        {cmp_name}(self.{a.name}) == {cmp_name}(self.{a.name})"
```
hynek, line 1874: You’re comparing the value against itself — why is this not caught by our tests? I hope your gainz aren’t just that this gets optimized away. Zanyy

changelog.d/1310.change.md
```diff
@@ -0,0 +1 @@
+Speed up the generated `__eq__` methods significantly.
```
hynek, line 1: We should document the change in behavior here. You can make it a breaking fragment, I don’t care that much, but it should definitely be there.

there might be some docs that claim the tuple way too?

hynek (approved):
cool thanks! 🏎️

filbranden:
You might want to update the comment above those lines that says we can't do exactly what we're doing now 😄 

https://github.com/python-attrs/attrs/blob/13e9a6aaeac156bf792878eceaba10321728fa36/src/attr/_make.py#L1535-L1536

hynek:
> You might want to update the comment above those lines that says we can't do exactly what we're doing now 😄
> 
> https://github.com/python-attrs/attrs/blob/13e9a6aaeac156bf792878eceaba10321728fa36/src/attr/_make.py#L1535-L1536

thanks, it's so hard to find quality free labor 🤪

PR #1316, opened 2024-07-31
Invert API docs to point from NG to OG

hynek:
People use and love the new APIs, we've got the Pont Neuf problem, let's invert the API docs.

- `define()`
   - https://attrs--1316.org.readthedocs.build/en/1316/api.html#attrs.define
   - https://attrs--1316.org.readthedocs.build/en/1316/api-attr.html#attr.s
- `field()`
   - https://attrs--1316.org.readthedocs.build/en/1316/api.html#attrs.field
   - https://attrs--1316.org.readthedocs.build/en/1316/api-attr.html#attr.ib


cipri-tom:
Well, in https://attrs--1316.org.readthedocs.build/en/1316/api.html#attrs.define , the doc string should also be modified so that 

> according to the specified attributes using attr.field or _these_

rather than

> according to the specified attributes using attr.ib or _these_

Yedfree (approved):

Yedfree (approved):

PR #1317, opened 2024-07-31
More f-strings

hynek:
Let's see if CodSpeed can tell the difference.


hynek:
It can't, but it's nicer. 🤓

PR #1322, opened 2024-08-03
Simplify & optimize own attribute detection

hynek:
I admittedly didn't know, that that class objects always have an `__dict__` with their own attributes (unlike instances).


hynek:
weee 
CleanShot 2024-08-03 at 13 23 31@2x

Issue #1327, opened 2024-08-05
Upgrade to 24.1.0 seems to break assignment to fields with chained converters


leroyvn:
Upon upgrading to v24.1.0, I noticed that some of my code using fields with chained converters raises an `AttributeError`. Here is a MWE:

```python
import attrs


def foo(x):
    return x


def bar(x):
    return x


@attrs.define
class MyClass:
    field = attrs.field(converter=[foo, bar])


x = MyClass(0)
x.field = 1
print(x)
```

That code runs successfully with v23.2.0, but it raises with the following stack trace with v24.1.0:

```
Traceback (most recent call last):
  File "/home/leroyv/Documents/src/rayference/rtm/eradiate/playgrounds/bug_attrs_241.py", line 18, in <module>
    x.field = 1
  File "/home/leroyv/miniforge3/envs/eradiate/lib/python3.9/site-packages/attr/_make.py", line 1057, in __setattr__
    nval = hook(self, a, val)
  File "/home/leroyv/miniforge3/envs/eradiate/lib/python3.9/site-packages/attr/setters.py", line 23, in wrapped_pipe
    rv = setter(instance, attrib, rv)
  File "/home/leroyv/miniforge3/envs/eradiate/lib/python3.9/site-packages/attr/setters.py", line 66, in convert
    return c(new_value)
AttributeError: __call__
```

hynek:
yeah this is really bad, #1328 should fix it

PR #1328, opened 2024-08-05
Make Converter a kind of adapter, fix converters.pipe

hynek:
Looks like piped converters are hosed. 😳

There's still the issue when a Converter comes last in a pipe, but this fixes at least #1327 and that pipe used to pass nonsense into converters.


src/attr/_make.py
```diff
@@ -2724,6 +2723,21 @@ def __init__(self, converter, *, takes_self=False, takes_field=False):
...
         ).get_first_param_type()
 
+    def __call__(self, value: object, instance: object, field: Attribute):
+        """
+        Call the converter with the appropriate subset of *instance* and
+        *field*.
+        """
+        return self.converter(
+            value,
+            *{
```
Tinche, line 2733: This dict thing is cute but it's going to be slower than some if statements, and less readable :/

Tinche: Can we do this decision at Converter instantiation time instead of calling time?

Tinche (approved):

src/attr/_make.py
```diff
@@ -2724,6 +2724,25 @@ def __init__(self, converter, *, takes_self=False, takes_field=False):
...
+
+    def _takes_instance(self, value, instance, field):
+        return self.converter(value, instance)
+
+    def _takes_field(self, value, instance, field):
+        return self.converter(value, field)
+
+    def _takes_both(self, value, instance, field):
+        return self.converter(value, instance, field)
+
```
hynek, line 2745: @Tinche how about now? here, the dict even makes a bit of sense ;) bummer we need `self`, otherwise I would make it a class var too

PR #1329, opened 2024-08-06
Restore basic functionality on 3.14[sic]

### LINKED ISSUE #1326: Latest release breaks hypothesis tests in CPython main branch
See https://github.com/python/cpython/issues/122686.  Pinning to "attrs<=23.2.0" - a workaround.

hynek:
Essentially switch to PEP 649 / 749 for annotations. Some tests need to be skipped for now, but the rest is working.

Fixes #1326

We probably should open issues for the failing tests, but I'd rather do that only after the first alpha.

### DIFF
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -585,15 +585,12 @@ class TestAnnotations:
         assert typing.List[int] == attr.fields(A).b.type
         assert typing.List[int] == attr.fields(A).c.type
 
-    @pytest.mark.skipif(
-        PY_3_14_PLUS,
-        reason="Forward references are changing a lot in 3.14. "
-        "Passes only for slots=True",
-    )
     def test_self_reference(self, slots):
         """
         References to self class using quotes can be resolved.
         """
+        if PY_3_14_PLUS and not slots:
+            pytest.xfail("References are changing a lot in 3.14.")
 
         @attr.s(slots=slots, auto_attribs=True)
         class A:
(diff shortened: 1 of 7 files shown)


tests/test_annotations.py
```diff
@@ -584,6 +585,11 @@ class A:
         assert typing.List[int] == attr.fields(A).b.type
         assert typing.List[int] == attr.fields(A).c.type
 
+    @pytest.mark.skipif(
+        PY_3_14_PLUS,
+        reason="Forward references are changing a lot in 3.14. "
+        "Passes only for slots=True",
```
webknjaz, line 591: Why not call `pytest.xfail()` conditionally, then?

hynek: because it passes with slots=True – can I refer to other fixtures in the xfail condition?

webknjaz: The `pytest.xfail()` helper is not a marker (`pytest.mark.xfail()`). You have to call it inside the test function body and it's more flexible since conditionals are in the code.

skirpichev: You can use xfail in function body like this:
```python
if PY_3_14_PLUS and slots is not True:
    pytest.xfail("")
```
(But I doubt it's doable with a decorator.)

hynek: True, but that's not a real xfail tho. I like strict xfails that start failing once the test isn't failing anymore, but sure, it's semantically more correct.

webknjaz: I suppose what you want could be implemented with indirect fixtures or other hooks. That would be a bit more complicated, though.

Yedfree (approved):

PR #1337, opened 2024-08-23
Fix license definition in project metadata

mschoettle:
# Summary

The license definition in `pyproject.toml` is not valid according to the [specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#license).

As a result, the PyPI API returns `license: null` for this package: https://pypi.org/pypi/attrs/json

This PR fixes this.

# Pull Request Check List

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "hatchling.build"
 [project]
 name = "attrs"
 authors = [{ name = "Hynek Schlawack", email = "hs@ox.cx" }]
-license = "MIT"
+license = { text = "MIT" }
 requires-python = ">=3.7"
 description = "Classes Without Boilerplate"
 keywords = ["class", "attribute", "boilerplate"]


webknjaz (approved):
Does `twine check --strict` not catch this? 🤔 Is it not running in CI?

hynek:
It didn't because it was completely OK, using a widely-supported, almost-done PEP 639 SPDX identifier which has been the default for Hatch projects for years now BTW.

Given PyPI's web interface has always shown the trove ID classifiers for license, I find it extremely unfortunate that the API chose to use a different source and breaking the ecosystem like this.

webknjaz:
Yeah, the free-form field still shows up additionally. I hope Karolina gets to finish the PEP this year..

hynek:
JFTR, I reverted this once PEP 639 passed and waited for `gh-action-pypi-publish` and hatchling to grow support for Metadata 2.4.

AFAICT, the [wheel metadata](https://github.com/python-attrs/attrs/actions/runs/12347519570#summary-34454748601) is correct:

```
Metadata-Version: 2.4
Name: attrs
Version: 24.3.0
Summary: Classes Without Boilerplate
Project-URL: Documentation, https://www.attrs.org/
Project-URL: Changelog, https://www.attrs.org/en/stable/changelog.html
Project-URL: GitHub, https://github.com/python-attrs/attrs
Project-URL: Funding, https://github.com/sponsors/hynek
Project-URL: Tidelift, https://tidelift.com/subscription/pkg/pypi-attrs?utm_source=pypi-attrs&utm_medium=pypi
Author-email: Hynek Schlawack <hs@ox.cx>
License-Expression: MIT
License-File: LICENSE
```

The API still returns bad data:

```
    "license": null,
    "license_expression": null,
    "license_files": null,
```

But AFAICT that's problem on PyPI's side, not ours. I'm open to suggestions if we're doing anything wrong, but I'm not going to stop using an accepted PEP because some corner of Python packaging's ecosystem isn't up to date yet.

PR #1346, opened 2024-09-14
Stop installing into CI's system

hynek:
We're starting to get permission denied errors when installing with `--system`.

Is it Halloween already or what federal holiday is coming up?


hynek:
i guess this is cleaner anyway 🫡 @hugovk

PR #1349, opened 2024-09-19
Use a fully locked test environment

hynek:
This is currently more of an open experiment to find out what's missing in uv for FOSS projects.

The first obvious problem is that it locks the local version of the package that changes after every commit (so we can have Test PyPI uploads).


hynek:
OK, problem number 2 is that we don't have `--installpkg` anymore.

That's a) a bummer because it's nice to test actual wheels, but also b) too much work to work around right now, so I'll get back to it later.

hynek:
> OK, problem number 2 is that we don't have `--installpkg` anymore.
> 
> That's a) a bummer because it's nice to test actual wheels, but also b) too much work to work around right now, so I'll get back to it later.

Problem number 2 has been fixed in tox-uv 1.13! https://github.com/tox-dev/tox-uv/releases/tag/1.13.0

hynek:
hrhm the failure on 3.8 is AFAICT caused by the fact that tox-uv dropped 3.8 and the older tox-uv versions have no support for groups. Well, let's see how long it take uv to fix the dynamic version problem.

hynek:
So the final piece appears to be that `uv sync` won't sync anything into our PyPy environment:

```
$ tox -re pypy3-tests
pypy3-tests: remove tox env folder /Users/hynek/FOSS/attrs/.tox/pypy3-tests
pypy3-tests: venv> /Users/hynek/.local/share/uv/tools/tox/bin/uv venv -p pypy3 --allow-existing /Users/hynek/FOSS/attrs/.tox/pypy3-tests
pypy3-tests: uv-sync> uv sync --frozen --no-dev --group tests
pypy3-tests: commands[0]> pytest tests/test_functional.py
pypy3-tests: failed with pytest is not allowed, use allowlist_externals to allow it
  pypy3-tests: FAIL code 1 (1.29 seconds)
  evaluation failed :( (1.35 seconds)
```

and:

```
$ ls .tox/pypy3.10-tests/lib/pypy3.10/site-packages/
 _attrs.pth   _virtualenv.pth   _virtualenv.py   attrs-24.3.1.dev11.dist-info
```

@gaborbernat @charliermarsh do y'all think this is a tox-uv a an uv thing? I can't find a related issue in either bug tracker but I wouldn't be surprised if nobody has tried this with PyPy yet. FWIW, PyPy works [perfectly fine](https://github.com/python-attrs/attrs/actions/runs/12703951782/job/35412550661) without dependency groups/locking.

hynek:
FTR: if we'd be OK with dropping 3.8, uv has added all the bits and this would be ready to go.

hynek:
FTR: since this PR also introduces dependency groups, we can't merge it until Pip merges support for them. There's gonna be enough noise for depending on an up-to-date Pip version; I'm not ready to deal with the backlash of requiring _uv_.

Yedfree (approved):

Tinche (approved):

Tinche:
I need a cigarette

PR #1351, opened 2024-09-22
Rephrase the "Private Attributes and Aliases" section

geofft:
It took me a bit to figure out what was meant by the link to the characteristic issue and why there is no such thing as a private argument. Incorporate (my understanding of) the meaning inline and also provide an example of how to work around the _1 syntax error.

# Summary

Hopefully this new text is more understandable, but I'm happy to rework it further and I won't be sad if you disagree wholly.


hynek (approved):
thanks!

PR #1352, opened 2024-09-23
Document how to handle class variables/constants in typed syntax

geofft:
This is mentioned very briefly in the API docs for `auto_attrib` but it's not easy to find if you don't already know about `typing.ClassVar`.


hynek (approved):
thanks!!

PR #1355, opened 2024-09-27
Remove mention of cmp= from next gen docs

tjstum:
# Summary

`attrs.define` and `attrs.field` do not actually accept `cmp=` anymore, but the docs still discuss them

# Pull Request Check List
Documentation-only fix!


hynek (approved):
huh, thanks!

I could've sworn we've added cmp after un-deprecating it, since it seems like a useful shortcut to set both. But nobody seems to be missing it, so let's rip out the wrong docs!

PR #1358, opened 2024-10-13
Introduce NothingType

Tinche:
So I introduced NothingType, but from what I can tell we're not using this anywhere? This is solely for the benefit of third parties annotating their code to accept NOTHING?


hynek:
I presume this is based on https://discuss.python.org/t/do-we-need-a-valuealias-in-addition-to-typealias/67609? I guess it needs _some_ docs and a changelog entry?

hynek:
_nudge_

hynek:
Is there any chance at all to test this??

hynek (approved):

PR #1379, opened 2024-12-09
Use uvx to run Zizmor

hynek:


hynek:
even without caching 25% faster 🚢

PR #1382, opened 2024-12-13
Fix type signature for callables passed to Converter with takes_self=True

### LINKED ISSUE #1378: Cannot get attrs.Converter() with takes_self pass through mypy
I've been playing with the `attrs.Converter()` but I am unable to get it type-checked with mypy when using `takes_self`. Now I am not considering (the obvious) `Unsupported converter, only named functions, types and lambdas are currently supported` error. What I am struggling with is how to annotate the self. The following code:

```
def str2int(s: str|int, self_: "C") -> int:
    return int(s)

@attrs.define
class C:
    x: int = attrs.field(
        converter=attrs.Converter(str2int, takes_self=True)
    )
```

Leads to the following error:

```
error: No overload variant of "Converter" matches argument types "Callable[[str | int, C], int]", "bool"  [call-overload]
note: Possible overload variants:
note:     def [In, Out] Converter(self, converter: Callable[[In], Out]) -> Converter[In, Out]
note:     def [In, Out] Converter(self, converter: Callable[[In, AttrsInstance, Attribute[Any]], Out], *, takes_self: Literal[True], takes_field: Literal[True]) -> Converter[In, Out]
note:     def [In, Out] Converter(self, converter: Callable[[In, Attribute[Any]], Out], *, takes_field: Literal[True]) -> Converter[In, Out]
note:     def [In, Out] Converter(self, converter: Callable[[In, AttrsInstance], Out], *, takes_self: Literal[True]) -> Converter[In, Out]
```

Is this just my own ignorance or an issue with attrs or mypy?

I am using:
* attrs 24.2.0
* mypy 1.13.0

filbranden:
# Summary

Fix type signature for callables passed to `Converter` with `takes_self=True`, so that using the specific attrs class that contains the field with the converter is valid in the callable type signature.

Fixes #1378.

Note that python/mypy#15736 still triggers mypy errors when passing a Converter instance as a `converter=` kwarg to a field, so work around that using `# type: ignore[misc]` in test cases.

# Pull Request Check List

- N/A New features have been added to our [Hypothesis testing strategy](https://github.com/python-attrs/attrs/blob/main/tests/strategies.py).
    - N/A If they've been added to `attr/__init__.pyi`, they've *also* been re-imported in `attrs/__init__.pyi`.
- N/A Updated **documentation** for changed code.
- N/A Documentation in `.rst` and `.md` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).
- N/A Changes (and possible deprecations) have news fragments in [`changelog.d`](https://github.com/python-attrs/attrs/blob/main/changelog.d).

### DIFF
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -95,13 +95,16 @@ def Factory(
 In = TypeVar("In")
 Out = TypeVar("Out")
 
+# Callable might type check against the specific attrs-defined class.
+_AttrsInstance = TypeVar("_AttrsInstance", bound=AttrsInstance)
+
 class Converter(Generic[In, Out]):
     @overload
     def __init__(self, converter: Callable[[In], Out]) -> None: ...
     @overload
     def __init__(
         self,
-        converter: Callable[[In, AttrsInstance, Attribute], Out],
+        converter: Callable[[In, _AttrsInstance, Attribute], Out],
         *,
         takes_self: Literal[True],
         takes_field: Literal[True],
@@ -116,7 +119,7 @@ class Converter(Generic[In, Out]):
     @overload
     def __init__(
         self,
-        converter: Callable[[In, AttrsInstance], Out],
+        converter: Callable[[In, _AttrsInstance], Out],
         *,
         takes_self: Literal[True],
     ) -> None: ...
--- a/tests/test_mypy.yml
+++ b/tests/test_mypy.yml
@@ -911,28 +911,28 @@
     main:22: error: No overload variant of "Converter" matches argument types "Callable[[Any], str]", "bool"  [call-overload]
     main:22: note: Possible overload variants:
     main:22: note:     def [In, Out] Converter(self, converter: Callable[[In], Out]) -> Converter[In, Out]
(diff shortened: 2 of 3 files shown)


Tinche:
So what exact issue are we trying to solve here?

There's kind of no point in using a TypeVar if we're going to reference it only once - the usual use case is to link a type in the function parameters to a type in the function return value, or a type in the class to a type in a method.

filbranden:
@Tinche see #1372, though the author looped back in and mentioned finding a workaround.

But I think it's not unreasonable that a converter for a field in attrs class C would take an instance of class C as its argument.

Indeed it would be better if we could link the TypeVar to the definition, but I'm not sure how to do that part, because that would involve taking the class in which a `field(converter=...)` was called.

If you think the workaround in #1372 (taking an `attrs.AttrsInstance`) is fine, then it should be good to just drop this, though we might consider improving documentation to indicate that's the case. Thanks!

filbranden:
Closing, as per discussion using `attrs.AttrsInstance` is correct and acceptable.

hynek:
Thanks for the discussion tho!

PR #1383, opened 2024-12-14
Implement __replace__ on 3.13

### LINKED ISSUE #1313: `copy.replace` support
Python 3.13 introduced the `__replace__` dunder, exposed via [`copy.replace`](https://docs.python.org/3.13/library/copy.html#copy.replace), as a [generalised method](https://discuss.python.org/t/generalize-replace-function/28511) through which to replace fields in immutable dataclass-likes.  Would you be interested in supporting it in addition to attrs' own `evolve`?

hynek:
Fixes #1313

Had to move evolve over due to import cycles and fortunately RTD already supports 3.13.

Since it doesn't overwrite anything, I would wait with adding an own option for if someone actually misses it. 🤷‍♂️

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,7 +185,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           # Keep in sync with tox/docs and .readthedocs.yaml.
-          python-version: "3.12"
+          python-version: "3.13"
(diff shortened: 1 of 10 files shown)


Tinche (approved):
LGTM.

We could probably make this faster by generating a specialized version of `__replace__` and making evolve use it, but that would slow down class creation even further, so I'm not convinced it's worth it.

PR #1385, opened 2024-12-14
typing: allow union types for instance_of

### LINKED ISSUE #1336: Mypy false positive for union validator
Not sure if this is a duplicate but I couldn't find the problem in the open issues. If I overlooked it, please close, of course.

```python
from attrs import define, field
from attrs.validators import instance_of

LeTypeAlias = int | str


@define
class LeClass:
    x: LeTypeAlias = field(validator=instance_of(LeTypeAlias))
```

Works perfectly but `mypy` complains about:
```
error: No overload variant of "instance_of" matches argument type "UnionType[int, str]"  [call-overload]
```

hynek:
fixes #1336

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
         run: |
           DO_MYPY=1
 
-          if [[ "$V" == "3.8" ]]; then
+          if [[ "$V" == "3.8" || "$V" == "3.9" ]]; then
             DO_MYPY=0
           fi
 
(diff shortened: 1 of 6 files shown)


hynek:
Tiiiiin, is this fixableeeee?

```
tests/typing_example.py:231: error: Incompatible types in assignment (expression has type "UnionType", variable has type "int | C | str")
[assignment]
        kk: int | C | str = attr.ib(
                            ^
Found 1 error in 1 file (checked 1 source file)
```

Or are we doing typing bankruptcy (aka Any)?

Tinche:
I played around with this a little. I don't think the type system can do this. The overload should be:
```python
@overload
def instance_of(type: UnionType) -> _ValidatorType[Any]: ...
```

PR #1387, opened 2024-12-16
FIX: License metadata not published in Pypi

### LINKED ISSUE #1386: Pypi does not recognize license
Unless there is a delay, it seems that the LICENSE metadata is not available in Pypi for release 24.3.0, see https://pypi.org/project/attrs/24.3.0/

Note that this information was available in 24.2.0, see https://pypi.org/project/attrs/24.2.0/

I suspect the changes from https://github.com/python-attrs/attrs/pull/1377 are the cause for this problem.

SMoraisAnsys:
# Summary

~Temporary fix which revert "Apply PEP 639 (#1377)"  to avoid the current issue with license metadata not available in Pypi.~
Apply changes proposed in https://github.com/python-attrs/attrs/issues/1386#issuecomment-2545145428:
- keep Trove classifier (deprecated for now, not removed)
- keep the new PEP 638 fields

Close #1386

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,8 @@ build-backend = "hatchling.build"
 [project]
 name = "attrs"
 authors = [{ name = "Hynek Schlawack", email = "hs@ox.cx" }]
-license = { text = "MIT" }
+license = "MIT"
+license-files = ["LICENSE"]
 requires-python = ">=3.8"
 description = "Classes Without Boilerplate"
 keywords = ["class", "attribute", "boilerplate"]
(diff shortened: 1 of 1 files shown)


SMoraisAnsys:
Using `license = {file = "LICENSE"}` could also be an option

webknjaz:
`pypi-publish` has been released: https://bsky.app/profile/webknjaz.me/post/3lghhtwey522z. The issue will solve itself as soon as it's bumped in the workflow (and the release happens).

Issue #1286, opened 2024-05-19
Add "resolve_types" argument to attrs.define()


sscherfke:
Add a `resolve_types: bool = ?` argument to `attrs.define()` and call `resolve_types()` on the new class if `True` is passed.

*Why?*

According to the Typed Settings issue tracker, people seem to start playing with postponed annotations. 
For Typed Settings to work, annotated types need to be resolved.
We try to automatically do this but it is not always possible to do it.
So people need to do sth. like this:

```python
@attrs.resolve_types
@attrs.define
class Settings:
    opt: int = 0
```

It would be a bit nicer if they could write
```python
@attrs.define(resolve_types=True)
class Settings:
    opt: int = 0
```

instead.  This would also make it easier to write custom `define()` partials similar to `frozen()`.

I can create a PR if needed.

sscherfke:
Pydantic [automatically does this](https://docs.pydantic.dev/latest/concepts/postponed_annotations/) so I guess users might eventually expected this behavior from attrs, too.

hynek:
I'm open to PRs!

PR #1390, opened 2024-12-25
Draft: Add "resolve_types" argument to define()

sscherfke:
Fixes: #1286

# Summary

Add *resolve_types* kwarg to `define()` to allow it to automatically resolve the class' fields types.

This is atm just a PoC to evaluate if this is going into the right direction. What do you think, @hynek ?

# Pull Request Check List

- [ ] Do **not** open pull requests from your `main` branch – **use a separate branch**!

### DIFF
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -279,6 +279,7 @@ def attrs(
     field_transformer: _FieldTransformer | None = ...,
     match_args: bool = ...,
     unsafe_hash: bool | None = ...,
+    resolve_types: bool = ...,
 ) -> _C: ...
 @overload
 @dataclass_transform(order_default=True, field_specifiers=(attrib, field))
@@ -307,6 +308,7 @@ def attrs(
     field_transformer: _FieldTransformer | None = ...,
     match_args: bool = ...,
     unsafe_hash: bool | None = ...,
+    resolve_types: bool = ...,
 ) -> Callable[[_C], _C]: ...
 def fields(cls: type[AttrsInstance]) -> Any: ...
 def fields_dict(cls: type[AttrsInstance]) -> dict[str, Attribute[Any]]: ...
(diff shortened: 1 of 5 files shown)


hynek:
typing is @Tinche territory ;)

Tinche:
What's the exact use case here?

A couple of notes:
* On 3.14 we likely won't need resolve_types any more
* If resolve_types can be called immediately (no forward refs), the consumer of the annotations can call it if they see a string annotation - the producer shouldn't need to worry about it

sscherfke:
This is purely for convenience. I also implemented this in the Typed Settings' `settings` decorator ([MR](https://gitlab.com/sscherfke/typed-settings/-/merge_requests/35), [fixup](https://gitlab.com/sscherfke/typed-settings/-/commit/17c071b19a48a304c6fc164f7ae71d5df219fc42)).

So I don't _really_ require it for TS but I was wondering if this might improve convenience for other users / use cases as well.

Why won't this be needed on py314 anymore?  I thould that annotations would allways be "unresolved strings" in a future Python version and that thus, `resolve_types()` would be even more important?

sscherfke:
@Tinche ping :)

Tinche:
Here's my understanding of how PEP 649 works in 3.14.

Attrs itself will need to use `get_annotations` with `format=STRING`. attrs is only concerned with whether an annotation is a ClassVar, if memory serves.

All other code that runs after the class and all other classes it references are defined (cattrs) will use `format=VALUE` and let the interpreter resolve the types. The way we expose this would probably be by making `Attribute.type` a property or something that calls `get_annotations` under the hood.

Even ignoring all of that, I don't think `resolve_types` on `define` makes too much sense - it would make more sense on `fields`. Folks generally use stringified annotations when they have forward references, and those cannot be resolved when the class is constructed, but sometime later.

Issue #1397, opened 2025-01-27
Cloudpickle supports Python 3.14 already


frenzymadness:
I think the xfail marker here is not needed anymore:

https://github.com/python-attrs/attrs/blob/fbff14eeaf817e4b34e9c5727c8d546dc118a346/tests/test_3rd_party.py#L19-L21

as the latest release of cloudpickle is compatible with Python 3.14. See https://github.com/cloudpipe/cloudpickle/commit/7468d7284c79ba11dedc99afa72218ca6b7abf49 and https://github.com/cloudpipe/cloudpickle/blob/master/CHANGES.md#311

PR #1398, opened 2025-01-27
Don't expect TestCloudpickleCompat to fail with Python 3.14

frenzymadness:
Cloudpickle supports Python 3.14 already.

Fixes: https://github.com/python-attrs/attrs/issues/1397


hynek (approved):
Thanks!

Issue #1404, opened 2025-02-05
Valid Unicode identifier can't be used as `attrs.make_class` parameter


serge-sans-paille:
Simple reproducer:

```
>>> import attrs
>>> clsname = '䆊ㅁ'
>>> attrs/make_class(clsname, [])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/site-packages/attr/_make.py", line 3042, in make_class
    return _attrs(these=cls_dict, **attributes_arguments)(type_)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/attr/_make.py", line 1619, in wrap
    builder = _ClassBuilder(
              ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/attr/_make.py", line 724, in __init__
    attrs, base_attrs, base_map = _transform_attrs(
                                  ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/attr/_make.py", line 598, in _transform_attrs
    AttrsClass = _make_attr_tuple_class(cls.__name__, attr_names)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/attr/_make.py", line 374, in _make_attr_tuple_class
    return globs[attr_class_name]
           ~~~~~^^^^^^^^^^^^^^^^^
KeyError: '䆊ㅁAttributes'
```

This was raised by fuzzing when setting up https://github.com/google/oss-fuzz/pull/13009

hynek:
Thanks, but this is out of scope. I've added a warning in 5820ce717ed3, though.

PR #1406, opened 2025-02-08
Accept valid unicode identifiers as class parameter to make_class

serge-sans-paille:
Related to #1404

### DIFF
--- a/changelog.d/1406.change.md
+++ b/changelog.d/1406.change.md
@@ -1,0 +1,1 @@
+`attrs.make_class()` now allows for Unicode class names.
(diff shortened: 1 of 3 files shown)


hynek:
Thanks! How would you summarize the change for the changelog and `.. versionchanged`?

> `attrs.make_class()` now allows for Unicode class and field names.

?

serge-sans-paille:
It's actually just

> `attrs.make_class()` now allows for Unicode class name

Attribute name was correctly supported before that patch, it's just my testing code that was incorrect (I was testing a `kwargs` that contained the unnormalized attribute name, and that doesn't work for real class, there's no reason it should work with `attrs`)

hynek:
thanks; it was faster to do it myself, than explaining it: https://github.com/python-attrs/attrs/pull/1406/commits/fc8b9c6af03e812b1609a3ec0c4e873573e18ad2

hynek (approved):

PR #1414, opened 2025-03-09
Fix a typo in SPDX comment @ `pyproject.toml`

webknjaz:
It was introduced in https://github.com/python-attrs/attrs/commit/6bde3618237cdae312e4e7fb006690be2a3ee0f6


hynek (approved):
Ugh thanks!

PR #1422, opened 2025-03-26
Add MIT license pypi classifier

timmow:
# Summary

This adds a pypi [classifier](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) to specify this package is using the MIT license. This came up as we have tooling to mirror and check the license of pypi packages using this classifier


Tinche:
See https://github.com/python-attrs/attrs/pull/1377

Issue #1424, opened 2025-04-06
Can only populate attrs classes with `_CountingAttr` instances


redruin1:
A user might want to use attributes from an already `@define`d class to populate a new attrs class. However, `_ClassBuilder` expects all given attributes to be instances of `_CountingAttr` instead of `Attribute` instances, which raises an exception when provided:

```py
import attrs

@attrs.define
class Example:
    a: int

ExampleCopy = attrs.make_class("ExampleCopy", attrs={"a": attrs.fields(Example).a})
# AttributeError: 'Attribute' object has no attribute '_validator'. Did you mean: 'validator'?

@attrs.define(these={"a": attrs.fields(Example).a})
class ExampleCopy:
    pass
# AttributeError: 'Attribute' object has no attribute '_validator'. Did you mean: 'validator'?
```
I would either expect `_ClassBuilder` to handle both kinds of attributes, or there be an officially sanctioned function allowing you to convert an `Attribute` to a `_CountingAttr` for cases like this. You can write this conversion function manually, something like:
```py
def convert_to_countingattr(attr):
    """
    Convert an `Attribute` instance to an equivalent `_CountingAttr` instance.
    """
    return attrs.field(**{
        slot: getattr(attr, slot) 
        for slot in attr.__slots__ 
        if slot not in {"name", "eq_key", "order_key", "inherited"}
    })
```
But this feels janky and prone to breakage.

redruin1:
This seems related to #637; the proposed `to_field()` method would make this trivial.

The utility of this behavior is that it would make it much easier to manipulate the user class before it is passed to attrs plumbing, so you could do things like custom attribute ordering or similar.

Issue #946, opened 2022-04-04
Inherit annotation type from base class that defines new default value


R0ll1ngSt0ne:
Consider this code:

```python
import attr


@attr.s(auto_attribs=True, kw_only=True)
class A:
    foo: str


@attr.s(auto_attribs=True, kw_only=True)
class B(A):
    foo = "bar"


b = B()
```

When compiled, this fails with:

```
TypeError: __init__() missing 1 required keyword-only argument: 'foo'
```

Yes, in this case, it would not be trouble to _re-annotate_ "foo" as "foo: str" in the child class as well - but our classes use heavy/multi-level annotations like:

```python
@attr.s(auto_attribs=True, kw_only=True)
class A:
    foo: List[Union[Need, Job, Sequence]]
    bar: List[Union[Job, Sequence]]
    baz: Union[bool, str, int, List[int]]
    .
    .
    .
```

We have _dozens_ of such required parameters - it would be a great hassle to keep re-annotating them in every child class that provides a new default value. It would also be almost impossible to keep track of all re-annotations, as the number of child classes is growing every day. The whole idea of using attrs was to avoid having to re-annotate the arguments in every child class. Is it possible to achieve with the current framework? If not, can this be implemented?

hynek:
I don't think this can be implemented in a way that doesn't introduce a load of edge cases and surprising behavior.

I suspect this is more another case for #637.

Issue #876, opened 2021-11-29
Allow transfer of attrs between classes.


peteroconnor-bc:
I've run into many cases where, to avoid duplicating types and defaults, it would be nice to be able to "transfer" attributes between classes.  

Example: We would like to do the following:

```
@attrs
class InnerClass:
    a = attrib(type=float, default=3.)

@attrs
class OuterClass:
    inner = attrib(type=InnerClass)
    b = attrib(type=int, default=1)

@attrs
class BuilderClass:
    a = InnerClass.a  # <- This Line fails, forcing us to duplicate types and defaults
    b = OuterClass.b  # <- This one too..

    def build_inner(self) -> InnerClass:
        return InnerClass(self.a)

    def build_outer(self) -> OuterClass:
        # Just to demonstrate why we use this builder class
        return OuterClass(inner=self.build_inner(), b=self.b)

builder = BuilderClass()
inner = builder.build_inner()
assert inner.a == 3.

builder = BuilderClass(a=4.)
inner = builder.build_inner()
assert inner.a == 4.

```

It would be really nice if attrs supported this syntax: 

```
@attrs  
class BuilderClass:
    a = InnerClass.a  # <- This Line fails
```
Instead, we usually end up just doing 
```
@attrs  
class BuilderClass:
    a = attrs(type=float, default=3.) 
```
which requires us to duplicate type/default/validators/doc/etc. 

Is that something that Attrs would be willing to support?

hynek:
could this be part of #829?

peteroconnor-bc:
Seems similar.  Would really be nice to have that syntax supported.  It lets us get around a general shortcoming of python - that there's not really a clean way to defer defaults (see [thread](https://mail.python.org/archives/list/python-ideas@python.org/thread/XQS6RN6MAS5VX66RWBIEULUJWDEFWAEV/)).

hynek:
Closing this in favor of #829.

Your desired syntax won't work tho, since we don't attach attributes to the classes anymore. You're gonna have to use `attrs.fields()` unless we'll find a nicer syntactic sugar.

Issue #829, opened 2021-06-29
Change defaults of an 3rd party attrs class


tc8:
I'm using the [grafanalib library](https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L384) which uses attrs extensively and one of the class looks something like this

```
@attr.s
class Target(object):
    intervalFactor = attr.ib(default=2)
...
```

I want to change the default of `intervalFactor` to `1`. I could specify `1` each time in the constructor but that's very tedious when I have to do it each time. 

I tried subclassing it and playing with `__attrs_pre_init__ `, `__attrs_post_init__ `, and `__attrs_init__ ` but could not find any combination that gave me the desired behavior. 

I've also looked at using `attr.evolve` like `attrs.evolve(default_target, myarg1, myarg2, ...)` where default_target is a Target instance with the default values set. This works but witing `attrs.evolve(default_target` each time in lieu of the constructor is tedious. 

I could wrap it in a helper function, like 
```
def create_target(*args, **kwargs) -> Target: 
  return attrs.evolve(default_target, *args, **kwargs)
```
The problem with this is that I lose the IDE (PyCharm) autocompletion that I would get with writing `attrs.evolve(...)` manually. 

I tried setting the `__signature__` manually by doing `create_target.__signature = inspect.signature(Target.__init__)` but this didn't give me the correct tooltip in the editor model. 

Any help would be greatly appreciated. 

** On a related note**
How is `attrs.evolve` able to provide the right auto-complete suggestions in PyCharm? Is it because of the built-in integration with PyCharm?

hynek:
> I'm using the [grafanalib library](https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L384) which uses attrs extensively and one of the class looks something like this
> 
> ```
> @attr.s
> class Target(object):
>     intervalFactor = attr.ib(default=2)
> ...
> ```

This should work:

```python
@attr.define
class MyTarget(Target):
    intervalFactor = attr.field(default=1)
```

The only downside is that the order of the arguments changes. But that shouldn't be a problem if you don't want to pass it anyway.

> ** On a related note**
> How is `attrs.evolve` able to provide the right auto-complete suggestions in PyCharm? Is it because of the built-in integration with PyCharm?

I think so. PyCharm has added attrs support a long time ago so I wouldn't be surprised by advanced features.

tc8:
Hi @hynek ,

Thank you for your response.

[...]

PR #1429, opened 2025-05-02
Add `to_field()` to `Attribute`

### LINKED ISSUE #637: Option to "partially override" an attribute from a parent class?
Hi,

I'm part of a team that's in the early stages of migrating a large legacy app to Python.  I'm pushing pretty hard for the team to adopt attrs as our standard way of doing OOP.  I know the "attr way" is to avoid inheritance as much as possible, but that's not always possible, especially when you are part of a team with a project that likes to rely on inheritance (generally nothing too crazy, just simple stuff, but still - it's used a lot in the app and it's a good fit a lot of how the app works).  One thing we do a lot is have base classes that device attributes that subclasses "further refine".  With these, we want to be able to "extended" the attribute definition from the parent classes vs. totally overwrite it.  It doesn't seem like this is possible today in attrs.  For example:

```
import attr
import inspect

@attr.s
class One:
    myattrib = attr.ib(
        default=1,
        validator=attr.validators.instance_of(int),
        converter=int,
    )

@attr.s
class Two(One):
    myattrib = attr.ib(
        default=2
    )

print(inspect.getsource(One().__init__))
print(inspect.getsource(Two().__init__))
```
Output when run:
```
def __init__(self, myattrib=attr_dict['myattrib'].default):
    self.myattrib = __attr_converter_myattrib(myattrib)
    if _config._run_validators is True:
        __attr_validator_myattrib(self, __attr_myattrib, self.myattrib)

def __init__(self, myattrib=attr_dict['myattrib'].default):
    self.myattrib = myattrib
```

So if we want to extend the definition of `myattrib`, we would have to copy/paste the full definition vs. just modify what is different.  What would people think about an option that says "take the full definition from the superclasses and just "merge in" this one extra aspect."  Having to repeat the full definition across an inheritance hierarchy gets old (and ugly) pretty quick. :-)

Just by way of context on both this ticket and the other ticket I have opened (#573 - lazy=True option), the legacy app we are using is coming from Perl.  Yes, I know everyone loves to hate Perl, :-) but hear me out. :-)  The app uses Moose (https://metacpan.org/pod/Moose) and it's "lightweight cousin" Moo (https://metacpan.org/pod/Moo) and for all the bad things everyone wants to cast at Perl, the Moose/Moo "OOP framework" is really nice and has some great features.  It has completely changed how OOP is done in Perl and took it from horrible to really nice (nobody has done OOP in Perl for 10+ years without using Moose.)   (The creator of Moose, Stevan Little, spent lots of time studying and using lots of other OOP methodologies and was able to incorporate the "best of the best" ideas.)  It's very similar to attrs in many ways, but there are a few things missing that are SOOO handy, useful, powerful, etc. :-)

In terms of how Moose would write the above example (and use the "+" to signify an "override"):
```
package One;
use Moo;
use Types::Standard 'Int';
has 'myattrib' => (
    is      => 'ro',
    default => 1,
    isa     => Int,
    coerce  => sub { int(shift) },  # Not required for str->int but to show how others work
);

package Two;
use Moo;
extends 'One';
has '+myattrib' => (                # <== Note the '+'
    default => 2,
);
```
Would something like an `extend=True` option to attrs be a nice addition to trigger similar behavior to the "+" shown above?

Thank you

redruin1:
# Summary

Preliminary implementation of "evolving" existing `Attribute` instances back into `_CountingAttr` instances, so that users may (easily) reuse attribute definitions from already `define`d classes. Should resolve #637, #698, #829, #876, #946, and #1424.

Usage:
```py
@attrs.define
class Parent:
    a: int = 100
    
@attrs.define
class Child:
    a = attrs.fields(Parent).a.evolve(default=200).to_field()
    
    # Now that a is in scope of Child, we can attach methods to it like normal:     
    @a.validator
    def a_validator(...):
        ...

assert attrs.fields(Child).a.type is int
assert Child().a == 200
```
This syntax is verbose, but least magical. And because you manually specify which class you want to pull attributes from, inheritance is not required; you can compose new classes entirely from parts of existing classes regardless of structure:
```py
@attrs.define
class A:
    a: int = 1
    
@attrs.define
class B:
    b: int = 2
    
@attrs.define
class Composite:
    a = attrs.fields(A).a.evolve(...).to_field()
    b = attrs.fields(B).b.evolve(...).to_field()
```
Utility methods like `attrs.make_class()` and the `these` kwarg in `attrs.define()` also work as you would expect.

One potential pain point with a simple implementation of this feature is inherited attributes being reordered when redefined in this manner:
```py
@attr.s
class BaseClass:
    x: int = attr.ib(default=1)
    y: int = attr.ib(default=2)

@attr.s
class SubClass(BaseClass):
    x = attr.fields(BaseClass).x.evolve(default=3).to_field()

# Because x was redefined, it appears after y
assert "SubClass(y=2, x=3)" == repr(SubClass())
```
In my opinion, this behavior is a failure of the implementation, as it is reasonable to expect users to want to preserve this ordering, as in a worst-case scenario it can lead to non-constructable classes. This PR implements a new bool keyword argument `inherited` to `field`, which tells attrs to use the ordering of the field in the parent class as opposed to adding to the end:
```py
@attr.s
class SubClass(BaseClass):
    x = attr.fields(BaseClass).x.evolve(default=3, inherited=True).to_field()

assert "SubClass(x=3, y=2)" == repr(SubClass())
```
Criticisms of this `inherited` keyword are welcome, as I'm not convinced it is the best solution. However, I do think this functionality needs to be present in attrs in order to make this kind of attribute evolution a tenable approach.

# Pull Request Check List

### DIFF
--- a/src/attr/__init__.pyi
+++ b/src/attr/__init__.pyi
@@ -139,6 +139,7 @@ class Attribute(Generic[_T]):
     alias: str | None
 
     def evolve(self, **changes: Any) -> "Attribute[Any]": ...
+    def to_field(self) -> Any: ...
 
 # NOTE: We had several choices for the annotation to use for type arg:
 # 1) Type[_T]
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -118,6 +118,7 @@ def attrib(
     order=None,
     on_setattr=None,
     alias=None,
+    inherited=False,
 ):
     """
     Create a new field / attribute on a class.
@@ -206,6 +207,7 @@ def attrib(
         order_key=order_key,
         on_setattr=on_setattr,
         alias=alias,
+        inherited=inherited,
     )
 
 
@@ -434,16 +436,32 @@ def _transform_attrs(
 
     if collect_by_mro:
         base_attrs, base_attr_map = _collect_base_attrs(
-            cls, {a.name for a in own_attrs}
+            cls, {a.name for a in own_attrs if a.inherited is False}
         )
     else:
         base_attrs, base_attr_map = _collect_base_attrs_broken(
-            cls, {a.name for a in own_attrs}
+            cls, {a.name for a in own_attrs if a.inherited is False}
         )
 
     if kw_only:
         own_attrs = [a.evolve(kw_only=True) for a in own_attrs]
         base_attrs = [a.evolve(kw_only=True) for a in base_attrs]
+
+    own_attr_map = {attr.name: attr for attr in own_attrs}
+
+    # Overwrite explicitly inherited attributes in `base` with their versions in `own`
+    base_attrs = [
+        base_attr
+        if base_attr.name not in own_attr_map
+        else own_attr_map[base_attr.name]
+        for base_attr in base_attrs
+    ]
+    # Strip explicitly inherited attributes from `own`, as they now live in `base`
+    own_attrs = [
+        own_attr
+        for own_attr in own_attrs
+        if own_attr.name not in base_attr_map
+    ]
 
     attrs = base_attrs + own_attrs
 
@@ -2501,7 +2519,7 @@ class Attribute:
             None,
             ca.hash,
             ca.init,
-            False,
+            ca.inherited,
             ca.metadata,
             type,
             ca.converter,
@@ -2531,6 +2549,35 @@ class Attribute:
         new._setattrs(changes.items())
 
         return new
+
+    def to_field(self):
+        """
+        Converts this attribute back into a raw :py:class:`_CountingAttr` object,
+        such that it can be used to annotate newly `define`d classes. This is
+        useful if you want to reuse part (or all) of fields defined in other
+        attrs classes that have already been resolved into their finalized state.
+
+        .. versionadded:: 25.4.0
+        """
+        return _CountingAttr(
+            default=self.default,
+            validator=self.validator,
+            repr=self.repr,
+            cmp=None,
+            hash=self.hash,
+            init=self.init,
+            converter=self.converter,
+            metadata=self.metadata,
+            type=self.type,
+            kw_only=self.kw_only,
+            eq=self.eq,
+            eq_key=self.eq_key,
+            order=self.order,
+            order_key=self.order_key,
+            on_setattr=self.on_setattr,
+            alias=self.alias,
+            inherited=self.inherited,
+        )
 
     # Don't use _add_pickle since fields(Attribute) doesn't work
     def __getstate__(self):
@@ -2608,6 +2655,7 @@ class _CountingAttr:
         "eq",
         "eq_key",
         "hash",
+        "inherited",
         "init",
         "kw_only",
         "metadata",
@@ -2646,6 +2694,7 @@ class _CountingAttr:
                 "init",
                 "on_setattr",
                 "alias",
+                "inherited",
             )
         ),
         Attribute(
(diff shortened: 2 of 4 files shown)


Tinche:
Why does it need to be `attrs.fields(Parent).a.evolve(default=200).to_field()` instead of maybe `attrs.fields(Parent).a.to_field(default=200)`? We'll need to add Mypy support to this before it can be used for real, and we should look at making that as easy as possible.

redruin1:
If the goal is to reduce function chains, then how about the keyword `reuse()`? That might be more descriptive in this context than `to_field()`:

```py
class A:
    x = attrs.fields(Example).x.reuse()
    y = attrs.fields(Example).y.reuse(default="whatever")
    _z = attrs.fields(OtherExample).a.reuse(alias="z")
```

Tinche:
I like it!

redruin1:
What exactly should I annotate for mypy? I notice that neither the return result of `attr.fields()` nor `_CountingAttr` itself have annotations already:

https://github.com/python-attrs/attrs/blob/755b1778afcee1a323d8efbd25df73da236ab4ab/src/attr/__init__.pyi#L311

Should I just try to annotate everything this new feature touches? The contribution guidelines are a bit sparse in this regard.

hynek:
So `y = attrs.fields(Example).y.reuse(default="whatever")` is cool indeed – I'm just not sure if that will ever be possible to get into typing at all??

I'm kinda OK with saying that you have to redefine your types, otherwise it gets hairy with ordering anyways (or some magic helper type, e.g., `y: Annotated[Inherit] = attrs.fields(Example).y.reuse(default="whatever")`).

But how could any of this be made type-safe? `y: Annotated[Inherit(attrs.fields(Example).y, default="whatever")]` would be probably even cooler since it circumvents the ordering problems, but I guess we should think ahead here? Does maybe Pydantic have something similar so we could join PEP forces or something?

Tinche:
We could contribute work to the mypy plugin so it would work 😇

redruin1:
> I'm kinda OK with saying that you have to redefine your types, otherwise it gets hairy with ordering anyways (or some magic helper type, e.g., y: Annotated[Inherit] = attrs.fields(Example).y.reuse(default="whatever")).

This would make sense. The only reason it is omitted in this particular implementation is because the type is already inherited from the `reuse()` call and (re)including a type annotation would be akin to defining it twice, which attrs complains about. This can simply be rewritten to not inherit the type, and to default to `Any` if no type annotation is provided.

> `y: Annotated[Inherit(attrs.fields(Example).y, default="whatever")]` would be probably even cooler since it circumvents the ordering problems

This is not a bad idea. If I understand you correctly, does that make the proposed syntax:

```py
@attrs.define
class A:
    x: int = 10
    y: str = attrs.field(default="blah")

@attrs.define
class B:
    z: dict = {"test": "result"}
    
@attrs.define
class Child(B, A):
    x: Annotated[Inherited(attrs.fields(A).x, default=20)] # Preserve parent ordering
    y: str = attrs.fields(A).y.reuse() # Add to end after inherited
    z: Annotated[Inherited(attrs.fields(B).z, factory=dict)] # Preserve parent ordering

assert "Child(x=20, z={}, y='blah')" == repr(Child())
```

It seems the question of what order to put inherited fields in is not trivial. I suppose you could argue that if you need absolute control of field order, then you should create an entirely composite class using `reuse()` and ignore inheritance entirely, which violates DRY a little but keeps things obvious.

> We could contribute work to the mypy plugin so it would work 😇

I'll look into this. I assume you're referring to [this?](https://github.com/python/mypy/blob/master/mypy/plugins/attrs.py)

hynek:
Hey sorry this got sidelined – I still love the concept of it all, it's a bummer that the typing landscape just keeps sucking the joy of creating exciting new things.

I've did my best to resolve the merge conflicts, but there's one fundamental problem here: you've added an `inherited` flag and… we already have one that signals that a field was inherited. So it's pretty close in meaning to cause maximum confusion. If you're interested in keeping going here, you'd need to rename that one (I think something like `keep_order` could work?).

Again, sorry.

redruin1:
Sure, `keep_order` seems a better name. I suppose I'm willing to make further changes as well if you're interested in merging it, but it's unclear to me what exactly those changes should be, so some guidance in that regard would be helpful.

AdrianSosic:
Just wanted to upvote and express my interest in this feature ⭐

PR #1433, opened 2025-05-29
ci: pin non-gh, non-hynek, non-pypa uses

hynek:


hynek:
> I'd just pin everything except github and remove the config file 😇

I'd have to add a Pinact config files otherwise. :D

PR #1445, opened 2025-07-20
docs: clarify how to use tox

hynek:
cc @judy2k

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -49,12 +49,16 @@ It's people like *you* who make this project such a great tool for everyone.
 
 First, **fork** the repository on GitHub and **clone** it using one of the alternatives that you can copy-paste by pressing the big green button labeled `<> Code`.
 
-You can (and should) run our test suite using [*tox*](https://tox.wiki/).
+You can (and should) run our test suite using [*tox*](https://tox.wiki/) with the [*tox-uv*](https://github.com/tox-dev/tox-uv) plugin.
+The easiest way is to [install *uv*] which is needed in any case and then run `uv tool install --with tox-uv tox` to have it globally available or `uvx --with tox-uv tox` to use a temporary environment.
+
+---
+
 However, you'll probably want a more traditional environment as well.
 
 We recommend using the Python version from the `.python-version-default` file in the project's root directory.
 
-We use a fully-locked development environment using [*uv*](https://docs.astral.sh/uv/) so the easiest way to get started is to [install *uv*](https://docs.astral.sh/uv/getting-started/installation/) and you can run `uv run pytest` to run the tests immediately.
+We use a fully-locked development environment using [*uv*](https://docs.astral.sh/uv/) so the easiest way to get started is to [install *uv*] and you can run `uv run pytest` to run the tests immediately.
 
 I you'd like a traditional virtual environment, you can run `uv sync` and it will create a virtual environment named `.venv` with the correct Python version and install all the dependencies in the root directory.
 
@@ -264,3 +268,4 @@ By participating in this project you agree to abide by its terms.
 Please report any harm to Hynek Schlawack in any way you find appropriate.
 
 [semantic newlines]: https://rhodesmill.org/brandon/2012/one-sentence-per-line/
+[install *uv*]: https://docs.astral.sh/uv/getting-started/installation/


Tinche:
Why not just add tox and tox-uv to our deps and run them via uv run? That way we also control the versions. 

Also why not just dump tox altogether in favor of a `just testall` 😇

hynek:
> Why not just add tox and tox-uv to our deps and run them via uv run? That way we also control the versions.

I find that a bit inception-ish since we then use our lockfile to install tox which then uses the lockfile to install dependencies… I think that could cause problems down the line, especially because we pull tox's dependencies into our lockfile and they're non-trivial: https://www.wheelodex.org/projects/tox/

I suspect this could lead to problems down the road especially around new Python versions etc.

> Also why not just dump tox altogether in favor of a `just testall` 😇

Sure and we install it by adding `rust-just` to our lockfile. 🤪

Tinche:
> Sure and we install it by adding rust-just to our lockfile. 🤪

I've considered this but it doesn't quite work since just calls uv, so using uv to install just violates the layering. Using just to install uv would make more sense but at this point in time that's too avant-garde 😇

All of my thinking that follows is enabled by the fact uv changes how I think about virtual envs. Venvs are cattle now. Nuking a venv to build docs or test on a different Python version is zero cost now.

Our goal should be to have a lean and minimal interface to usual dev tasks. My thinking is that uv needs to be required since it handles everything related to python. Just is just (heh) porcelain on top of lower level plumbing - most people should have it but stubborn folks can just look at the justfile and use uv directly. I would prefer this approach even if uv grew something like npm scripts since just is language agnostic, and I really enjoy that. 

Removing tox from the picture in lieu of `just` completes the story. For example, our current doc building pipeline involves tox creating an environment with the docs extra and then calling sphinx. I understand why this makes sense but if you look at it from a distance it's a clown car solution. Let's just run sphinx with the docs group. If the docs require a particular Python version, no problem, recreating the venv is instant.

hynek:
But you're also calling make from your Justfile my dude. :D

Generally, I'm partial to the argument (as proven by my video), but I'd still need to see an Justfile that that the matrix we do.

As in: versions x tests, versions x mypy… In other projects I also need different sets of optional dependencies.

As far as I can see it, you only measure coverage over one Python version in cattrs?

It also is less cross-platform, since tox doesn't use a shell to execute its stuff. Although this could be solved by using Python as the shell within the Justfile. See also https://just.systems/man/en/shebang-recipes.html and notably: https://just.systems/man/en/python-recipes-with-uv.html

And finally tox for all its faults is kinda a community standard. People know immediately what's happening when they see a tox.ini. But now that debian is finally gonna ship Just, it the combo might get interesting.

hynek:
Oh and one more thing, since we don't have a centralized place for this discussion: I dislike how `uv run -p X` will switch the main venv to X. This is a pretty strong impedance mismatch between regular projects and pseudo-tox. But Astral might be open to accommodate us idk.

hynek:
also JFTR https://github.com/astral-sh/uv/issues/5903 – task runner functionality seems to be very much on the radar. I suspect that might solve the venv switching problem.

Tinche:
> But you're also calling make from your Justfile my dude. :D

That's temporary.

> As far as I can see it, you only measure coverage over one Python version in cattrs?

How would I have 100% coverage if I only measured over one version? 🙈

> It also is less cross-platform, since tox doesn't use a shell to execute its stuff.

Hm, I did wonder about Windows.

> I dislike how uv run -p X will switch the main venv to X.

I'm ok with this, since venvs are disposable cattle. If a test fails on version X, it actually saves me from switching it to X manually. 

> task runner functionality seems to be very much on the radar.

Like I said, I love that just is language-independent. If they added a task runner I wouldn't use it. Unless it used justfiles 😇

Tox is a community standard, it's also a large piece of software that introduces a ton of complexity and can be slow and difficult to debug. Complexity means a higher barrier for contributors. We already have the pieces.

judy2k:
Agree with @Tinche that the benefit of Just is that it's language-independent. I use it for all my projects, whether they are Python, JavaScript, Rust... the major benefit to me is the standardised interface across all those environments, even when in some cases, `just run` simply calls `cargo run`.

judy2k:
(And for the same reason, I think it's fine for Just to call out to Make targets. 😁 )

hynek:
> > As far as I can see it, you only measure coverage over one Python version in cattrs?
> How would I have 100% coverage if I only measured over one version? 🙈

I don't know, I don't see no for loop in https://github.com/python-attrs/cattrs/blob/main/Justfile? I presume you do it in CI, but I'd like a coverage target that actually measures everything. `just cov-all` if you will.
 
> > task runner functionality seems to be very much on the radar.
> Like I said, I love that just is language-independent. If they added a task runner I wouldn't use it. Unless it used justfiles 😇

I don't disagree with that part at all, I just would call those targets from just. But I feel like it would be nice to get a bit more declarative abstractions over Python packaging.

I consider Just a glue (yes, yes judy also glue for make :)) and the less the glue has to know about the stuff underneath the better. 🤓

But again, that's not really the deal-breaker for me here.

I want a good local experience around matrices of facts.
 
> Tox is a community standard, it's also a large piece of software that introduces a ton of complexity and can be slow and difficult to debug. Complexity means a higher barrier for contributors. We already have the pieces.

As said, I'm leaning in the same direction, but I'm not quite convinced they're good enough, yet.

Tinche:
> I don't know, I don't see no for loop in https://github.com/python-attrs/cattrs/blob/main/Justfile? I presume you do it in CI

Yes, it kinda has to be like that for parallelization and accurate reporting.

> I'd like a coverage target that actually measures everything. just cov-all if you will.

I'm down for this. `just test-all`, `just cov-all`.

I'm also thinking of something like `just workon <version>` (or `just sync <version>`). It would just call `uv sync --python <version> --all-groups --all-extras`. Get the main venv set for a specific Python version.

PR #1447, opened 2025-07-27
Move pytest-mypy-plugins tests into mypy jobs

hynek:
Those tests are very slow and break under pytest-xdist.


hynek:
FTR, this lowers the local runtime of `pytest -n auto` down to 3–7s on my notebook (from 30+).

PR #1449, opened 2025-07-29
Allow deep_iterable and deep_mapping take lists/tuples

### LINKED ISSUE #1243: Collections of inner validators for `deep_iterable`/`deep_mapping`
This is a copy of the first issue collected in #1206 to enable separate tracking of the bug.

### The issue

Some of the inner validators accept lists while others still require the explicit use of `and_`. For consistency, I think all of them should ideally be able to handle lists. For the `member_validator`, the change has been in implemented in https://github.com/python-attrs/attrs/pull/925, but `iterable_validator` still struggles with lists. For example, the following piece of code works, but replacing `any_` with a list will make it fail:
```python
@define
class A:
    x: List[str] = field(
        validator=deep_iterable(
            member_validator=[instance_of(str), min_len(1)],
            iterable_validator=and_(instance_of(list), min_len(1)),
        ),
    )
```
For `deep_mapping`, lists/tuples not supported for any of its three arguments.

hynek:
Fixes #1243

doesn't it @AdrianSosic?

### DIFF
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -363,7 +363,8 @@ def deep_iterable(member_validator, iterable_validator=None):
     Args:
         member_validator: Validator(s) to apply to iterable members.
 
-        iterable_validator: Validator(s) to apply to iterable itself (optional).
+        iterable_validator:
+            Validator(s) to apply to iterable itself (optional).
 
     Raises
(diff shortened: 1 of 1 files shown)


AdrianSosic:
LGTM! Thx 🎉

PR #1451, opened 2025-07-29
Allow forward-references on 3.14

Tinche:
Try using stringified annotations when constructing classes.


hynek:
Looks like data classes use `annotationlib.Format.FORWARDREF` instead:

https://github.com/python/cpython/blob/9d3b53c47fab9ebf1f40d6f21b7d1ad391c14cd7/Lib/dataclasses.py#L989-L990

Should that concern us?

PR #1459, opened 2025-08-15
Use sys._clear_type_descriptors() on 3.14 to break slot reference cycle

JelleZijlstra:
Followup from #1446. In the latest 3.14 release candidate, we added a `sys._clear_type_descriptors` function for dataclasses to use. Let's use that function here too so we don't have to reach into the internals.

Note that CPython may end up with a different solution in the future (https://github.com/python/cpython/pull/136966) that would remove the cycle without any need for hacks like this. If that merges, we can add additional conditional logic to skip this logic on 3.15+.

We could also drop the existing manual `__dict__` munging entirely; it depends on whether you care about the behavior on earlier 3.14 betas and rc1.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -870,18 +870,22 @@ class _ClassBuilder:
         }
 
         if PY_3_14_PLUS:
-            # Clean up old dict to avoid leaks.
-            old_cls_dict = self._cls.__dict__ | _deproxier
-            old_cls_dict.pop("__dict__", None)
-            if "__weakref__" in self._cls.__dict__:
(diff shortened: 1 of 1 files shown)


hynek:
Hey Jelle thanks for thinking of us! 🥲 

I don’t care about pre-release working, so feel free to kill the old workaround and I’ll merge the PR once GHA has rc2 — thanks!

JelleZijlstra:
Sounds good! The test suite passes for me locally on 3.14.0rc2 with this version.

hynek (approved):

Issue #1129, opened 2023-04-16
attrs.astuple() leaves performance on the table compared to operator.attrgetter


raphCode:
I played around and found out that `attrs.astuple` is quite slow in comparison to using `operator.attrgetter` with all the field names.
Maybe the performance of `attrs.astuple` can be improved, benefiting all library users? 

Here is an example, save as .ipy file and open with ipython (I couldn't be bothered to fiddle with the `timeit` module manually):
```python
from attrs import define
import attrs
import operator

@define
class A:
    a: int = 1
    b: int = 2
    c: int = 3
    d: int = 4
    e: int = 5
    f: int = 6

def faster_astuple(x):
    return operator.attrgetter(*(f.name for f in attrs.fields(type(x))))(x)

a = A()

print(attrs.astuple(a))
print(faster_astuple(a))

%timeit attrs.astuple(a)
%timeit faster_astuple(a)
```

output on my machine:
```
(1, 2, 3, 4, 5, 6)
(1, 2, 3, 4, 5, 6)
5.6 µs ± 159 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
1.35 µs ± 28.4 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
Speedup is around a factor of four.

ljluestc:
import attr
import operator

@attr.s
class A:
    a: int = 1
    b: int = 2
    c: int = 3
    d: int = 4
    e: int = 5
    f: int = 6

def faster_astuple(x):
    field_names = [f.name for f in attr.fields(type(x))]
    return operator.attrgetter(*field_names)(x)

a = A()

print(attrs.astuple(a, recurse=False))
print(faster_astuple(a))

# Timing the performance
import timeit

attrs_time = timeit.timeit(lambda: attrs.astuple(a, recurse=False), number=100000)
faster_time = timeit.timeit(lambda: faster_astuple(a), number=100000)

print(f"attrs.astuple time: {attrs_time:.2f} seconds")
print(f"faster_astuple time: {faster_time:.2f} seconds")

PR #1463, opened 2025-08-29
Improve performance of asdict in the common case

eendebakpt:
# Summary

We improve performance of `attrs.asdict` by:

* Using a fast check for atomic types (list int, float or str)
* Using `issubclass` instead of `isinstance` for various type checks

Benchmark:
```
asdict: Mean +- std dev: [main] 4.07 us +- 0.41 us -> [pr] 2.48 us +- 0.21 us: 1.64x faster

Benchmark hidden because not significant (1): instance creation

Geometric mean: 1.28x faster
```
<details><summary>Test script</summary>

```
import pyperf

setup = """
from attrs import define, asdict

@define
class Simple:
     i : int
     s : str
     l : list

s = Simple(10, 'hi', [3, 1, 4])

"""

runner = pyperf.Runner()
runner.timeit(name="instance creation", stmt="Simple(10, 'hi', [3, 1, 4])", setup=setup)
runner.timeit(name="asdict", stmt="asdict(s)", setup=setup)
```
</details>

Notes:

* a similar PR was added to cpython dataclasses as well https://github.com/python/cpython/pull/103005
* With a similar approach we can improve performance of `astuple`. That is left to a separate PR though. Also see https://github.com/python-attrs/attrs/issues/1129

# Pull Request Check List

### DIFF
--- a/changelog.d/1463.change.md
+++ b/changelog.d/1463.change.md
@@ -1,0 +1,1 @@
+The performance of `attrs.asdict()` has been improved by 45–260%.


tests/test_funcs.py
```diff
@@ -66,6 +66,20 @@ def test_recurse(self, C, dict_class):
             C(C(1, 2), C(3, 4)), dict_factory=dict_class
         )
 
+    def test_non_atomic_types(self, C):
+        """
+        Test subclasses of atomic types
```
hynek, line 71: This doesn't conform with our docstring standard and proves why it's important :)

What exactly is tested here?  (c.f. https://hynek.me/articles/document-your-tests/ for more context)

eendebakpt: The method name and docstring indeed did not match, I updated the name.

I was trying to test the `else` branch inside the recursive part of the `asdict` implementation. The branch covers non-atomic objects minus some container types. I could add some more tests (e.g. `fraction.Fraction`) if needed.

Issue #1333, opened 2024-08-09
cached_property in slotted classes doesn't work with super from child class


Hnasar:
With attrs 24.2.0 and 23.2.0, if you use `cached_property` and try to override it in a child, and access the parent value, you get an `AttributeError`. We ran into this in our code when converting some classes to use slots.

And I see there was support added in [24.2.0]() for
> Allow super() calls in slotted cached properties

But attrs still doesn't support _"super() calls **of** overridden cached_property's in slotted classes"_:

```python
from functools import cached_property
import attrs

@attrs.define(slots=True)
class Parent:
    @cached_property
    def name(self) -> str:
        return "Alice"

@attrs.define(slots=True)
class Child(Parent):
    @cached_property
    def name(self) -> str:
        return f"Bob (son of {super().name})"

p = Parent()
print(p.name)  # prints Alice
c = Child()
print(c.name)  # error:
```
```
Traceback (most recent call last):
  File "foo.py", line 23, in <module>
    print(c.name)
          ^^^^^^
  File "<attrs generated getattr __main__.Child>", line 6, in __getattr__
  File "foo.py", line 18, in name
    return f"Bob (son of {super().name})"
                          ^^^^^^^^^^^^
AttributeError: 'Child' object has no attribute 'name'
```

## Workarounds:
* disable slots
* don't use cached_property
* don't override the method

## Elaboration:
I see the special handling is described [in the docs](https://www.attrs.org/en/latest/how-does-it-work.html#cached-properties-on-slotted-classes) and I see [in the code](https://github.com/python-attrs/attrs/blob/main/src/attr/_make.py#L481) that it generates a special `__getattr__` that sets the generated value in the slot after being called. Perhaps when constructing a class, attrs should also walk the MRO for `cached_property`s and handle any overridden `cached_property` methods — perhaps create some sort of namespaced/name-manged properties for the parent.

hynek:
I think this is related to #1288? As we've established there, it seems to be a rather complicated issue and all solutions either led the concept of caching ad absurdum (read: performance impact made it questionable to be worth it) or just added another edge case. :|

Issue #1325, opened 2024-08-04
slots + cached_property + sphinx


AdrianSosic:
Hi @hynek, great to see that version `24.1.0` is finally out, have been looking forward to it for quite a while 👍🏼 🙃 

One reason why I was waiting for it is because I wanted to enable slots for our classes that rely on `cached_property`. After upgrading to `24.1.0` and activating slots, everything seemed to work fine ... except for our docs 😕 

We build them with sphinx and the error I'm getting indicates trouble coming from exactly the `cached_property` part. Here is an excerpt from the critical class
```python
@define(frozen=True)
class DiscreteParameter(Parameter, ABC): 

  @cached_property
  @abstractmethod
  def comp_df(self) -> pd.DataFrame: ...
```

for which I get the following error during our sphinx built
```bash
/home/runner/work/baybe/baybe/.tox/docs-py312/lib/python3.12/site-packages/baybe/parameters/base.py:docstring of baybe.parameters.base.DiscreteParameter.__init__:1:<autosummary>:1:py:obj reference target not found: baybe.parameters.base.DiscreteParameter.comp_df
```

I haven't yet tried to pull together a minimal self-contained example, because I first wanted to ask if you already have an idea what could cause the problem – perhaps the `cached_property` integration is simply not yet correct/complete?

All I can say for now is:
* it works when I disable the slots
* it works when I turn the `cached_property` into a regular `property`

so the issue really does seem to be related to the latests `attrs` changes.

Let me know if you need more information!

AdrianSosic:
Hi @hynek, thanks for the quick answer. Seems exactly to be the case!

```python
from functools import cached_property

from attrs import define


@define
class A:
    @property
    def x() -> int:
        """ABC"""
        return 0

    @cached_property
    def y() -> int:
        """DEF"""
        return 0


print(A.x.__doc__)
print(A.y.__doc__)
```

The first one correctly yields "ABC", the second is `None`.

hynek:
Hmm, given how this is implemented (descriptors galore), I'm afraid this is not gonna be easy. I'm open to PRs, tho.

AdrianSosic:
@diabolo-dan: I saw that you added the support in #1200. Do you see any easy solution to the problem?

diabolo-dan:
Hmm, having had a brief look, I don't think there's a solution compatible with the current approach.  The problem is that slots are implemented with a C based descriptor, and that, afaict, you can't set a __doc__ on them.

[...]

PR #1488, opened 2025-12-06
Replace cached_property on slotted classes with a new descriptor instead of writing `__getattr__`

### LINKED ISSUE #1288: `__getattr__` in child class gets called when mixed with `cached_property`

DavidCEllis:
# Summary

This PR reimplements `cached_property` in slotted classes as a slot wrapping descriptor replacing the current generated `__getattr__` implementation.

Accessing the attributes will probably be slightly slower due to going through this descriptor but if the performance is reasonable this will close #1288 #1325 and #1333.

This idea came up again as part of discussion on https://discuss.python.org/t/extensible-cached-property/105188/25. I hadn't realised it would work, but it looks like there was a partial attempt to implement it for attrs before in #1357 

I've implemented it for my own attrs/dataclasses-like [here](https://github.com/DavidCEllis/ducktools-classbuilder/pull/76) (the _SlottedCachedProperty code is mostly identical).

As I saw there were outstanding issues for it and the code is fairly similar I thought you might also find it useful. (Passing all of your existing tests also gave me some confidence that it does work correctly).

---

Changes:
  * Adds a `_SlottedCachedProperty` descriptor class that replaces the slot descriptors after attrs has rebuilt the class.
    * This is very similar to `cached_property` itself, other than reading/writing/deleting a wrapped slot rather than dict entry.
  * Removes the customised `__getattr__` function you were making previously.
    * There are probably some redundant tests that were checking the `__getattr__` behaviour that are now just checking `__getattr__` works in Python.

---

Outstanding issues:

This was already the case for the `__getattr__` implementation but I'm now aware of one difference in behaviour this hasn't yet resolved as it would require changing the logic of when you create a new cached property. There's currently a test marked as xfail for this.

```python
import attrs
import functools

for slotted in [False, True]:
  @attrs.define(slots=slotted)
  class Parent:
      name: str = "Alice"
  
  @attrs.define(slots=slotted)
  class Child(Parent):
      @functools.cached_property
      def name(self):
          return "Bob"
  
  # This isn't to imply that this is good
  # just that it's consistent
  p = Parent()
  c = Child()
  
  assert p.name == "Alice"
  assert c.name == "Alice"
  del c.name
  assert c.name == "Bob"  # True without slots, errors under slots.
```

This is because the cached_property replacement check `attrs` does skips field names that already exist, so `name` is skipped in the check and doesn't get replaced. The descriptor would behave like this, but it's never placed.

# Pull Request Check List


DavidCEllis:
Oof, that's worse than I thought. I have one idea that should make it slightly better, but probably still significantly worse than the direct access it has on main.

DavidCEllis:
Ah, it would have needed to be on main to see the difference. I haven't used codspeed before, sorry for the noise there. Locally it seemed to be significantly faster to construct classes at least. Probably less important than attribute access, but worth noting it's not uniformly worse performance-wise.

hynek:
would you like me to transplant the benchmark to main? i'm a bit busy (just came back from vacation with the obvious results), but i'm open to mechanical work ;)

DavidCEllis:
Sure, that would be great thanks. No rush if you're already swamped.

I tried a few other things to see if they affected performance but it seems it's entirely linked with wrapping the slot in another descriptor.

The only other thing I've tried so far that improved performance at all was replacing the new descriptor with a regular `property` that handles the caching. I don't like that as much as it makes recovering the slot/function more awkward (as you can't add attributes to a property) and it was only a little faster than the custom descriptor.

hynek:
done!

PR #1512, opened 2026-02-19
Put a custom `@property` in place of slotted `@cached_property` for documentation purposes

### LINKED ISSUE #1325: slots + cached_property + sphinx

clayote:
Add a tiny `@property` in place of any deleted `@cached_property` in a rewritten class with `slots=True`. The `@property` does the same thing as its `@cached_property`.

I did this so that `help()` and [Sphinx](https://www.sphinx-doc.org/) can read the docstrings.

# Summary

Fixes #1325

# Pull Request Check List

- n/a Changes or additions to public APIs are reflected in our type stubs (files ending in ``.pyi``).
    - n/a ...and used in the stub test file `typing-examples/baseline.py` or, if necessary, `typing-examples/mypy.py`.
    - n/a If they've been added to `attr/__init__.pyi`, they've *also* been re-imported in `attrs/__init__.pyi`.
    - n/a New functions/classes have to be added to `docs/api.rst` by hand.
    - n/a Changes to the signatures of `@attr.s()` and `@attrs.define()` have to be added by hand too.
    - ??? Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` 
      Is this still a thing with strictly internal functions?
- [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).
      - n/a If something changed that affects both `attrs.define()` and `attr.s()`, you have to add version directives to both.
- n/a Documentation in `.rst` and `.md` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -555,8 +555,11 @@ def _make_cached_property_uncached(original_cached_property_func, cls):
     unique_filename = _generate_unique_filename(
         cls, original_cached_property_func
     )
-    glob = {"original_cached_property": original_cached_property_func,
-            "cached_property_results": _cached_property_results, "NOTHING": NOTHING}
+    glob = {
+        "original_cached_property": original_cached_property_func,
+        "cached_property_results": _cached_property_results,
+        "NOTHING": NOTHING,
+    }
     return _linecache_and_compile("\n".join(lines), unique_filename, glob)[
         name
     ]


clayote:
I'm not super committed to this approach; if #1488 bears fruit, this can be closed.

clayote:
I don't think this approach will ever perform very well.

PR #1513, opened 2026-02-20
Fix validators.disabled() to save/restore state on nesting

bysiber:
## Summary

`validators.disabled()` unconditionally re-enables validators in its `finally` block instead of restoring the previous state. This breaks nested usage — the inner context manager prematurely re-enables validators when it exits.

```python
with disabled():
    with disabled():
        pass
    # BUG: validators are now ENABLED even though we're still in the outer disabled()
    C(x="not_int")  # raises TypeError unexpectedly
```

The fix saves the current state with `get_run_validators()` before disabling, then restores it in `finally`. This is a one-line semantic change.

The same pattern of nesting breakage also applies when `set_run_validators(False)` was called before entering `disabled()` — the context manager would unexpectedly override that.

### DIFF
--- a/src/attr/validators.py
+++ b/src/attr/validators.py
@@ -79,6 +79,7 @@ def disabled():
         This context manager is not thread-safe!
 
     .. versionadded:: 21.3.0
+    .. versionchanged:: 26.1.0 The contextmanager is nestable.
     """
     prev = get_run_validators()
     set_run_validators(False)


hynek:
I'm going to be honest with you: I've spent much more time in the past weeks thinking about how to deal with your pull requests than this is saving time. It was inevitable that my projects get hit by AI slop, so I guess it's nothing personal, but it's kinda putting many things that I've practiced in the past on its head.

In the end, I've decided to finish up the PRs that I find worthwhile, since I don't want to appear petty but I also don't have the nerves to point out things to you that are clearly documented.

If you plan to contribute more in the future, I ask you to follow our processes and heed our [AI policy](https://github.com/python-attrs/attrs/blob/main/.github/AI_POLICY.md). Just deleting the pull request checklist is plain rude and violating half of the items on it too. None of you pull requests have added tests or changelog entries. I'm honestly not sure what your thinking process behind that was but in the future I will close such pull requests without comment.

Thank you.

PR #1515, opened 2026-02-20
Allow on_setattr=setters.NO_OP on frozen classes

bysiber:
## Summary

Using `on_setattr=setters.NO_OP` on individual fields of a frozen class raises `ValueError: Frozen classes can't use on_setattr`, even though `NO_OP` explicitly means "do nothing for this attribute's setattr hook."

```python
@attr.s(frozen=True)
class Frozen:
    x = attr.ib(on_setattr=setters.NO_OP)
# ValueError: Frozen classes can't use on_setattr.
```

This happens because the check `if a.on_setattr is not None` treats `NO_OP` the same as any real hook — `NO_OP` is an `object()` sentinel, not `None`, so it enters the frozen rejection branch.

The fix skips the rejection when `on_setattr` is specifically `NO_OP`, since it's semantically equivalent to having no hook at all. Real hooks like `setters.validate` are still correctly rejected on frozen classes.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -2014,11 +2014,13 @@ def _make_init_script(
         attr_dict[a.name] = a
 
         if a.on_setattr is not None:
-            if frozen is True:
+            if a.on_setattr is setters.NO_OP:
+                pass
+            elif frozen is True:
                 msg = "Frozen classes can't use on_setattr."
                 raise ValueError(msg)
-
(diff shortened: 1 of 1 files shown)


hynek:
this is too convoluted, breaks tests without adding any, and has no changelog. i've added it in https://github.com/python-attrs/attrs/commit/a572c3a4c558a8b9b48cf989fcd956ab1a279439

PR #1519, opened 2026-02-21
Add a Sphinx extension to document cached properties of slots classes

clayote:
# Summary

This is a small extension for [Sphinx](https://www.sphinx-doc.org/), the documentation generator that attrs uses. Where cached properties on slots classes were missing previously (#1325), this extension restores them.

I added a dictionary to slotted class objects, `__attrs_cached_properties__`, where they keep the cached property objects for Sphinx.

I wrapped the `__slots__` tuple itself in a custom `collections.abc.Sequence` that acts exactly like its tuple, but isn't considered one for the purposes of `isinstance()`, because, currently, I have to do that in order to make Sphinx use the extension on slots. I have [a PR for Sphinx](https://github.com/sphinx-doc/sphinx/pull/14316) open to make that unnecessary.

To use the extension, add `"attrs.sphinx_cached_property"` to `extensions` in your Sphinx `conf.py`.

# Pull Request Check List


clayote:
I wrote a test for this, but it won't run in CI currently, because that test requires Sphinx to be installed. That's why the pre-commit check is failing.

clayote:
I need to make sure the `:members:` directive works, too...

clayote:
I haven't a clue what's going on with those `interrogate` failures! It seems to be related to `coverage` somehow? @hynek ?

hynek:
> I haven't a clue what's going on with those `interrogate` failures! It seems to be related to `coverage` somehow? @hynek ?

on a quick look I see two tests without docstrings

clayote:
Checks are passing, I think it's ready @hynek

clayote:
[Looks like a misconfigured test to me](https://github.com/python-attrs/attrs/actions/runs/22286328400/job/64465373469?pr=1519)

sscherfke:
@clayote Your actions create a lot of noise and many notifications. It would be helpful if you did more work offline and and only post updates and questions when really needed. This would make it easier for reviewers/maintainers to help you.

AdrianSosic:
@clayote: feel free to ping me when it's ready for me to test 👌🏼

clayote:
@AdrianSosic Please do

clayote:
Huh, I guess I'd better write documentation for the extension.

It's not part of the API, really? Where should it go...

src/attr/_make.py
```diff
@@ -103,6 +103,34 @@ def __reduce__(self, _none_constructor=type(None), _args=()):  # noqa: B008
         return _none_constructor, _args
 
 
+class _TupleProxy(Sequence):
+    __slots__ = ("_tup",)
```
webknjaz: A docstring must be the first node in the class definition for it to end up in `__doc__`. Otherwise, it's a docstring of the `__slots__` attribute which most tools don't parse as such (I only know of Sphinx). Though, looking at the text, it seems you meant it to be a class docstring...

tests/test_slots.py
```diff
@@ -736,6 +750,91 @@ def f(self):
...
+)
+
+
+@need_sphinx
+def test_sphinx_autodocuments_cached_property():
+    """
+    Sphinx can generate autodocs for cached properties in slots classes
+    """
+    here = Path(__file__).parent
+    with TemporaryDirectory() as td:
```
webknjaz: Why not use the native `tmp_path`?

hynek:
Hey I'm sorry for not turning up earlier, and I'm extra sorry if my innocent question for a test on the other PR lead us to this +340 diff that also contains a whole Sphinx extension.

I will not maintain a Sphinx extension, but if you want to, we can let yours live within the `python-attrs` organization. However, shouldn't it be something that could be fixed in Sphinx proper?

Either way, your changes to the very heart of attrs and its behavior is just too deep and far-reaching for a too narrow of a use-case.

I should've stopped you earlier and I've also amended our contribution guidelines to avoid this situation in the future in #1528.

I'm very sorry, I feel bad about how all this went down.

clayote:
> I will not maintain a Sphinx extension, but if you want to, we can let yours live within the `python-attrs` organization. However, shouldn't it be something that could be fixed in Sphinx proper?

Mostly! I did a little more experimentation, and it turns out that, if [my PR to Sphinx](https://github.com/sphinx-doc/sphinx/pull/14316) is merged, I can write the extension with no changes to `attr._make` at all. Any assistance getting that merged would be appreciated.

There still needs to be a Sphinx extension to reconstruct the `cached_property` object. [The new extension](https://codeberg.org/clayote/sphinx-attrs-cached-property-ext/src/branch/main/sphinx_attrs_cached_property_ext.py) doesn't need to be a part of this organization.

clayote:
Well, that extension does work, but it relies on the current way that cached properties get made on slotted classes. If the [_make_cached_property_getattr](https://github.com/python-attrs/attrs/blob/main/src/attr/_make.py#L507) utility function were to change slightly, the extension would break.

@hynek would you be open to putting the cached property functions in the public API somewhere, perhaps `__attrs_attrs__`?

PR #1541, opened 2026-04-13
Expose converter as a decorator

### LINKED ISSUE #240: Add decorator for convert

LecrisUT:
# Summary

Now that the `Converter` can accept `self`, we can expose it as a decorator similar to `validator`

Follow-up to #1267.
Closes #240

# Pull Request Check List

- [ ] There's **tests** for all new and changed code.


docs/init.md
```diff
@@ -384,6 +384,19 @@ If you need more control over the conversion process, you can wrap the converter
 C(x=410)
 ```
 
+Or as a decorator
```
hynek, line 388:
```suggestion

Or as a decorator
```

docs/init.md
```diff
@@ -384,6 +384,19 @@ If you need more control over the conversion process, you can wrap the converter
...
+```{doctest}
+>>> @define
+... class C:
+...     factor = 5  # not an *attrs* field
```
hynek:
```suggestion
...     factor: ClassVar[int] = 5  # ClassVars are ignored by attrs
```

this is cleaner since it allows you to still have type annotations.

src/attr/_make.py
```diff
@@ -2840,6 +2840,25 @@ def default(self, meth):
 
         return meth
 
+    def converter(self, meth):
+        """
+        Decorator that adds *meth* to the list of converters.
+
+        Returns *meth* unchanged.
+
+        .. versionadded:: TBD
```
hynek:
```suggestion
        .. versionadded:: 26.2.0
```

hynek: let's be optimistic ;)

src/attr/_make.py
```diff
@@ -2840,6 +2840,25 @@ def default(self, meth):
...
+        """
+        decorated_converter = Converter(
+            lambda value, _self, field: meth(_self, field, value),
+            takes_self=True,
+            takes_field=True,
+        )
+        if self._converter is None:
+            self._converter = decorated_converter
+        else:
+            self._converter = pipe(self._converter, decorated_converter)
```
hynek, line 2859:
```suggestion
            self._converter = pipe(self._converter, decorated_converter)

```

src/attr/_make.py
```diff
@@ -2840,6 +2840,25 @@ def default(self, meth):
 
         return meth
 
+    def converter(self, meth):
+        """
+        Decorator that adds *meth* to the list of converters.
```
hynek:
```suggestion
        Decorator that appends *meth* to the list of converters.
```

let's encode clearly that the decorated one goes to the end

hynek (approved):
nice thanks!

for the next time: please link the news fragment to the PR, not the issue. I'll fix it myself this time, no worries.

hynek:
so, who's volunteering to go over to mypy and add it to their plugin? ;)

LecrisUT:
I'm up for it if can find where it is defined. I need to add something similar for my project as well

hynek:
> I'm up for it if can find where it is defined. I need to add something similar for my project as well

ping @Tinche 😇

LecrisUT:
Found it with fun comments 😁
https://github.com/python/mypy/blob/ef03ccd9b31bfbf0c193d6f2ea4b0e33ece8be89/mypy/plugins/attrs.py#L598

I guess the bigger question is if and how to move the plugin inside here

hynek:
huh I think there was an effort/issue about that before?? I'm at a conference rn but shouldn't be too hard to find. There's been a bunch of pro/cons but I guess the main q is if we can trick Ivan and Nikita to contribute here too since they seem to be the most active ones. ;)

https://github.com/python/mypy/commits/ef03ccd9b31bfbf0c193d6f2ea4b0e33ece8be89/mypy/plugins/attrs.py

PR #1542, opened 2026-04-13
Update mypy to 1.20.1

LecrisUT:
# Summary

Noticed that the tests were failing locally but not in CI in my recent PR, and turned out it was mypy not working with the latest version. I don't know what the policy is for supporting older mypy so here is a more minimal patch to make it work with the version in the uv.lock. Fwiw, Fedora's tests do not run the mypy tests.

# Pull Request Check List


hynek (approved):
oh my god it they changed the syntax all again – thanks for the fix!

hynek:
OK, it might be that we have to revert this and talk to the Mypy people. With both 1.20.0 and 1.20.1, the `pytest -k test_mypy` passes for me only once which makes no sense:

CleanShot 2026-04-13 at 19 55 40

Does it work for you reliably?

cc @Tinche

LecrisUT:
> Does it work for you reliably?

That is really weird, yes it works reliably for me, although I did not run it with `uv run`, just directly `pytest -k mypy` :thinking:.

Edit: Ok I've ran them with `uv` as well and it worked on my side. What black magic is going on?

hynek:
what platform are you on?

LecrisUT:
> what platform are you on?

Linux (Fedora). I reckon it's not specific to me and could replicate in a container

hynek:
I mean **I** am the one who has a weirdly behaving Mypy. :) And I'm on macOS.

LecrisUT:
Well, at least in the worst case those test cases can be converted to a regex  🙃

hynek:
looks like I "fixed" it by disabling caching on that test #1546

Issue #1543, opened 2026-04-13
Making third-party extensions a bit easier


LecrisUT:
I am researching how we can expand the `field` with more metadata on our project. The documentation was very helpful for getting started, but I've hit a few places where there is a bit of friction

#### Adding field decorators requires extending `attr._make._CountingAttr`

This is way more complicated then it seems at first glance because:
- the `_CountingAttr` keeps an internal counter for the ordering
- changing the return type of `field` makes it be picked up as a custom `default` instead of an extension of `_CountingAttr`

The initial design I considered was
```python
def field(*, metadata=None, **kwargs: Any) -> Any:
    if not metadata:
        metadata = {}
    metadata.setdefault("my_ext", MyMetadata())
    return MyCountingAttr(attrs.field(**kwargs, metadata=metadata))

class MyCountingAttr(attr._make._CountingAttr):
    def __init__(self, ca):
        for f in ca.__slots__:
            setattr(self, f, getattr(ca, f))
        attr._make._CountingAttr.cls_counter -= 1
        self.counter -= 1

    def new_decorator(self, meth):
        my_metadata: TmtAttrsMetadata = self.metadata["my_ext"]
        my_metadata.func = meth
        return meth

@attrs.define
class MyMetadata:
    func: Callable[..., Any] | None = None

@attrs.define
class Example:
    x: field()

    @x.new_decorator
    def _x_my_func(self):
        pass
```
but this fails when it gets to
https://github.com/python-attrs/attrs/blob/f0e420b71af1a5cd5dbf43efd1a42aece6ce8f4e/src/attr/_make.py#L403-L407
Could not find a way around it besides changing that to a `issubclass` check

#### Many useful internal functions are not exposed

For example `attr._make._determine_whether_to_implement` would be really nice to not have to re-implement, especially when it comes to the correct handling of `classmethod`. Admittedly my usage there is quite weird (injecting some methods/classmethods based on inputs to the `@define` decorator) and it could be better handled with inheritance/metaclass, but it's the first example of the internal helpers that I've found in my code.

LecrisUT:
Quite related, afaiu `_CountingAttr` is the reason we must use `field` to allow for decorator approaches. But with `Annotated` also becoming a thing, would there be a way to move this logic. I don't know if it's a limitation in the python parser that would not be able to resolve where to get the decorator from. Probably there is already research on this in an issue?

hynek:
[...]

PR #1544, opened 2026-04-13
Check for _CountingAttr subclasses

LecrisUT:
# Summary

The main purpose for this is to allow custom fields to extend the field decorators. This does technically enable that by subclassing the private class and adding the appropriate methods as in #1543.

I do not believe this is the best approach for handling #1543, it's just the quickest way to unblock it. The class should probably become something more public. If there are other ideas happy to explore them.

# Pull Request Check List

- [ ] There's **tests** for all new and changed code.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -403,7 +403,7 @@ def _transform_attrs(
         ca_names = {
             name
             for name, attr in cd.items()
-            if attr.__class__ is _CountingAttr
+            if issubclass(attr.__class__, _CountingAttr)
         }
         ca_list = []
         annot_names = set()
@@ -413,7 +413,7 @@ def _transform_attrs(
             annot_names.add(attr_name)
             a = cd.get(attr_name, NOTHING)
 
-            if a.__class__ is not _CountingAttr:
+            if not issubclass(a.__class__, _CountingAttr):
                 a = attrib(a)
             ca_list.append((attr_name, a))
 
@@ -431,7 +431,7 @@ def _transform_attrs(
             (
                 (name, attr)
                 for name, attr in cd.items()
-                if attr.__class__ is _CountingAttr
+                if issubclass(attr.__class__, _CountingAttr)
             ),
             key=lambda e: e[1].counter,
         )


hynek:
While I'm always super open to expand our… expandability… this needs more meat for the motivation. Like _concrete_ examples that it enables – which could go straight into, probably, expanding.md?

In other words it seems a bit speculative rn in whether it actually achieves what you need/want in the second/third step. So I would like to see some proof (and docs) to show that.

LecrisUT:
Luckily I did make a test repo of my experimentations 😉 https://github.com/lecrisut/attrs-playground

hynek:
> Luckily I did make a test repo of my experimentations 😉 https://github.com/lecrisut/attrs-playground

does any of the examples use this? because the readme just says it's not supported. :)

as said if this helps you with something concrete I'm open to it. long-term we should get rid of _CountingAttr, since it comes with a bunch of surprising behavior. Unfortunately, there's probably a non-trivial amount of software depending on it it, tho. So I wouldn't want to block you with that.

LecrisUT:
> does any of the examples use this? because the readme just says it's not supported. :)

Yeah, the readme is more intended for my colleagues. You can try out the example with and without the patch to see the difference. It is required, otherwise it would not detect the field [^1] as a `attrs.field `, instead it treats it as an annotation and the metadata is gone.



[^1]: https://github.com/LecrisUT/attrs-playground/blob/c95acef82de6d04c5b7e4be000a3623b4a658486/container.py#L68

> long-term we should get rid of _CountingAttr, since it comes with a bunch of surprising behavior.

Right now it is only this idea that would depend on it, so there is no immediate rush. If there are ideas of what can take its place, then I'm all up for pivoting the design.

At the core of this I think the issue is that `_CountingAttr` has double-duty to guarantee the ordering of the attributes and to create the entry-point for the field decorators. The former could be dropped for sure, but the latter would still have to be around in one way or another. How about making a specific subclass to contain the latter to keep it "safe" for users to subclass from, and find a way to decouple these and drop the need for the former.

hynek:
No, I do not really have any ideas. I just want to get rid of the counting that we needed before Python 3.6. I'm mostly saying that I'm not married to anything about _CountingAttr.

I guess your subclassing idea makes sense; the q is just what the new class should do except being public?

LecrisUT:
> I guess your subclassing idea makes sense; the q is just what the new class should do except being public?

I will try to make a PR to highlight what will be in the subclass and why it would need to be public with an xfail test, and we can go on from there.

PR #1556, opened 2026-05-10
Validate init aliases before generated code compilation

uwezkhan:
attrs generates init methods dynamically using field aliases. currently invalid aliases can make it all the way into generated source code and fail later during compilation with confusing SyntaxError exceptions.

this change adds validation for init aliases before code generation happens. it checks for invalid identifiers, python keywords, reserved names like self  and unicode normalization collisions that can result in duplicate generated parameter names.

validation happens after alias resolution and field transformers run so all final aliases are checked consistently before generating the initializer.

added regression tests for:

invalid identifiers
keyword aliases
self shadowing
unicode normalization collisions
init=False behavior


uwezkhan:
can i get a review on this PR?

hynek:

**I will not finish this PR for your LLM bot.**

Given the current volume of LLM-assisted contributions, we cannot spend maintainer time reviewing half-finished pull requests that do not follow the project’s contribution process.

If anything is unclear about our CONTRIBUTING.md, AI_POLICY.md, or PULL_REQUEST_TEMPLATE.md, please reach out.

PR #1561, opened 2026-05-18
Add input type validation to max_len and min_len validators to preven…

uwezkhan:
The max_len and min_len validators called len() on input values without checking if they support the len() function. This causes cryptic TypeError: object of type 'X' has no len() when non-sized objects are passed. Example of the Problem: @attr.s class Example: items = attr.ib(validator=max_len(10)) # This raises confusing error: # TypeError: object of type 'generator' has no len() Example(items=(x for x in range(5))) Why This Matters: Reliability - Prevents confusing TypeErrors hard to debug Validation - Catches invalid inputs early with clear messages Consistency - Follows validation patterns in other validators Solution Added explicit type validation using isinstance(value, collections.abc.Sized): _MaxLengthValidator validates input is Sized before calling len() _MinLengthValidator validates input is Sized before calling len() Raises clear TypeError: '<attr_name>' must be a sized object (got <value> that is a <type>).


uwezkhan:
any update?

PR #1565, opened 2026-05-28
clarify frozen class immutability behavior with factories

sahvx655-wq:
This pull request clarifies the behavior of `frozen=True` classes when factory defaults return mutable objects that are shared. 
In `attrs`, `frozen=True` provides shallow immutability (preventing attribute reassignment), but it does not make the contents of mutable objects (like lists or dicts) immutable. If a factory returns a shared mutable object, all instances of the frozen class share that object, and external mutations will affect all frozen instances.
This PR adds documentation and a doctest to clarify this behavior.
Changes
Added a new **Factories and Frozen Classes** subsection under the **Defaults** section in `docs/init.md` to explain shallow immutability and document this behavior with a clear doctest example.
Key Points
`frozen=True` does not prevent changes inside mutable objects.
Factory defaults that return shared mutable references lead to shared-state behavior across frozen instances.

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -151,6 +151,28 @@ Similar to the [common gotcha with mutable default arguments](https://docs.pytho
 
 This is why *attrs* comes with factory options.
 
+### Factories and Frozen Classes
+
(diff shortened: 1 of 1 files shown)


sahvx655-wq:
gentle ping

PR #1567, opened 2026-05-28
Fix intermediate __setattr__ handling in mixed inheritance chains

sahvx655-wq:
This PR fixes a correctness issue in attrs where __setattr__ could be incorrectly overridden when using intermediate base classes that define a custom __setattr__.

Previously, only immediate base classes were checked, which caused custom __setattr__ implementations in deeper inheritance hierarchies to be ignored or replaced unexpectedly.

This change introduces MRO-based resolution for both slotted and dict-based classes, ensuring that intermediate __setattr__ implementations are properly respected throughout the inheritance chain.

Adds and updates tests covering multiple inheritance and edge-case hierarchies to validate the fix. is this matches for this title Fix intermediate __setattr__ handling in mixed inheritance chains


sahvx655-wq:
any update?

PR #1571, opened 2026-06-04
Add `ne` validator

jamesbeith:
# Summary

This change adds a validator that raises `ValueError` if the
initializer is called with a number equal to *val*.

An example use case. I want a field that is a signed `int` of any
value except for zero. I could do the following.

```python
attrs.field(
    validator=attrs.validators.or_(
        attrs.validators.le(-1),
        attrs.validators.ge(1),
    )
)
```

With this proposed `ne` validator I could do.

```python
attrs.field(validator=attrs.validators.ne(0))
```

# Pull Request Check List


hynek (approved):
thanks!

PR #1573, opened 2026-06-25
Fix pickling auto_exc exceptions with kw_only fields

### LINKED ISSUE #734: pickling and kw_only

lin-hongkuan:
## Summary

Fixes #734.

This adds an attrs-generated `__reduce__` for `auto_exc=True` exception classes that have `init=True` keyword-only fields. The generated reducer reconstructs the exception without calling `__init__`, then restores the normal pickle state, so required keyword-only fields work across plain, frozen, slotted, and frozen-slotted exception classes.

The generation is intentionally narrow: it only applies to attrs-generated exception `__init__` methods with keyword-only init fields, leaves positional-only exceptions on `BaseException.__reduce__`, and does not override custom `__reduce__` implementations, including inherited ones.

## Tests

- `PYTHONPATH=src python -m pytest -q`
- `PYTHONPATH=src python -m ruff check src/attr/_make.py tests/test_functional.py`
- `PYTHONPATH=src python -m ruff format --check src/attr/_make.py tests/test_functional.py`
- `git diff --check`

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -39,6 +39,8 @@ from .exceptions import (
 
 # This is used at least twice, so cache it here.
 _OBJ_SETATTR = object.__setattr__
+_BASE_EXCEPTION_REDUCE = BaseException.__dict__.get("__reduce__")
+_BASE_EXCEPTION_SETSTATE = BaseException.__dict__.get("__setstate__")
 _INIT_FACTORY_PAT = "__attr_factory_%s"
 _CLASSVAR_PREFIXES = (
     "typing.ClassVar",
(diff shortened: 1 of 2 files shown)


hynek:

Given the current volume of LLM-assisted contributions, we cannot spend maintainer time reviewing pull requests that do not follow the project’s contribution process.

Before submitting further contributions, please read and follow this project’s CONTRIBUTING.md and AI_POLICY.md.

Thank you for your understanding!

PR #1584, opened 2026-07-07
Match filter attributes by identity instead of equality

### LINKED ISSUE #864: asdict filter exclude list matches on name instead of attribute
Hi, first of all, love the project and use it nearly every day for my job. 

I have come across some behaviour that is unexpected. 
Please consider this minimal worked example:
```
import attr
from typing import List

@attr.s(auto_attribs=True)
class Cop:
    my_field: str
    repeated_field_name: str


@attr.s(auto_attribs=True)
class Robber:
    my_other_field: str
    repeated_field_name: str


@attr.s(auto_attribs=True)
class Payload:
    cops: List[Cop]
    robbers: List[Robber]


if __name__ == "__main__":
    cop = Cop("abc", "shared1")
    robber = Robber("def", "shared2")
    payload = Payload([cop], [robber])

    print(attr.asdict(payload, recurse=True))
    """
    {
        'cops': [{'my_field': 'abc', 'repeated_field_name': 'shared1'}], 
        'robbers': [{'my_other_field': 'def', 'repeated_field_name': 'shared2'}]
    }
    """
    print(
        attr.asdict(
            payload, 
            recurse=True, 
            filter=attr.filters.exclude(attr.fields(Cop).repeated_field_name)
        )
    )
    """
    {
        'cops': [{'my_field': 'abc'}], 
        'robbers': [{'my_other_field': 'def'}]
    }
    """

```

When the asdict `filter` argument is passed a variadic collection of attributes (available from `attr.fields`), and you have a range of attrs models that are recursively `asdict`-ified that may share a field name (in this example `repeated_field_name`) then the returned dictionary excludes this attribute for ALL classes, not just the one passed to fields. 
This is understandable to a degree given `assert attr.fields(Cop).repeated_field_name == attr.fields(Robber).repeated_field_name` holds, but still a little unexpected given the method explicitly does not accept field names as strings.

As it happens, this behaviour is what I wanted 😄 though again I found it unexpected and wanted to confirm it is not a bug before I move forward with the implementation. 

If this is a bug, a suggestion would be to permit string field names

PranavMishra28:
## What was broken

`attrs.filters.include` / `exclude` accept types, field-name strings, and `Attribute` instances. The `Attribute` instances were stored in a `frozenset` and matched with `attribute in attrs`, i.e. by **equality**. But `Attribute.__eq__`/`__hash__` ignore the owning class, so two identically-configured same-named attributes on *different* classes compare equal:

```python
@attr.s
class Robber:
    repeated = attr.ib(default=1)
    only_robber = attr.ib(default=2)

@attr.s
class Cop:
    repeated = attr.ib(default=1)   # identically configured
    only_cop = attr.ib(default=4)

attr.asdict(Cop(), filter=attr.filters.exclude(attr.fields(Robber).repeated))
# -> {'only_cop': 4}   # Cop.repeated wrongly dropped
```

Excluding (or including) one class's attribute silently affects an identically-configured attribute of an unrelated class.

## Why it matters

This is the behavior tracked in #864, which is labeled `Bug`. As noted there, `Attribute` instances are independent of the classes they're defined on, so equality-based matching "is absolutely not what people would expect." `attr.fields(A).x` and `attr.fields(B).x` are distinct objects, so identity is the correct test.

## What the fix does

`_split_what` keeps the passed attributes in a `tuple` (instead of a `frozenset`), and `include`/`exclude` match them with `any(attribute is a for a in attrs)` — by identity. Since `fields()` returns the same cached `Attribute` object for a given class, the intended attribute still matches, while an equal-but-distinct attribute on another class no longer does.

## What it deliberately does not change

- Type and field-name-string matching are untouched. The cross-class "drop this field name from many classes" use case raised in #864 is already served by passing the **name string** (`exclude("repeated")`), which matches by name across classes.
- Public `include`/`exclude` signatures and the `.pyi` stub are unchanged.

## Note on compatibility

This narrows `Attribute` matching from equality to identity. Anyone who (perhaps unknowingly) relied on the equal-but-distinct cross-class match would see a behavior change; per the issue this was considered a bug rather than intended behavior, so I filed it under `change`. Happy to reclassify as `breaking` if you'd prefer.

## Testing

- New identity regression tests for both `include` and `exclude` (equal-but-distinct attribute on another class is not matched); updated the `_split_what` test for the tuple.
- Full suite green: `1383 passed, 9 skipped, 1 xfailed`; `ruff check` and `ruff format --check` clean. Added `changelog.d/864.change.md`.

Fixes #864

Developed with Claude Code; reviewed and tested by Pranav before marking ready for review.

### DIFF
--- a/changelog.d/864.change.md
+++ b/changelog.d/864.change.md
@@ -1,0 +1,1 @@
+`attrs.filters.include` and `attrs.filters.exclude` now match `attrs.Attribute` instances by identity instead of equality, so filtering by an attribute of one class no longer also matches an identically-configured attribute of a different class.
--- a/src/attr/filters.py
+++ b/src/attr/filters.py
@@ -9,12 +9,18 @@ from ._make import Attribute
 
 def _split_what(what):
     """
-    Returns a tuple of `frozenset`s of classes and attributes.
+    Returns a `frozenset` of classes, a `frozenset` of names, and a `tuple` of
+    attributes.
+
+    Attributes are kept as a tuple and matched by identity rather than put in a
+    `frozenset`: `Attribute` equality ignores the owning class, so two
+    identically-configured attributes on different classes compare equal and
+    set membership would match the wrong one (see #864).
     """
     return (
         frozenset(cls for cls in what if isinstance(cls, type)),
         frozenset(cls for cls in what if isinstance(cls, str)),
-        frozenset(cls for cls in what if isinstance(cls, Attribute)),
+        tuple(cls for cls in what if isinstance(cls, Attribute)),
     )
 
 
@@ -39,7 +45,7 @@ def include(*what):
         return (
             value.__class__ in cls
             or attribute.name in names
-            or attribute in attrs
+            or any(attribute is a for a in attrs)
         )
 
     return include_
@@ -66,7 +72,7 @@ def exclude(*what):
         return not (
             value.__class__ in cls
             or attribute.name in names
-            or attribute in attrs
+            or any(attribute is a for a in attrs)
         )
 
     return exclude_
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -30,7 +30,7 @@ class TestSplitWhat:
         assert (
             frozenset((int, str)),
             frozenset(("abcd", "123")),
-            frozenset((fields(C).a,)),
+            (fields(C).a,),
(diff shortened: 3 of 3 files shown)


PranavMishra28:
just wanted to highlight that this is ready for a look whenever you have a moment. It directly fixes the equality bug in #864 and all 19 checks are passing green.

PR #1588, opened 2026-07-11
Add typing examples for Pyrefly's new attrs support

hynek:
This is extremely exciting! ✨


hynek:
reminder to myself: merge together with #1594 once Pyrefly 1.2.0 final is out.

PR #1592, opened 2026-07-18
Make on_setattr hooks accept generators

vladfedoriuk:
# Summary

# Pull Request Checklist


changelog.d/1592.change.md
```diff
@@ -0,0 +1,2 @@
+`attrs.setters.pipe()` now rejects generator functions.
```
webknjaz: I think this should be linkable like so
```suggestion
{func}`attrs.setters.pipe` now rejects generator functions.
```
*(https://webknjaz.github.io/intersphinx-untangled/www.attrs.org/)*

hynek: we don’t make our symbols linkable in the changelog such that CHANGELOG.md is readable by itself

hynek:
as a quick preamble: there is no such thing as 25.5.0. The current version is 26.1.0 so the next is 26.2.0.

hynek:
@KRRT7 any idea if anything could be further optimized/laziied? I’m not sure how much I trust the benchmark given that none of the new code runs on import? 🤔

KRRT7:
@hynek like I was saying before -X importtime (or importlens, but I haven't open sourced it yet, I should do that now) is far more accurate because with wall time you're also measuring Python's startup overhead and probably AVs / security programs (though that probably doesn't apply here). I'll take a look

KRRT7:
@hynek I've optimized it a bit but looks like you've enabled the anti-slop features and since I already have a PR open, can't open another

KRRT7 (commented):
bad ai

KRRT7 (commented):
bad ai

KRRT7:
sorry for the spam here - was trying to get my AI to do the reviews since I'm lazy, let me just push the full branch for now

the work should be here, feel free to port.
https://github.com/KRRT7/attrs/tree/work-on-1437-on-setattr-as-generator

hynek:
Honestly, I think I'll just drop the check and see if it's enough. We'll document the invariant and call it a day. No need to add a ton of complexity for babysitting our users.

hynek:
Did the trick! 🥳

hynek (approved):

KRRT7:
@hynek I accidentally opened my PR to their fork: https://github.com/vladfedoriuk/attrs/pull/1
do you want me to bring it over? the changes are minimal but do shave off import time

hynek:
since the codspeed has proven itself (btw if you come up with ways to measure differently go for it; but the way we measure are dedicated on-metal ARM workers) I’m open to further optimizations, but some of it is kinda noisy so could we take it piecemeal? Some of it is kinda noisy so we should weight it against the upsides.

BTW what does your un-open-sourced-yet tool say about importing dataclasses and pydantic?

PR #1593, opened 2026-07-19
Fix ClassVar forward reference detection

### LINKED ISSUE #1575: attrs doesn't recognize ClassVar if not imported in Python 3.14
The following code

```python
from typing import TYPE_CHECKING

import attrs

if TYPE_CHECKING:
    from typing import ClassVar

@attrs.define
class T:
    t: ClassVar[str]

T()
```

results in:
```python
  File "/tmp/test.py", line 12, in <module>
    T()
    ~^^
TypeError: T.__init__() missing 1 required positional argument: 't'
```

Since the `if TYPE_CHECKING` block doesn't run, `ClassVar` isn't defined and attrs can't evaluate the ForwardRef or doesn't recognize `ClassVar`:

```python
>>> attrs.fields_dict(T)["t"]
Attribute(name='t', default=NOTHING, validator=None, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type=ForwardRef('ClassVar[str]', is_class=True, owner=<class '__main__.T'>), converter=None, kw_only=False, inherited=False, on_setattr=None, alias='t')
```

Before Python 3.14, `T`'s definition would only run if you used a string annotation or `from __future__ import annotations` (which if used with 3.14, the above code runs fine), but with the new annotations behavior the class definition runs fine and this error appears. This is actually issue for dataclasses too, as I get the same error with them.

I'm not sure if this is fixable, but decided to report it, because it catches you by surprise

wanxiankai:
# Summary

Fixes #1575.

`_is_class_var()` now unwraps `ForwardRef.__forward_arg__` before applying the existing string-prefix checks.
This lets `attrs` recognize Python 3.14-style forward references such as `ForwardRef("ClassVar[str]", is_class=True)` as class variables when `ClassVar` was only imported under `TYPE_CHECKING`.

Tests cover both the direct `_is_class_var()` case and the Python 3.14 `attrs.define` behavior from the issue.

# Pull Request Checklist


hynek (approved):
cool, thanks!

PR #1598, opened 2026-07-30
Fix generated __init__ forward-ref annotations under get_annotations / 3.14

### LINKED ISSUE #1596: Incorrect annotation for forward reference in generated constructor
When a class is decorated with `@define` or `@frozen`, and an attribute references a type declared later in the file, the generated `__init__` method contains a type annotation that is different from the one on a hand-written `__init__` method.

In Python 3.14 without `from __future__ import annotations`:
```python
import inspect

from attrs import define


class Works:
    def __init__(self, foo: Foo) -> None:
        self._foo = foo


@define
class DoesNotWork:
    _foo: Foo


class Foo:
    pass


# prints `(foo: __main__.Foo) -> None`
print(inspect.signature(Works))

# prints `(foo: Foo) -> None` (a ForwardRef)
print(inspect.signature(DoesNotWork))
```

With `from __future__ import annotations` (tested on both 3.14 and 3.13 - same result):
```python

import inspect

from attrs import define


class Works:
    def __init__(self, foo: Foo) -> None:
        self._foo = foo


@define
class DoesNotWork:
    _foo: Foo


class Foo:
    pass


# prints `(foo: __main__.Foo) -> None`
print(inspect.signature(Works, eval_str=True))

# crashes with `NameError: name 'Foo' is not defined`
print(inspect.signature(DoesNotWork, eval_str=True))

MohammedAnasNathani:
# Summary

Fixes #1596.

Generated `__init__` / `__attrs_init__` methods no longer diverge from hand-written constructors when:

1. A field uses a **late forward reference** (the name is only bound *after* the class body), and
2. Callers introspect with `inspect.signature(..., eval_str=True)` / `annotationlib.get_annotations(..., format=VALUE)`.

### Root causes

There were two independent problems:

**1. Snapshot globals (all supported Pythons)**  
`_make_init_script` used to copy `sys.modules[cls.__module__].__dict__` into the generated function's globals at class-creation time. Names defined later (e.g. `Foo` after `@define class C: x: Foo`) were therefore missing from `__init__.__globals__`, so `eval_str=True` raised `NameError`. Hand-written methods keep a live module `__dict__` as globals, so they succeed.

**2. Stale static annotations (Python 3.14+)**  
attrs set `init.__annotations__` to a frozen map of `ForwardRef` objects captured at class create. Setting `__annotations__` clears PEP 649 `__annotate__`. Hand-written methods (and dataclasses) keep an `__annotate__` that re-reads live class annotations, so VALUE-format evaluation can see the real class once it exists.

### Fix

- Compile `__init__` / `__attrs_init__` via a small factory so helpers (`NOTHING`, converters, validators, …) are **closure freevars**, while `__globals__` is the **live module `__dict__`** (same model as dataclasses / hand-written code). No more module snapshot.
- On 3.14+, attach **`__annotate__` only** (do not set `__annotations__`). The annotate function re-fetches field types from the class via a `field_arg_map` and reuses static entries for `return` / converter types.
- When building slots classes, also rewrite cells on generated **`__annotate__`** callables (and property fget/fset/fdel) so they close over the **final** class, not the pre-slots original.

### Tests

- `tests/test_forward_references.py` (3.14+): signatures match hand-written classes, VALUE format resolves to the real type, live globals, slots annotate-cell rewrite, converters stay static.
- `tests/test_annotations.py`: 3.13+ regression with a real `ModuleType` + `from __future__ import annotations` late ref.
- `tests/test_dunders.py`: internal `_add_init` helper updated for the new 4-tuple + compile path; init `co_filename` is method-specific by design.

Verified locally on 3.13 (full suite green aside from pre-existing packaging version parse on dirty trees) and 3.14 (manual `#1596` repro + targeted checks; system 3.14 lacks a working pytest install here).

# Pull Request Checklist

### DIFF
--- a/changelog.d/1596.change.md
+++ b/changelog.d/1596.change.md
@@ -1,0 +1,4 @@
+Generated ``__init__`` methods now resolve late forward references the same way as hand-written constructors.
+
+Previously, *attrs* snapshotted the defining module's globals when building ``__init__``, so names defined *after* the class body were missing from ``__init__.__globals__``.
+On Python 3.14+, annotations are now attached via a PEP 649 ``__annotate__`` that re-reads live class annotations instead of a static dict of stale ``ForwardRef`` objects, and slot classes rewrite ``__annotate__`` closure cells to the final class.
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -25,6 +25,7 @@ from ._compat import (
     PY_3_10_PLUS,
     PY_3_11_PLUS,
     PY_3_13_PLUS,
+    PY_3_14_PLUS,
     _AnnotationExtractor,
     _get_annotations,
     get_generic_base,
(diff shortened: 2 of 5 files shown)


MohammedAnasNathani:
### CodSpeed class-creation note (−13.4% on create benches)

CodSpeed’s bot flags `test_create_simple_class` / `test_create_simple_class_make_class` at about **−13%** vs main, with a **different runtime environments** warning on the comparison. The CodSpeed *check* itself is **NEUTRAL / skipping** (not a required failure); all required CI is green.

**What the cost is:** this fix stops snapshotting `sys.modules[cls.__module__].__dict__` into generated `__init__` / `__attrs_init__` globals (the root of #1596). Helpers stay as **closure freevars** from a small factory; `__globals__` is the **live module dict** so late names resolve under `inspect.get_annotations(..., eval_str=True)` / `signature(eval_str=True)` the same way hand-written constructors do. That is one extra `compile` + factory call at class-creation time versus main’s flat joint snippet compile with a frozen globals copy.

**What it is not:** instance construction is unchanged. This is class-*definition* microbench only.

**What I tried and rejected** (so we don’t ship a “faster” wrong fix):

- Dict-subclass / proxy “live globals” that falls through to the module — matches snapshot speed, but on 3.13 `inspect.get_annotations(..., eval_str=True)` and `inspect.signature(..., eval_str=True)` still `NameError` (they need a real module `__dict__`, not a proxy). Also breaks the intentional `__init__.__globals__ is module.__dict__` guarantee.
- Temporarily stuffing helpers into the module dict — races / name collisions / pollution.
- Folding the factory into the joint `_eval_snippets` compile (one `compile` for all dunders) — local measurement did **not** close the gap vs main (compile of the larger nested script still dominates).

Happy to take review direction if maintainers prefer a different tradeoff; I believe the live-globals model is the correct fix for #1596 and the class-create cost is the honest price of that correctness.

PR #1602, opened 2026-07-31
Cache generated method compilation

KRRT7:
## Summary

Cache compiled generated method scripts by script shape to avoid recompiling identical methods for repeated `attrs` class definitions.

- Keep the cache bounded at 128 entries.
- Preserve class-specific traceback and debugger filenames by rewriting cached code objects recursively.
- Compile cache misses directly with their real filename, avoiding a regression for unique class shapes.

### DIFF
--- a/src/attr/_make.py
+++ b/src/attr/_make.py
@@ -223,8 +223,34 @@ def _compile_and_eval(
     Evaluate the script with the given global (globs) and local (locs)
     variables.
     """
-    bytecode = compile(script, filename, "exec")
+    bytecode = _compiled_script_cache.pop(script, None)
+    if bytecode is None:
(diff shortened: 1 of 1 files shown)


hynek:
I think this is an artificial optimization. Unless I'm missing something, it only works when someone keep creating (not instantiating) the same class over and over again. We do that in benchmarks, but that's not something regular codes does. On the flip side, this creates a cache that every class pays for.

KRRT7:
> it only works when someone keep creating (not instantiating) the same class over and over again.

that exactly was my use case, I'm sure I'm not the only one doing these sort of things, and the cache perf is negligible imo, but can close.

hynek:
Yeah sorry, I don’t think that’s worth it considering big picture.

PR #1605, opened 2026-08-02
Increase laziness

hynek:
OK I think I've ported the big ones that don't involve re-implementing the stdlib – WDYT @KRRT7? According to Tuna we're just _slightly_ over dataclasses (0.13 vs 0.14 on my computer) which I wish we were better, but I guess features have their price.

I wish we put AttrsInstance and NothingType into a separate module so we could save ourselves the typing import that also implies functools (which is why a lazy partial makes no diff).

### DIFF
--- a/bench/test_benchmarks.py
+++ b/bench/test_benchmarks.py
@@ -11,6 +11,8 @@ import pytest
 
 import attrs
 
+from attrs import asdict, astuple
+
 
 pytestmark = pytest.mark.benchmark()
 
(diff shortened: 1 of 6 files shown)


hynek:
hm looks like no measurable difference in import time but mysterious runtime performance regression. I think we've hit the sweet spot already.

hynek:
To make it more concrete, locally `uv run python -X importtime -c 'import attr' 2>import.log && uvx tuna import.log` gives me 0.14–0.16 with this PR and 0.13–0.16 without. They both tend to land on 0.14 or 0.15 tho so the real-life difference is less than a millisecond.

PR #1608, opened 2026-08-03
Add tox to lockfile

hynek:
Better reproducibility everywhere.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@
 [tox]
 requires =
     tox>=4.51.0
-    tox-uv
+    tox-uv-bare
 env_list =
     lint-{ruff,pre-commit},
     3.1{0-5}-tests,
@@ -98,14 +98,16 @@ commands =
         docs
 
 [testenv:docs-sponsors]
-runner = uv-venv-runner
-package = skip
 description = Ensure sponsor logos are up to date.
+runner = uv-venv-runner
+base_python_file = .python-version-default
+package = skip
 deps = cogapp
 commands = cog -rP README.md docs/index.md
 
 
 [testenv:lint-{ruff,pre-commit}]
+base_python_file = .python-version-default
 dependency_groups = lint
 package = skip
 commands =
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -3,7 +3,7 @@
 [tox]
 requires =
     tox>=4.51.0
-    tox-uv
+    tox-uv-bare
```
webknjaz, line 6: Wait, what? Is it published under two names now? 🤔

tox.ini
```diff
@@ -3,7 +3,7 @@
 [tox]
 requires =
```
webknjaz, line 4: This looks like a use case for having tox to implement referencing this section from dep groups to dedupe the list..

cc @gaborbernat

tox.ini
hynek: Bare doesn’t install uv as a dependency

tox.ini
hynek: In this case it’s more of a belts-and-suspenders thing since _we_ always run tox using uv. Maybe should be removed to fail loudly? Idk

webknjaz: Well, my understanding is that this is for people that have `tox` as a primary pre-existing entry-point (and not `uv`) — `tox` then makes a managed venv under `.tox/.tox/` and installs all the goodies + a newer version of itself. So to me this sounds like a more logical hook point for sticking automatic provisioning. I'm wondering, though, if tox's API allows plugins to extend this list — this would make for a powerful mechanism of setting up centralized ecosystem-centric infras.

I think this might also be in use by downstreams. But we'd have to ask @hroncok @befeleme @encukou @mgorny etc.

encukou: The Fedora use case is verifying if things work in a given pre-existing environment (without PyPI wheels & without internet access). `tox` doesn't make a venv; instead a plugin checks if dependencies are installed. If any are missing it reports them, and gets re-run in a container that includes them.

For that, it doesn't really matter if `uv` gets *installed*, as long as it doesn't try downloading/provisioning anything :)

Anyway, removing `tox.requires` seems OK.

PR #1609, opened 2026-08-04
ci: build docs with default Python, too

hynek:


tox.ini
```diff
@@ -76,8 +76,7 @@ commands = pytest --codspeed -n auto bench/test_benchmarks.py
 
 
 [testenv:docs-{build,doctests,linkcheck}]
-# Canonical version for everything documentation-related.
-base_python = 3.14
+base_python_file = .python-version-default
```
webknjaz, line 79: cc @sirosen I wonder if this is something that could be useful in a shared tox plugin 🤔

PR #1610, opened 2026-08-05
Add `dataclass_compatible` flag

leontrolski:
# Summary

Short of having some [stdlib fields interface](https://discuss.python.org/t/add-a-fields-interface/108364), IMHO the de facto standard for marking a class as "some non-opaque class that you can iterate over the fields of" is the presence of `__dataclass_fields__`.

One example of this de facto standard (and the motivation for this PR) is in Pydantic. Pydantic will happily use dataclass definitions for (de)serialization, both in the case of nested definitions, and when using `pydantic.TypeAdapter(T)`. The only way it can tell if something is a dataclass is using `dataclasses.is_dataclass`, which itself just checks for the presence of `__dataclass_fields__` - an `attrs` class masquerading as a dataclass via `__dataclass_fields__` works just fine. 

Using `pydantic.TypeAdapter` at the edges and normal `attrs` in the core application avoids @tinche's concerns around things like:

> Is it really necessary to re-validate all your objects while reading them from a trusted database?

that this PR's author [largely agrees with](https://leontrolski.github.io/pydantic-wrong.html). One riposte to this might be "just use cattrs or switch to dataclasses", but I work on a large codebase that makes extensive use of both attrs and Pydantic - switching would be very painful.

Another example would be a "map over everything" function like the following:

```python
def map_[T](o: T, f: Callable[[T], T]) -> T:
    if isinstance(o, list):
        return [map_(v, f) for v in o]
    # ditto for tuple, set, frozenset, dict, then,
    if dataclasses.is_dataclass(o):
        kwargs = {k: map_(getattr(o, field.name), f) for field in dataclasses.fields(o)}
        return copy.replace(o, **kwargs)
    return o
```

With `dataclass_compatible=True` set, functions like this will work without having to muck around checking for the installation of `attrs` and using the subtly different `attrs.fields` function - see an example of mine [in the wild](https://github.com/leontrolski/ocdiff/blob/2a0232f2a46c0c31e5e82a429ada882d996daa04/python/ocdiff/repr_helpers.py#L250-L275).

# Pull Request Checklist

    - [ ] ...and used in the stub test file `typing_tests/baseline.py` or, if necessary, `typing_tests/mypy.py`.


hynek:
correct me if I'm wrong, but since this is a purely additive action, this should be possible to achieve using either a decorator or some kind of dynamic proxy/adapter that intercepts `__dataclass_fields__` and fills them from attrs fields?

leontrolski:
> this should be possible to achieve using either a decorator

It's true, and for now at least I have [make_attrs_class_dataclass_compatible.py](https://gist.github.com/leontrolski/52cafb2bba70c849d0319e14544d8d9c).

However, two reasons I still think this PR is a good idea:

- Boringly, it would make it really easy to feed attrs classes into pydantic.
- More philosophically...

<hr>

`is_dataclass(o)` is now in the stdlib, but what does it _mean_? One of: 
- `was_decorated_with_dataclass(o)`
- `has_traversable_fields(o)`

The first meaning to me seems pointless, if you can't get at the fields, why would you ever care that something is a dataclass? Depending on kwargs the decorator might not even add any methods. 

The latter meaning is useful, either a class is an opaque, atom-y thing (say `pathlib.Path`), or it has traversable fields (say `ast.FunctionDef`) that can be got at with `dataclasses.fields()`.

For Pydantic to work on a class, it is enough to know it has traversable fields, ditto with pretty printing, ditto with deeply copying/anonymising data, etc. If the `ast` nodes implemented `__dataclass_fields__`, the specialized `ast.walk` wouldn't have to exist.

**Turns out `__dataclass_fields__` is a useful public interface to implement, like `__iter__`.**

I've only just had this epiphany, so sorry if it sounds like the ramblings of a madman. However, if it convinces you, I might try convince msgspec, mashumaro, SQLAlchemy, marshmallow, Django ORM, etc. to do the same and we can all bask in a warm dataclass-fields-y future. If I haven't, continue as you were 🫡

PR #1611, opened 2026-08-12
Add `eq` validator

RHammond2:
# Summary

This PR adds the `eq` validator, which raises a `ValueError` when any value other than the expected value is passed. 

There are two possible alternatives to having an `eq` validator, but it would be nice to have a direct means to effectively freeze the value of an attribute without the freezing the whole class for the few niche use cases I've run across.

```python
# option 1
attrs.field(validator=attrs.validators.in_([2022]))

# option 2, as of 26.2
attrs.field(validator=attrs.validators.not_(attrs.validators.ne(2022))
```

Given the recent addition of `validators.ne` it felt right to add `validators.eq` to be used as follows.

```python
attrs.field(validator=attrs.validators.eq(2022))
```

# Pull Request Checklist

    - [ ] ...and used in the stub test file `typing_tests/baseline.py` or, if necessary, `typing_tests/mypy.py`.


hynek:
Thanks but eq is already covered by `attrs.validators.in_()`

hynek:
sorry for being glib, I just noticed you mentioned `in_`, however you didn't really say anything against it, except, that you wouldn't have to type square brackets I guess?

RHammond2:
The proposed addition of `eq` was so that the options for `in_` couldn't be expanded, effectively freezing the value. It's not a common use case, I've come across, so I'm not too worried about using `in_`.

hynek:
Internally it’s transformed into a tuple so that shouldn’t be an issue!

PR #1619, opened 2026-09-05
Resolve forward references in generated init annotations

### LINKED ISSUE #1596: Incorrect annotation for forward reference in generated constructor

Nikhi00718:
Fixes #1596.

On Python 3.14, generated attrs initializers can lose access to forward references declared later in the defining module because their private globals dictionary is copied during decoration. This adds a lazy annotation provider that resolves generated initializer annotations against the defining module while preserving unresolved forward references. A regression test covers `inspect.signature(..., eval_str=True)` and `typing.get_type_hints()`.

Tests: `python -m pytest tests -q` (1,408 passed, 4 skipped, 2 xfailed).


Nikhi00718:
Hi maintainers — this PR is still open and all 20 checks are passing. I would appreciate a review when convenient. Thanks!

Nikhi00718:
Hi maintainers — gentle follow-up on this PR. The current check suite is green; I’d appreciate a review when you have time. Happy to make any changes you suggest. Thanks!

